PR #94, opened 2025-11-21
MAUI Dockerfile updates so that it doesn't depend on the generated folder being present.

swensel:
# Description 

MAUI Dockerfile updates so that it doesn't depend on the generated folder being present.

Tested with `build.sh` and `run.sh` scripts on my local environment.

### DIFF
--- a/apps/main/Dockerfile
+++ b/apps/main/Dockerfile
@@ -26,24 +26,25 @@ ARG NODEJS_VERSION=22
 
 FROM node:$NODEJS_VERSION-bookworm-slim AS deps
 
-ENV NODE_ENV=production
+ENV NODE_ENV=development
 
 # Disable NextJS spyware
 ENV NEXT_TELEMETRY_DISABLED=1
 
 WORKDIR /app
 COPY .yarnrc.yml ./
-COPY packages/ui-common/generated ./generated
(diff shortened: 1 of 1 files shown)


apps/main/Dockerfile
```diff
@@ -55,15 +56,23 @@ ENV NEXT_TELEMETRY_DISABLED=1
 
 WORKDIR /app
 
-# From deps stage
-COPY --from=deps /app/node_modules ./node_modules
-COPY --from=deps /app/yarn.lock ./
+# Install only production dependencies
+COPY .yarnrc.yml ./
```
swensel: We still use production dependencies for this stage.  See `ENV NODE_ENV=production` on line [52](https://github.com/cognizant-ai-lab/neuro-san-ui/pull/94/files#diff-02fe02ee8a5c207223525838a51ac25fc12a0ffa4d2e51a7dc59c3bb26f6682eR52).

d1donlydfink (approved):
LGTM AFAICT.
Happy to approve for progress for Moment people.

./build.sh Has some transient warnings about "peer dependencies" but this doesn't seem to hamper a finished container at all for me, so that's an optimization for (perhaps much) later.

Issue #90, opened 2025-11-18
MAAUI: Honor sample_queries metadata


d1donlydfink:
neuro-san-studio has undergone some organization lately and now many agent networks have a "sample_queries" key on its metadata that contains a list of strings pertaining to how it could be invoked without a bunch of time-consuming typing.  This metadata comes over the Connectivy service.

Folks like Paul, who tend to use MAAUI more than nsflow are looking for ways to speed up their demos.

PR #96, opened 2025-12-10
#90 Follow up to sample queries PR: tests & copilot suggestions

dsargent:
Follow on from #95. Add suggested tests and other items.

### DIFF
--- a/packages/ui-common/__tests__/components/AgentChat/ChatCommon.test.tsx
+++ b/packages/ui-common/__tests__/components/AgentChat/ChatCommon.test.tsx
@@ -368,23 +368,6 @@ describe("ChatCommon", () => {
         expect(onChunkReceivedMock).toHaveBeenCalledWith(testResponseText)
     })
 
-    it("Should handle receiving empty chunks correctly", async () => {
-        const onChunkReceivedMock = jest.fn().mockReturnValue(true)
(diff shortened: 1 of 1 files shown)


packages/ui-common/components/AgentChat/ChatCommon.tsx
```diff
@@ -528,7 +529,7 @@ export const ChatCommon = forwardRef<ChatCommonHandle, ChatCommonProps>((props,
      * send the query to the agent.
      */
     const renderSampleQueries = (sampleQueries: string[]) => {
-        return (
+        return sampleQueries?.length > 0 ? (
```
dsargent: Copilot suggestion -- avoid generating empty `Box` if no sample queries.

packages/ui-common/components/AgentChat/ChatCommon.tsx
```diff
@@ -542,9 +543,12 @@ export const ChatCommon = forwardRef<ChatCommonHandle, ChatCommonProps>((props,
                         <Chip
                             id={`sample-query-${query}`}
                             key={query}
-                            label={query.length > 60 ? `${query.slice(0, 60)}...` : query}
+                            label={
+                                query.length > QUERY_TRUNCATE_LENGTH
```
dsargent: Use a constant for truncation length

PR #133, opened 2026-01-29
#687 Fix barrel imports and disallow them with an ESLint rule

dsargent:
"Mile wide, inch deep" PR.

The MUI documentation recommends avoiding the "barrel imports" style and instead importing components individually. More info: https://mui.com/material-ui/guides/minimizing-bundle-size/

This PR:

1) Fixes all existing barrel imports to use the recommended style
2) Adds an ESLint rule to prevent such imports from being used in future

No functional or visual changes in this PR. Purely technical as explained above.

### DIFF
--- a/packages/ui-common/components/Common/ConfirmationModal.tsx
+++ b/packages/ui-common/components/Common/ConfirmationModal.tsx
@@ -1,3 +1,19 @@
+/*
+Copyright 2025 Cognizant Technology Solutions Corp, www.cognizant.com.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 import Box from "@mui/material/Box"
 import Button from "@mui/material/Button"
 import {styled} from "@mui/material/styles"


apps/main/pages/api/userInfo/index.ts
```diff
@@ -68,7 +68,7 @@ function fetchUserInfoFromALB(req: NextApiRequest): UserInfoResponse {
     let oidcProvider: OidcProvider = null
     let username: string = null
 
-    // Look for "well-known" fields in the OIDC headers to figure out if we're using Github or AD
+    // Look for "well-known" fields in the OIDC headers to figure out if we're using GitHub or AD
```
dsargent: I also fixed some style issues in this PR, as in, incorrect brand names, grammar in comments. Seemed like a decent place to do that as it's already a clean-up PR that touches a load of files.

For example `GitHub` is correctly written that way, capital G, capital H.

packages/ui-common/utils/agentConversations.ts
```diff
@@ -134,7 +134,7 @@ export const processChatChunk = (
             }
             const textToShow = inquiryText || chatMessage.text
             // Show inquiry (from structure), that's only for networks that use AAOSA with a JSON format.
-            // Otherwise show the raw data from the `text` field of the chat message.
+            // Otherwise, show the raw data from the `text` field of the chat message.
```
dsargent: I agree with the language tool; a comma reads better here.

eslint.config.mjs
```diff
@@ -280,7 +280,17 @@ const config = [
...
-
+            "no-restricted-imports": [
+                "error",
+                {
+                    /*
+                    Bans any import that starts with @mui/ followed by a single segment (i.e., no additional slashes 
+                    after the first segment). For example, imports like @mui/core or @mui/icons would be restricted, 
+                    while deeper imports such as @mui/icons-material/Button would not be affected.
+                     */
+                    patterns: [{regex: "^@mui/[^/]+$"}],
```
dsargent: This is the regex suggested in the MUI doc and, I guess it's fine, but it's not perfect. It doesn't catch "second level" barrel imports like `import {Qux} from @mui/foo/bar`. But it will catch 90% and the regex would have to be a _lot_ more complex to catch those, if it's even possible.

Issue #130, opened 2026-01-23
neuro-san-ui: Fix Dependabot alerts and vulnerabilities listed in npmx.dev


dsargent:
There's currently 29. Fix/dismiss these as appropriate.

swensel:
Darren fixed a bunch of these already. There are 3 left:

<img width="1212" height="456" alt="Image" src="https://github.com/user-attachments/assets/cb00c846-29d2-48d4-97a8-049f97edec75" />

There's also 11 listed on npmx.dev:

<img width="881" height="693" alt="Image" src="https://github.com/user-attachments/assets/6260ba9e-c16b-413f-adcf-3729e4d869a3" />

PR #140, opened 2026-01-30
#130 More fixes for Dependabot alerts

dsargent:
Despite the previous fixes, some Dependabot [alerts](https://github.com/cognizant-ai-lab/neuro-san-ui/security/dependabot) still remain. `yarn audit --all` doesn't show all the same alerts as Dependabot, and I couldn't find a good way to get the Dependabot alerts locally; the only way seem to be to merge to `main` (or trick the repo into thinking your feature branch is the main branch) and Dependabot will then report.

This PR upgrades a few dependencies that were the subject of alerts. Also migrates us away from the deprecated, abandoned `dagre` library to the supported and maintained `@dagrejs/dagre` library, which was a source of some of the alerts.

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -20,12 +20,28 @@ limitations under the License.
 // https://nextjs.org/docs/pages/building-your-application/optimizing/testing#jest-and-react-testing-library
 import type {Config} from "@jest/types"
 
+// Packages in ESM format that need to be transformed by Jest because they are not distributed in CommonJS format
+const esmPackagesToTransform = [
+    "character-.*",
+    "comma-.*",
+    "decode-.*",
+    "hast.*",
+    "is-.*",
+    "lodash-es",
+    "mdast.*",
+    "parse-.*",
+    "property-.*",
+    "react-syntax-highlighter",
+    "refractor",
+    "space-.*",
+]
+
 const config: Config.InitialOptions = {
     // For details on these settings: https://jestjs.io/docs/configuration
 
     preset: "ts-jest/presets/default-esm",
     extensionsToTreatAsEsm: [".ts", ".tsx"],
-    transformIgnorePatterns: ["/node_modules/(?!next-auth|@next-auth|react|react-dom|react/jsx-runtime|lodash-es)"],
+    transformIgnorePatterns: [`/node_modules/(?!(${esmPackagesToTransform.join("|")})/)`],
     transform: {
         "^.+\\.(ts|tsx)$": [
             "ts-jest",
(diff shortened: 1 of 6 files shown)


jest.config.ts
```diff
@@ -20,12 +20,28 @@ limitations under the License.
...
+    "space-.*",
+]
+
 const config: Config.InitialOptions = {
     // For details on these settings: https://jestjs.io/docs/configuration
 
     preset: "ts-jest/presets/default-esm",
     extensionsToTreatAsEsm: [".ts", ".tsx"],
-    transformIgnorePatterns: ["/node_modules/(?!next-auth|@next-auth|react|react-dom|react/jsx-runtime|lodash-es)"],
+    transformIgnorePatterns: [`/node_modules/(?!(${esmPackagesToTransform.join("|")})/)`],
```
dsargent: Use the exception list items from the array.

packages/ui-common/__tests__/components/AgentChat/FormattedMarkdown.test.tsx
```diff
@@ -15,7 +15,6 @@ limitations under the License.
 */
 
 import {render, screen, waitFor} from "@testing-library/react"
-import type {SyntaxHighlighterProps} from "react-syntax-highlighter"
```
dsargent: We don't need this import any more

packages/ui-common/package.json
```diff
@@ -9,17 +9,17 @@
         "prepack": "echo PREPACK_RUNNING && yarn build"
     },
     "dependencies": {
+        "@dagrejs/dagre": "1.0.4",
```
dsargent: Migrate to newer `dagre` library.

Note: there is a much newer release available -- `2.0.4` or something -- _but_ we can't use it due to this [breaking issue](https://github.com/dagrejs/dagre/issues/491) which affects us directly. Very annoying. We should upgrade as soon as they fix it.

packages/ui-common/package.json
```diff
@@ -9,17 +9,17 @@
...
         "http-status": "1.7.3",
         "jsonrepair": "3.8.0",
         "lodash-es": "4.17.23",
         "next": "16.1.6",
         "next-auth": "5.0.0-beta.30",
         "notistack": "3.0.1",
         "react-dom": "18.2.0",
-        "react-markdown": "9.0.1",
-        "react-syntax-highlighter": "15.5.0",
+        "react-markdown": "10.1.0",
```
dsargent: More version bumps to address vulnerabilities.

swensel (approved):
LGTM. Thanks for doing this @dsargent

PR #148, opened 2026-02-12
#147 Enforce & use arrow functions throughout

dsargent:
Adds a new ESLint plugin + rule to enforce this in future. Fixes all existing violations.

Boring, shallow, mostly mechanical PR.

The weirdest thing for me: this migration actually made the code briefer in quite a few places, yet coverage didn't budge by one iota. I feel cheated.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1426,7 +1426,7 @@ __metadata:
     eslint-plugin-import: "npm:2.32.0"
     eslint-plugin-jest: "npm:29.0.1"
     eslint-plugin-jest-dom: "npm:5.5.0"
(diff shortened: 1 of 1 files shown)


eslint.config.mjs
```diff
@@ -40,6 +41,7 @@ const compat = new FlatCompat({
 })
 
 const config = [
+    preferArrowFunctions.configs["all"],
```
dsargent: Default "bagel with everything" config for the new rule. It offers [configuration options](https://github.com/JamieMason/eslint-plugin-prefer-arrow-functions) but the default seems fine.

PR #150, opened 2026-02-13
Adapt cull script for npmjs.org; remove deprecate fallback

donn-leaf:
## Summary

The existing cull script used the GitHub Packages API (`gh api`) which no longer applies now that we publish to npmjs.org. This rewrites the script to use `npm view` for listing versions/timestamps and `npm unpublish` for removal, authenticated via an `NPM_TOKEN` secret. The deprecation fallback was removed per team decision — versions that fail to unpublish are counted and reported rather than silently deprecated.

### DIFF
--- a/build_scripts/cull_npm_packages.sh
+++ b/build_scripts/cull_npm_packages.sh
@@ -278,7 +278,7 @@ print_summary() {
 emit_github_output() {
     if [ -n "${GITHUB_OUTPUT:-}" ]; then
         {
-            echo "cutoff_date=$CUTOFF_DATE"
+            echo "cutoff_date=$(date --date "@$CUTOFF_DATE" --utc +%Y-%m-%dT%H:%M:%SZ)"
             echo "total=$TOTAL_VERSIONS"
             echo "release=$RELEASE_VERSIONS"
             echo "kept=$KEPT_VERSIONS"


donn-leaf:
Add a comment in the appropriate place in the cull script to document this esoteric knowledge:

npm unpublish is blocked by npm's 72-hour/300-download policy

In our use case, we will be past npm's 72 hour limit, but it's highly unlikely we will surpass the 300 download limit on any non-released version. Therefore, don't mention the 72 hour limit in any errors/hints logged during the scripts, as that would be misleading to the uninitiated user.

build_scripts/cull_npm_packages.sh
```diff
@@ -85,17 +98,16 @@ is_release_version() {
 # ================================================================================
 
 parse_args() {
-    PACKAGE_NAME="${1:?Usage: $0 <package_name> <package_owner> <retention_days> <dry_run> <min_keep_versions>}"
-    PACKAGE_OWNER="${2:?Usage: $0 <package_name> <package_owner> <retention_days> <dry_run> <min_keep_versions>}"
-    RETENTION_DAYS="${3:?Usage: $0 <package_name> <package_owner> <retention_days> <dry_run> <min_keep_versions>}"
-    DRY_RUN="${4:?Usage: $0 <package_name> <package_owner> <retention_days> <dry_run> <min_keep_versions>}"
-    MIN_KEEP_VERSIONS="${5:?Usage: $0 <package_name> <package_owner> <retention_days> <dry_run> <min_keep_versions>}"
+    SCOPED_PACKAGE_NAME="${1:?Usage: $0 <scoped_package_name> <retention_days> <dry_run> <min_keep_versions>}"
```
dsargent: Could be DRY-er: the "usage" is defined in a comment in the header, _and_ repeated in each of these. Instead of all that, why not have a standard usage() function that displays usage info and call it from all relevant places?

dsargent (approved):

Issue #131, opened 2026-01-14
Display temporary networks in MAUI


dsargent:
Placeholder. TBD.

PR #158, opened 2026-02-20
#131 Add support for temporary networks

dsargent:
Like the title says -- add support for temporary networks coming back from Neuro-san in the `sly_data` stream. They show up under a pseudo category `Temporary` in the networks list and can be interacted with just like any other network.

Tested with the "copy cat" meta network; couldn't get it to work, yet, with Agent Network Designer due to various timeouts and errors.

Screenshot:

image

Demo:

https://github.com/user-attachments/assets/d4dd309a-256e-430b-a0a7-39190e273f18

Known issues:
1) Sometimes thought bubbles hang around when aborting agent interaction (sporadic, can't repro consistently). See screenshot.
2) Serial number for networks only displayed in sidebar, not in Chat, notifications etc. So "Foo 1", "Foo 2", etc. but in Chat you just get "Foo".
3) "New network created" alert doesn't always show up. Probably a rendering/timing issue.
4) Weird "spider" glitch in flow sometimes, unable to repro but does happen. (see screenshot)
5) Newly generated network doesn't receive instructions, for example if you ask CopyCat to clone "hello world" and ask it something, you get the cloned network but the "ask it something" doesn't happen.

Thought bubbles hanging around:

image

Unwanted spider:

flow_spider_glitch

### DIFF
--- a/packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
+++ b/packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
@@ -324,14 +324,6 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
             const currentNetworks = useTempNetworksStore.getState().tempNetworks
             useTempNetworksStore.getState().setTempNetworks([...currentNetworks, ...newTemporaryNetworks])
 
-            // Set recommended icon (hourglass) for new temporary networks
-            setNetworkIconSuggestions((prev) => ({
-                ...prev,
-                ...Object.fromEntries(
-                    newTemporaryNetworks.map((network) => [network.agentInfo.agent_name, "HourglassTop"])
-                ),
-            }))
-
             // record the new temporary networks so we can select them for the user. For now, we only
             // care about the first one.
             setNewlyAddedTemporaryNetworks(new Set(newTemporaryNetworks.map((network) => network.agentInfo.agent_name)))
@@ -360,27 +352,33 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
     }, [haveShownPopup])
 
     const onStreamingComplete = useCallback(() => {
-        // When streaming is complete, clean up any refs and state
-        conversationsRef.current = null
-        setCurrentConversations(null)
-        resetState()
-
         const network = newlyAddedTemporaryNetworks?.values().next().value
         if (network?.length > 0) {
             // We show an Alert after streaming completes (in case of Zen mode where the user might miss it)
             const agentNameDisplay = cleanUpAgentName(removeTrailingUuid(network))
             setAlertContents(`A temporary network "${agentNameDisplay}" has been created.`)
+        }
 
+        // When streaming is complete, clean up any refs and state
+        conversationsRef.current = null
+        setCurrentConversations(null)
+        resetState()
+    }, [newlyAddedTemporaryNetworks])
+
+    useEffect(() => {
+        if (alertContents?.length > 0) {
             // Set a timer to clear the alert after a few seconds so it doesn't overstay its welcome
-            const timeoutId = setTimeout(() => {
+            const timeoutId = window.setTimeout(() => {
                 setAlertContents(null)
-            }, 10_000) // Clear after 10 seconds
+            }, 10_000)
 
-            return () => clearTimeout(timeoutId)
+            return () => {
+                window.clearTimeout(timeoutId)
+            }
         } else {
             return undefined
         }
-    }, [newlyAddedTemporaryNetworks])
+    }, [alertContents])
 
     const [confirmationModalOpen, setConfirmationModalOpen] = useState<boolean>(false)
 
@@ -545,46 +543,51 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
         )
     }
 
+    const getConfirmationModal = () =>
+        confirmationModalOpen ? (
+            <ConfirmationModal
+                id="delete-network-confirmation-modal"
+                content={
+                    `The network "${cleanUpAgentName(removeTrailingUuid(networkToBeDeleted))}" will be deleted. ` +
+                    "This action cannot be undone. Are you sure you want to proceed?"
+                }
+                handleCancel={() => {
+                    setConfirmationModalOpen(false)
+                    setNetworkToBeDeleted(null)
+                }}
+                handleOk={() => {
+                    useTempNetworksStore
+                        .getState()
+                        .setTempNetworks(
+                            temporaryNetworks.filter((network) => network.agentInfo.agent_name !== networkToBeDeleted)
+                        )
+                    setNetworkToBeDeleted(null)
+                    setConfirmationModalOpen(false)
+                }}
+                title="Delete Network"
+            />
+        ) : null
+
+    const getAlert = () => (
+        <Collapse
+            in={alertContents?.length > 0}
+            timeout={1000}
+        >
+            <MUIAlert
+                id="temporary-network-created-alert"
+                closeable={true}
+                severity="success"
+                sx={{marginTop: "1rem", marginBottom: 0}}
+            >
+                {alertContents}
+            </MUIAlert>
+        </Collapse>
+    )
+
     return (
         <>
-            <Collapse
-                in={alertContents?.length > 0}
-                timeout={1000}
-            >
-                <MUIAlert
-                    id="temporary-network-created-alert"
-                    closeable={true}
-                    severity="success"
-                    sx={{marginTop: "1rem", marginBottom: 0}}
-                >
-                    {alertContents}
-                </MUIAlert>
-            </Collapse>
-            {confirmationModalOpen && (
-                <ConfirmationModal
-                    id="delete-network-confirmation-modal"
-                    content={
-                        `The network "${cleanUpAgentName(removeTrailingUuid(networkToBeDeleted))}" will be deleted. ` +
-                        "This action cannot be undone. Are you sure you want to proceed?"
-                    }
-                    handleCancel={() => {
-                        setConfirmationModalOpen(false)
-                        setNetworkToBeDeleted(null)
-                    }}
-                    handleOk={() => {
-                        useTempNetworksStore
-                            .getState()
-                            .setTempNetworks(
-                                temporaryNetworks.filter(
-                                    (network) => network.agentInfo.agent_name !== networkToBeDeleted
-                                )
-                            )
-                        setNetworkToBeDeleted(null)
-                        setConfirmationModalOpen(false)
-                    }}
-                    title="Delete Network"
-                />
-            )}
+            {getAlert()}
+            {getConfirmationModal()}
             <Grid
                 id="multi-agent-accelerator-grid"
                 container
--- a/packages/ui-common/components/MultiAgentAccelerator/Sidebar/AgentNetworkTreeItem.tsx
+++ b/packages/ui-common/components/MultiAgentAccelerator/Sidebar/AgentNetworkTreeItem.tsx
@@ -18,7 +18,7 @@ import {TreeItemProvider} from "@mui/x-tree-view/TreeItemProvider"
 import {useTreeItem} from "@mui/x-tree-view/useTreeItem"
 import {FC, useRef} from "react"
 
-import {NodeIndex} from "./TreeUtils"
+import {NodeIndex} from "./TreeBuilder"
 import {cleanUpAgentName} from "../../AgentChat/Utils"
 
 // Palette of colors we can use for tags
@@ -106,7 +106,7 @@ export const AgentNetworkTreeItem: FC<AgentNetworkNodeProps> = ({
     const isTemporaryNetwork = Boolean(expirationTime)
     const isExpired = isChild && isTemporaryNetwork && isTemporaryNetworkExpired(expirationTime)
 
-    const iconNameSuggestion = isChild ? networkIconSuggestions?.[itemId] : null
+    const iconNameSuggestion = isTemporaryNetwork ? "HourglassTop" : isChild ? networkIconSuggestions?.[itemId] : null
 
     let muiIconElement = null
     if (iconNameSuggestion && MuiIcons[iconNameSuggestion as keyof typeof MuiIcons]) {
--- a/packages/ui-common/components/MultiAgentAccelerator/Sidebar/Sidebar.tsx
+++ b/packages/ui-common/components/MultiAgentAccelerator/Sidebar/Sidebar.tsx
@@ -37,7 +37,7 @@ import {
 } from "react"
 
 import {AgentNetworkNodeProps, AgentNetworkTreeItem} from "./AgentNetworkTreeItem"
-import {buildTreeViewItems} from "./TreeUtils"
+import {buildTreeViewItems} from "./TreeBuilder"
 import {testConnection, TestConnectionResult} from "../../../controller/agent/Agent"
 import {NetworkIconSuggestions} from "../../../controller/Types/NetworkIconSuggestions"
 import {AgentInfo} from "../../../generated/neuro-san/NeuroSanClient"
@@ -230,6 +230,10 @@ export const Sidebar: FC<SidebarProps> = ({
                 prev.includes(TEMPORARY_NETWORK_FOLDER) ? prev : [...prev, TEMPORARY_NETWORK_FOLDER]
             )
             highlightTimeout = setTimeout(() => {
+                // Scroll the selected node into view and add an animation to draw the user's attention to it.
+                // Hacky: use a DOM query to find the node. I tried the various ways to do this programmatically
+                // in MUI RichTreeView including the imperative API (https://mui.com/x/react-tree-view/rich-tree-view/selection/#imperative-api)
+                // but couldn't get it to work so resorting to this for now.
                 const selectedNode = document.querySelector("[role=treeitem][aria-checked=true]")
                 if (selectedNode) {
                     selectedNode.scrollIntoView({behavior: "smooth", block: "nearest", inline: "nearest"})
--- a/apps/main/__tests__/pages/api/handlers/Handlers.test.ts
+++ b/apps/main/__tests__/pages/api/handlers/Handlers.test.ts
@@ -1,4 +1,5 @@
 import {ChatPromptTemplate} from "@langchain/core/prompts"
+import httpStatus from "http-status"
 import {NextApiRequest, NextApiResponse} from "next"
 import {createMocks} from "node-mocks-http"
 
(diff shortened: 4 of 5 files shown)


apps/main/__tests__/pages/MultiAgentAcceleratorPage.test.tsx
```diff
@@ -366,12 +407,11 @@ describe("Multi Agent Accelerator Page", () => {
...
 
-        // Assert that none of the calls has TEST_AGENT_MATH_GUY
-        expect(
-            calls.some((call) =>
-                call[0]?.some((conv: {agents: Set<string>}) => Array.from(conv.agents).includes(TEST_AGENT_MATH_GUY))
-            )
-        ).toBe(false)
+        // Assert that conversationMock was always called with an empty array (no conversations)
+        // due to the failure in extractConversations
+        calls.forEach((call) => {
```
dsargent: Simplify assertion

apps/main/pages/api/Common/Types.ts
```diff
@@ -0,0 +1,6 @@
+/**
```
dsargent: New type for ErrorResponse. I made it a separate module so that it can be re-used by the caller (`Agent.ts`) without pulling in the server-side dependencies, which is what would happen if I put it in `LlmHandler.ts`. Doesn't really matter for our "trivial" bundle, but good practice.

packages/ui-common/__tests__/components/AgentChat/ChatCommon.test.tsx
```diff
@@ -425,6 +421,28 @@ describe("ChatCommon", () => {
         expect(screen.queryByText(/Error occurred:/u)).not.toBeInTheDocument()
     })
 
+    it("Should handle other types of errors than AbortError correctly", async () => {
```
dsargent: More tests to hit and improve coverage numbers. This one is actually useful.

packages/ui-common/__tests__/components/Common/Footer.test.tsx
dsargent: And yet more sacrificial tests.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentConversations.test.ts
```diff
@@ -15,25 +15,16 @@ limitations under the License.
...
 import {
     AgentConversation,
     createConversation,
+    extractConversations,
     isFinalMessage,
-    processChatChunk,
-    updateAgentCounts,
-} from "../../../utils/agentConversations"
-
-// Mock the chatMessageFromChunk utility
```
dsargent: No longer needed as we convert to a `ChatMessage` _before_ calling the functions in this module now. See implementation in the actual module.

packages/ui-common/components/AgentChat/ChatCommon.tsx
```diff
@@ -952,252 +956,281 @@ export const ChatCommon = forwardRef<ChatCommonHandle, ChatCommonProps>((props,
                 flexDirection: "column",
                 flexGrow: 1,
                 height: "100%",
+                position: "relative",
             }}
         >
-            {title && (
+            {!targetAgent && (
```
dsargent: This is the new feature to disable interaction if no target agent is specified.

Another terminology issue: "agent" vs "network". We tend to conflate the two, and I know Dan's take is "the user shouldn't care whether it's an agent or network, that's just an implementation detail". But I find it super confusing.

packages/ui-common/components/MultiAgentAccelerator/Sidebar/AgentNetworkTreeItem.tsx
```diff
@@ -38,41 +41,51 @@ type TagColor = (typeof TAG_COLORS)[number]
...
-    readonly setSelectedNetwork: (network: string) => void
-    readonly shouldDisableTree: boolean
+    readonly nodeIndex: NodeIndex
+    readonly onDeleteNetwork?: (network: string, isExpired: boolean) => void
     readonly networkIconSuggestions: Record<string, string>
+    readonly temporaryNetworkExpirationTimes?: Record<string, Date>
+}
+
+/**
+ * Helper function to determine if a temporary network is expired based on its expiration date
```
dsargent: See JSDoc.

packages/ui-common/components/MultiAgentAccelerator/Sidebar/Sidebar.tsx
```diff
@@ -63,23 +66,31 @@ enum CONNECTION_STATUS {
 export interface SidebarProps {
     readonly customURLCallback: (url: string) => void
     readonly customURLLocalStorage?: string
-    readonly networkIconSuggestions?: Record<string, string>
     readonly id: string
     readonly isAwaitingLlm: boolean
+    readonly networkIconSuggestions?: NetworkIconSuggestions
```
dsargent: Alphabetic sort + temp network stuff.

packages/ui-common/components/MultiAgentAccelerator/Sidebar/Sidebar.tsx
```diff
@@ -228,17 +323,23 @@ export const Sidebar: FC<SidebarProps> = ({
...
                     }}
                     // Pass custom props to tree items via slotProps.
                     // Reference: https://github.com/mui/mui-x/issues/13351
                     slotProps={{
                         item: {
-                            nodeIndex: index,
-                            setSelectedNetwork,
-                            shouldDisableTree: isAwaitingLlm,
                             networkIconSuggestions,
+                            nodeIndex,
```
dsargent: Alphabetic sort + pass temp networks stuff.

packages/ui-common/components/MultiAgentAccelerator/Sidebar/TreeUtils.ts
```diff
@@ -16,90 +27,119 @@ const sortTreeNodes = (nodes: TreeViewBaseItem[]): void => {
...
+            })
             queue.push(...node.children)
         }
     }
 }
 
+/**
+ * Add a single AgentInfo entry into the tree structure
+ */
+const addNetworkToTree = (
```
dsargent: Factored this into a common function since now we have two groups of networks to add: regular and temp ones.

packages/ui-common/components/MultiAgentAccelerator/Sidebar/TreeUtils.ts
```diff
@@ -16,90 +27,119 @@ const sortTreeNodes = (nodes: TreeViewBaseItem[]): void => {
...
+
+        parts.forEach((part, index) => {
+            const nodeId = parts.slice(0, index + 1).join("/")
+            let node = map.get(nodeId)
+
+            if (!node) {
+                node = {id: nodeId, label: part, children: []}
+                map.set(nodeId, node)
+                if (index === parts.length - 1) {
+                    const agentNameWithoutUuid = removeTrailingUuid(part)
```
dsargent: Convoluted logic to add a sequence number when we have dupe names due to shortening -- "Foo 1", "Foo 2" etc. Requested by Dan.

packages/ui-common/components/MultiAgentAccelerator/AgentConversations.ts
dsargent: Not new, but refactored:

1) Moved from `utils` to here, which is a better place for it IMO.¹
2) Changed name casing to match our PascalCase standard.
3) Split out unrelated stuff like agent counts into its own module. [Separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns), [single responsibility principle](https://en.wikipedia.org/wiki/Single_responsibility_principle).

Now this module _only_ deals with agent conversations, like its name says.

¹ In general I'm not a fan of "utils", "misc" type modules. They tell you nothing about what they do. There's almost always a better place. Yes, I know we violate this in our codebase and yes I'm gradually addressing it. In my opinion, "utils", "tools", "misc" etc. should be a _last_ resort: "This thing is so generic, so general, that I really just can't think of a better name/place for it.".

packages/ui-common/components/MultiAgentAccelerator/AgentNode.tsx
```diff
@@ -27,9 +27,9 @@ import Typography from "@mui/material/Typography"
 import {FC} from "react"
 import {Handle, NodeProps, Position} from "reactflow"
 
+import {AgentConversation} from "./AgentConversations"
```
dsargent: Moved/refactored.

packages/ui-common/components/MultiAgentAccelerator/GraphLayouts.ts
```diff
@@ -154,7 +155,7 @@ export const layoutRadial = (
     currentConversations: AgentConversation[] | null, // For plasma edges (live) and node highlighting
     isAwaitingLlm: boolean,
     thoughtBubbleEdges: Map<string, {edge: Edge<EdgeProps>; timestamp: number}>,
-    agentIconSuggestions: Record<string, string> = {}
+    agentIconSuggestions: AgentIconSuggestions = null
```
dsargent: Take advantage of better typing throughout this module.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -223,21 +265,90 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
...
+
+            // Set recommended icon (hourglass) for new temporary networks
+            setNetworkIconSuggestions((prev) => ({
+                ...prev,
+                ...Object.fromEntries(
+                    newTemporaryNetworks.map((network) => [network.agentInfo.agent_name, "HourglassTop"])
+                ),
+            }))
+
+            // record the new temporary networks so we can select them for the user. For now, we only
```
dsargent: Limitation -- we assume only one temp network per chat stream, or else it gets a lot more comlpex. I don't feel like handling the general case here, and I don't see a use case for it, so off it goes punted into the future.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -249,11 +360,38 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
     }, [haveShownPopup])
 
     const onStreamingComplete = useCallback((): void => {
+        const network = newlyAddedTemporaryNetworks?.values().next().value
```
dsargent: This is the buggy notification thing. When streaming completes, if we saw a temp network, we display the notification.

I googled and Copiloted for ways to have an "expiring MUI Alert" but they all seem to recommend this, using a timer, which is hokey but i guess it's The Way?!

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -249,11 +360,38 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
...
         // When streaming is complete, clean up any refs and state
         conversationsRef.current = null
         setCurrentConversations(null)
         resetState()
-    }, [])
+    }, [newlyAddedTemporaryNetworks])
+
+    const [confirmationModalOpen, setConfirmationModalOpen] = useState<boolean>(false)
+
+    const handleDeleteNetwork = (networkId: string, isExpired: boolean) => {
```
dsargent: When a user manually deletes a temp network (functionality requested by Dan).

packages/ui-common/components/MultiAgentAccelerator/TemporaryNetworks.ts
```diff
@@ -0,0 +1,39 @@
+import {ChatMessage, ChatMessageType} from "../../generated/neuro-san/NeuroSanClient"
+
+/**
+ * Definition of a temporary network. No schema for this provided by backend so we second-guess it here.
```
dsargent: Dan is [working](https://github.com/cognizant-ai-lab/neuro-san/issues/743) on the "no schema for output `sly_data` issue.

packages/ui-common/controller/agent/Agent.ts
```diff
@@ -43,7 +46,13 @@ import {sendLlmRequest} from "../llm/LlmChat"
  * @returns The path with the target agent name inserted.
  */
 const insertTargetAgent = (agent: string, path: string) => {
-    return path.replace("{agent_name}", agent)
+    let agentTmp = agent
```
dsargent: See comment.

packages/ui-common/controller/Types/Branding.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Items returned by the branding API.
+ */
+export type BrandingSuggestions = {
```
dsargent: Strong typing for API.

packages/ui-common/const.ts
```diff
@@ -19,13 +19,11 @@ export const LOGO: string = "Neuro® AI"
...
 
-export const CONTACT_US_CONFIRMATION_DIALOG_TITLE = "Contact Us"
-
-export const CONTACT_US_CONFIRMATION_DIALOG_TEXT =
+export const getContactUsConfirmationText = (email: string): string =>
     "Would you like to send the Cognizant Neuro AI support team an email? " +
     "You will need to have an email client installed on your device in order " +
     "to continue. If you don't have an email client, you can still contact us at " +
-    "NeuroAiSupport@cognizant.com using a web based email client."
+    `${email} using a web based email client.`
```
dsargent: This was the bug with the hardcoded email address. I found it while writing unit tests to get coverage up.

apps/main/__tests__/pages/MultiAgentAcceleratorPage.test.tsx
```diff
@@ -344,13 +389,9 @@ describe("Multi Agent Accelerator Page", () => {
...
     })
 
-    it("should handle receiving a bad chunk", async () => {
-        // Make processChatChunk return success as false for this one test to simulate a critical error
-        ;(processChatChunk as jest.Mock).mockReturnValue({
-            success: false,
-            newCounts: new Map([[TEST_AGENT_MATH_GUY, 1]]),
-            newConversations: [{agents: new Set([TEST_AGENT_MATH_GUY])}],
-        })
+    it("should handle receiving a bad message", async () => {
```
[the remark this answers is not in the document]
dsargent: Yeah I removed some unnecessary try-catches from the extraction logic and changed the return contract to be simpler. Either conversations array, `[]`, or `null` on error. But looking at the logic I don't even see how `null` is even possible any more...

apps/main/pages/api/Common/Types.ts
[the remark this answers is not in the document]
dsargent: Will fix

packages/ui-common/components/MultiAgentAccelerator/Sidebar/Sidebar.tsx
```diff
@@ -172,10 +185,92 @@ export const Sidebar: FC<SidebarProps> = ({
...
+        // If we got a new temporary network, select it and expand the temporary category in the tree view
+        if (newlyAddedTemporaryNetworks?.size > 0) {
+            const firstItem = newlyAddedTemporaryNetworks.values().next().value
+            setSelectedItem(firstItem)
+            setSelectedNetwork(firstItem)
+            setExpandedItems((prev) =>
+                prev.includes(TEMPORARY_NETWORK_FOLDER) ? prev : [...prev, TEMPORARY_NETWORK_FOLDER]
+            )
+            highlightTimeout = setTimeout(() => {
+                const selectedNode = document.querySelector("[role=treeitem][aria-checked=true]")
```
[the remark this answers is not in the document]
swensel: Yeah I don't love it either. I also wonder if `[aria-checked=true]` is a little fragile. I do think we should proceed. Maybe we can revisit it at some point.

packages/ui-common/components/MultiAgentAccelerator/Sidebar/TreeUtils.ts
```diff
@@ -16,90 +27,119 @@ const sortTreeNodes = (nodes: TreeViewBaseItem[]): void => {
...
+    network: AgentInfo,
+    result: TreeViewBaseItem[],
+    uncategorized: TreeViewBaseItem,
+    map: Map<string, TreeViewBaseItem>,
+    nodeIndex: NodeIndex,
+    displayNameCounts: Map<string, number>
+): void => {
+    const parts = network.agent_name.split("/")
+
+    if (parts.length === 1) {
```
swensel: Some more comments in this function might be helpful, but not necessary.

packages/ui-common/components/MultiAgentAccelerator/AgentCounts.ts
```diff
@@ -0,0 +1,27 @@
+import {Origin} from "../../generated/neuro-san/NeuroSanClient"
```
swensel: We may want to add a copyright here.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -223,21 +265,90 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
...
+        if (result != null) {
+            conversationsRef.current = result
+            setCurrentConversations(result)
+        }
+
+        // Agent hit counts
+        agentCountsRef.current = getUpdatedAgentCounts(agentCountsRef.current, chatMessage?.origin)
+
+        // Temporary networks/reservations
+        const reservationsResult = extractReservations(chatMessage)
```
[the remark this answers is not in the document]
swensel: I like it. Thanks for doing this.

packages/ui-common/components/MultiAgentAccelerator/Sidebar/Sidebar.tsx
```diff
@@ -172,10 +185,92 @@ export const Sidebar: FC<SidebarProps> = ({
...
+
+        return () => {
+            clearTimeout(highlightTimeout)
+            clearTimeout(removeHighlightTimeout)
+        }
+    }, [newlyAddedTemporaryNetworks])
 
     return (
         <>
+            <style>
```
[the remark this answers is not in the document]
dsargent: I hesitated...we are sort of moving away from `globals.css`.

packages/ui-common/controller/agent/Agent.ts
```diff
@@ -43,7 +46,13 @@ import {sendLlmRequest} from "../llm/LlmChat"
...
+    // Remove "temporary" prefix from network name if it exists since the server doesn't know about that convention
+    if (agentTmp.startsWith("temporary/")) {
```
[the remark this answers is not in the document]
swensel: Good idea.

swensel (approved):
Approved.

I think the two things that stood out are:

1. Using [document.querySelector](https://github.com/cognizant-ai-lab/neuro-san-ui/pull/158/changes#r2897653551) in React code. I think it's OK to leave it though, at least for now.
2. [Splitting the timeout in the useCallback to a useEffect](https://github.com/cognizant-ai-lab/neuro-san-ui/pull/158/changes#r2898217575), I think that would be better. I probably would update that before merging, but it could also be done in a quick follow up.

Also, nice work. This is a great feature for MAUI 🎉 .

PR #167, opened 2026-02-26
Restore jest/expect-expect rule.

swensel:
# Description

Restore jest/expect-expect rule.

This was on my radar with the lint rules. I thought it was OK, but once I brought the lint rules over to LMX (https://github.com/cognizant-ai-lab/idea-brainstorm-demo/pull/115), I decided we should bring it back.

FWIW, there were no other rules that I was on the fence about, it was only this one.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -402,6 +402,8 @@ const config = [
             "func-style": "off",
             "id-length": "off",
             "init-declarations": "off",
+            // Indicate that the findBy and getBy calls are implicit assertions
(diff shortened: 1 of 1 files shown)


eslint.config.mjs
```diff
@@ -402,6 +402,8 @@ const config = [
             "func-style": "off",
             "id-length": "off",
             "init-declarations": "off",
+            // Indicate that the findBy and getBy calls are implicit assertions
```
dsargent: Whoa, I didn't notice this one got dropped. Yes, we absolutely should have this.

Issue #73, opened 2025-10-20
[UN-3491] MAA UI Agent designer "Vibe mode"


dsargent:
**Migrated from Jira:** [UN-3491](https://leaf-ai.atlassian.net/browse/UN-3491)

## Description
Start with fresh network and add nodes as network is built

Keep user on the same screen.

For agent designer, don’t show the network that builds networks, start fresh and add nodes as the new network is built.

## Jira Details
- **Issue Type:** Task
- **Priority:** Major
- **Status:** Triage
- **Reporter:** Darren Sargent
- **Assignee:** Sean Wensel
- **Created:** 2025-09-08 13:28:32
- **Updated:** 2025-10-20 10:30:07

### Comment by Sean Wensel (2025-09-08 14:46:38)
Copilot style UX

### Comment by Sean Wensel (2025-09-15 14:54:28)
Sounds like we would like if:

** This could be demoed at Dreamforce
** It’s used for next vibe coding hackathon for Cognizant

### Comment by Sean Wensel (2025-09-15 15:03:33)
Apparently a lot of `sly_data` is used, so we’ll need to look into that.

### Comment by Sean Wensel (2025-09-26 17:15:46)
Paul/Julianna mentioned that Lovable is a good example.

### Comment by Sean Wensel (2025-09-26 17:21:53)
Different mode of interaction vs a different page.

We could also have a “+” icon to create a new network.

Brainstorm solutions with team.

### Comment by Sean Wensel (2025-10-06 14:39:32)
NSFlow is going to have the agent designer, not MAUI.

PR #226, opened 2026-03-19
#73 Fix network vibe coding

dsargent:
Fixes a very fundamental issue in `LlmChat.ts` whereby long messages were being skipped.  This was causing network vibe coding (Agent Network Designer interactions) to fail.

The code was making the naive, incorrect assumption that a chunk always contains one or more full messages; for long chat messages, this is not valid. Such a message could and does get broken into multiple chunks by the underlying infrastructure.

Unfortunately, one example of such a long message is the `AGENT_FRAMEWORK` completion message from Agent Network Designer. Net result, MAUI was not processing the messages with the info on the generated temporary networks. With this fix, it does.

This is a change to the low-level `sendLlmRequest` and thereby to `ChatCommon`. This is also used by Neuro UI for communication with Agents/LLMs so I tested that too. Everything seems fine.

The change adds a new enum and option to the LLM communication functions, allowing for messages to be received in chunks, or as entire lines.

### DIFF
--- a/packages/ui-common/__tests__/unit/Controller/LlmChat.test.ts
+++ b/packages/ui-common/__tests__/unit/Controller/LlmChat.test.ts
@@ -17,7 +17,28 @@ limitations under the License.
 import httpStatus from "http-status"
 
 import {withStrictMocks} from "../../../../../__tests__/common/strictMocks"
-import {sendLlmRequest} from "../../../controller/llm/LlmChat"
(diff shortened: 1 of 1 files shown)


packages/ui-common/__tests__/unit/Controller/Agent.test.ts
dsargent: Add new param to test

PR #257, opened 2026-04-17
Linter updates, sync with Neuro UI

dsargent:
Technical PR only.

Updates a bunch of versions of dev dependencies to bring them in line with Neuro UI. In fact, this project is now ahead of that one, so we'll have to leapfrog and do a similar PR going the other way.

Clean up the linter config and make use of native flat configs where possible.

Re-enable linter rule `import/no-extraneous-dependencies` -- was a simple fix to make it work with workspaces 🎉 

I used the new `lint-rules` yarn script to see what changed in the rules compared to `main`. It's good news -- almost entirely additive, new rules due to bumping the various linter plugins and ESLint itself. I'll attach the before-and-after diffs for you to compare.

[branch.json](https://github.com/user-attachments/files/26848325/branch.json)
[main.json](https://github.com/user-attachments/files/26848327/main.json)

Summary of the changes I see when diffing those files:
* Some rule levels changed from 2 to 1. That is "meh". That means error -> warning. We fail the build on all levels of violation so it's moot to us. The plugin authors must have decided to "downgrade" some violations to warnings.
* `import/no-extraneous-dependencies` now enabled and fixed for workspaces 🎉 
* `react-hooks` gets lots of new rules due to the upgrade; I had to disable some because we're not compliant _but_ we should re-enable them and fix the issues as there's some very good stuff in there that helps with the various React pitfalls.
* Had to disable `react/forward-ref-uses-ref` as it doesn't play nicely with ESLint 10 and the various upgrades. See comment in `eslint.config.mjs`. We should revisit this once the vendors fix it.
* Added `ignoreStateless: false` to `react/no-multi-comp` to get the same behavior as `main`
* Some rules get new default config settings inherited from vendor updates -- expected
* A few new "unicorn" rules 🎉 

We should really be looking to activate as many dormant rules as possible in `eslint.config.mjs` to keep us honest and safe. The linter is our friend.


knip.config.ts
```diff
@@ -111,9 +92,6 @@ const config: KnipConfig = {
             ignoreDependencies: [
                 // Declared to satisfy eslint-config-next peer dep; lint itself runs from root
                 "eslint",
```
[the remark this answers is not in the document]
dsargent: I sort of agree with Copilot here but if we don't provide it, yarn complains:

`@cognizant-ai-lab/neuro-san-web@workspace:apps/main doesn't provide @mui/system (p35b046), requested by @mui/x-tree-view.`

`knip` doesn't complain when we provide it either.

apps/main/package.json
```diff
@@ -33,10 +33,11 @@
         "sharp": "0.33.2"
     },
     "devDependencies": {
+        "@testing-library/react": "16.3.0",
+        "@testing-library/user-event": "14.6.1",
         "@types/lodash-es": "4.17.12",
         "@types/react": "19.2.14",
-        "eslint": "10.0.2",
-        "eslint-config-next": "16.1.6",
```
dsargent, line 39: `eslint-config-next` doesn't seem to be needed here. We get the same set of Next.js linter rules without it.

apps/main/package.json
```diff
@@ -33,10 +33,11 @@
         "sharp": "0.33.2"
     },
     "devDependencies": {
+        "@testing-library/react": "16.3.0",
```
dsargent, line 36: These are used by the package (in tests) and yarn complains about them not being declared.

build_scripts/eslint_list_rules.sh
dsargent, line 1: New script to do the thing we've discussed before -- generate the linter rules for each file type, for diff'ing against `main`.

Read inline comments for details.

Mostly vibed, some tweaking by me.

packages/ui-common/__tests__/components/AgentChat/ChatCommon.test.tsx
```diff
@@ -936,15 +936,11 @@ describe("ChatCommon", () => {
             [TEST_AGENT_MATH_GUY]: customPlaceholder,
         }
 
-        // eslint-disable-next-line testing-library/no-unnecessary-act
-        await act(async () => {
-            renderChatCommonComponent({
-                agentPlaceholders: customPlaceholders,
-            })
+        renderChatCommonComponent({
```
dsargent, line 939: Got rid of the whole "disable rule" and `act` thing by simply using `findBy...` which has a built-in `act`.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
dsargent, line 1: Modiied this file throughout to comply with the new rules:
1) Provide initial values for `Map` at construction time rather than two steps (unnecessary)
2) Add type declarations all over the place. This actually revealed a couple of violations where tests were using invalid fields. Will annotate inline below.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
```diff
@@ -89,17 +90,18 @@ describe("AgentFlow", () => {
         })
     })
 
+    const currentConversations2: AgentConversation[] = [
```
dsargent, line 93: Example -- pull this out as a separate variable so we can provide a type annotation.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
```diff
@@ -776,17 +778,21 @@ describe("AgentFlow", () => {
...
-            edge: {
-                id: "thought-bubble-test",
-                source: "agent1",
-                target: "agent2",
-                type: "thoughtBubbleEdge",
-                data: {text: "Test thought bubble"},
-            },
-            startedAt: Date.now(),
-        })
+        const thoughtBubbleEdgesMap = new Map<string, {edge: ThoughtBubbleEdgeShape; timestamp: number}>([
```
dsargent, line 781: Example: Initialize `Map` at construction time rather than "construct-then-set". There's now a linter rule about this, with the new upgrades.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
```diff
@@ -927,15 +937,17 @@ describe("AgentFlow", () => {
     })
 
     it("Should handle hover state changes for thought bubbles", () => {
+        const currentConversations: AgentConversation[] = [
+            {
+                id: "hover-test-conv",
+                agents: new Set(["agent1", "agent2"]),
+                startedAt: new Date(),
+                text: "Invoking Agent with inquiry: Hover test",
+                type: ChatMessageType.AGENT,
```
dsargent, line 946: Several of these test structures were missing `type` which is declared as required in the interface. Adding these type annotations caused that to show up as errors. It's always good to help `tsc` to help us.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
```diff
@@ -1120,25 +1134,29 @@ describe("AgentFlow", () => {
     })
 
     it("Should handle thought bubble edges without text field", () => {
-        const existingEdgesMap = new Map()
+        const existingEdgesMap = new Map<string, {edge: ThoughtBubbleEdgeShape; timestamp: number}>([
```
dsargent, line 1137: Initialize `Map` at construction time.

packages/ui-common/components/MultiAgentAccelerator/AgentCounts.ts
```diff
@@ -1,4 +1,4 @@
-import {Origin} from "../../generated/neuro-san/NeuroSanClient"
+import {Origin as NSOrigin} from "../../generated/neuro-san/NeuroSanClient"
```
dsargent, line 1: `Origin` now causes a shadowing violation, so rename it to avoid having to disable the rule here.

packages/ui-common/controller/llm/LlmChat.ts
```diff
@@ -36,7 +36,7 @@ const handleStreamingCallback = async (
     streamingUnit: StreamingUnit
 ) => {
     const reader = res.body.getReader()
-    const utf8decoder = new TextDecoder("utf8")
+    const utf8decoder = new TextDecoder("utf-8")
```
dsargent, line 39: New linter rule for names of encodings.

packages/ui-common/package.json
```diff
@@ -25,8 +25,10 @@
     "devDependencies": {
         "@babel/core": "7.23.7",
         "@babel/preset-env": "7.28.3",
+        "@testing-library/react": "16.3.0",
```
dsargent, line 28: yarn wants these here as they are used in tests in this package. Also bumped the node types to match the Node.js version we actually use.

eslint.config.mjs
```diff
@@ -1,78 +1,65 @@
-/*
```
dsargent, line 1: Removed out of date copyright. We have a ticket to update all these in one shot.

eslint.config.mjs
```diff
@@ -1,78 +1,65 @@
...
 import next from "@next/eslint-plugin-next"
-import typescriptEslint from "@typescript-eslint/eslint-plugin"
-// @ts-expect-error: parser has no types, but works
-import tsParser from "@typescript-eslint/parser"
+import {defineConfig, globalIgnores} from "eslint/config"
 import eslintConfigPrettier from "eslint-config-prettier/flat"
 import eslintPluginImport from "eslint-plugin-import"
-// eslint-disable-next-line no-shadow
-import jest from "eslint-plugin-jest"
+import eslintJest from "eslint-plugin-jest"
```
dsargent, line 9: Rename import to avoid shadow violation.

eslint.config.mjs
```diff
@@ -1,78 +1,65 @@
...
-See the License for the specific language governing permissions and
-limitations under the License.
-*/
+// @ts-check
 
 import {fixupPluginRules} from "@eslint/compat"
 import js from "@eslint/js"
 import next from "@next/eslint-plugin-next"
-import typescriptEslint from "@typescript-eslint/eslint-plugin"
-// @ts-expect-error: parser has no types, but works
```
dsargent, line 21: No longer causes an error.

eslint.config.mjs
```diff
@@ -1,78 +1,65 @@
...
+import typescriptEslint from "typescript-eslint"
 
-const config = [
-    preferArrowFunctions.configs["all"],
-    eslintPluginUnicorn.configs.all,
-    {
-        ignores: ["**/.next", "**/coverage", "**/generated", "**/embed", "**/dist", "**/babel.jest.config.cjs"],
-    },
+export default defineConfig([
+    globalIgnores(["**/.next", "**/coverage", "**/generated", "**/embed", "**/dist", "**/babel.jest.config.cjs"]),
```
dsargent: Moved this to first place as it's, well, global for the whole config,, and use the trendy ESLint `globalIgnores` function which does the same as we were doing but (arguably) more readable.

eslint.config.mjs
```diff
@@ -1,78 +1,65 @@
...
 import globals from "globals"
+import typescriptEslint from "typescript-eslint"
 
-const config = [
-    preferArrowFunctions.configs["all"],
-    eslintPluginUnicorn.configs.all,
-    {
-        ignores: ["**/.next", "**/coverage", "**/generated", "**/embed", "**/dist", "**/babel.jest.config.cjs"],
-    },
+export default defineConfig([
```
dsargent, line 19: All the cool kids, including the ESLint doc, recommend using `defineConfig` to help with type checking and find configuration issues...or something.

eslint.config.mjs
```diff
@@ -1,78 +1,65 @@
...
-import jest from "eslint-plugin-jest"
+import eslintJest from "eslint-plugin-jest"
 import jestDom from "eslint-plugin-jest-dom"
 import preferArrowFunctions from "eslint-plugin-prefer-arrow-functions"
 import eslintPluginReact from "eslint-plugin-react"
 import reactHooks from "eslint-plugin-react-hooks"
 import testingLibrary from "eslint-plugin-testing-library"
 import eslintPluginUnicorn from "eslint-plugin-unicorn"
 import globals from "globals"
+import typescriptEslint from "typescript-eslint"
```
dsargent, line 17: According to the doc, this is the new, recommended way of including this plugin.

The old `@typescript-eslint/stuff` plugins get pulled in by this "umbrella" plugin as needed...or something?! Not sure. The doc is far from clear.

eslint.config.mjs
```diff
@@ -1,78 +1,65 @@
...
+    globalIgnores(["**/.next", "**/coverage", "**/generated", "**/embed", "**/dist", "**/babel.jest.config.cjs"]),
 
     // This enables *all* base ESLint rules. We selectively disable those we are not yet ready for in the
     // "rules" section below.
     js.configs.all,
 
-    // See: https://nextjs.org/docs/pages/building-your-application/configuring/eslint
-    {rules: {...next.configs["recommended-legacy"].rules}},
-    {rules: {...next.configs["core-web-vitals-legacy"].rules}},
+    // See: https://github.com/typescript-eslint/typescript-eslint/issues/9724
```
dsargent, line 26: Throughout, I removed "rules spreading" in favor of the simpler, more modern flat config native support that most plugins are now providing.

However, there are some bumps in the road. See comments.

eslint.config.mjs
```diff
@@ -1,78 +1,65 @@
...
+    // and https://github.com/JamieMason/eslint-plugin-prefer-arrow-functions/pull/70
+    // @ts-expect-error see issues linked above
+    preferArrowFunctions.configs["all"],
+    eslintPluginUnicorn.configs.all,
 
-    // Equivalent to plugin:@typescript-eslint/all, expressed as flat config.
-    ...typescriptEslint.configs["flat/all"],
+    typescriptEslint.configs.all,
 
+    // Have to configure import plugin in the "legacy" way due to:
```
dsargent, line 34: Another one where I couldn't just use the native flat config. See comment.

eslint.config.mjs
```diff
@@ -1,78 +1,65 @@
...
         settings: {...eslintPluginImport.configs.typescript.settings},
     },
-    {rules: {...jest.configs.recommended.rules}},
-    {rules: {...reactHooks.configs.recommended.rules}},
-    {rules: {...eslintPluginReact.configs.all.rules}},
-    {rules: {...eslintPluginReact.configs["jsx-runtime"].rules}},
 
-    // Equivalent to extending "prettier" (kept here to preserve rule-order precedence), and then re-applied below
-    // exactly as in main.
+    eslintJest.configs["flat/recommended"],
```
dsargent, line 42: These ones _could_ be simplified to use the native flat config.

eslint.config.mjs
```diff
@@ -1,78 +1,65 @@
...
+    // process and doesn't *need* to be in scope.
+    eslintPluginReact.configs.flat["jsx-runtime"],
+
     eslintConfigPrettier,
 
     {
         plugins: {
-            "@next/next": next,
-            "@typescript-eslint": typescriptEslint,
+            "typescript-eslint": typescriptEslint.plugin,
```
dsargent, line 55: Some changes here since some no longer need a separate "plugins" line item, as their flat config already provides that.

eslint.config.mjs
```diff
@@ -1,78 +1,65 @@
...
             "jest-dom": jestDom,
-            "react-hooks": reactHooks,
+            "@next/next": next,
+            // Have to configure import plugin in the "legacy" way due to:
+            // https://github.com/import-js/eslint-plugin-import/issues/3227
```
dsargent, line 59: If it weren't for this one pesky plugin, we could get rid of the fixup/compat dependency.

eslint.config.mjs
```diff
@@ -106,7 +92,11 @@ const config = [
 
         settings: {
             react: {
-                version: "detect",
+                // If we use "detect" here, we get this error:
```
dsargent, line 95: This made me sad, but known issues with the Next.js plugins. See link. We should revisit these "warts" later.

eslint.config.mjs
```diff
@@ -121,6 +111,8 @@ const config = [
         },
 
         rules: {
+            ...next.configs.recommended.rules,
```
dsargent, line 114: This seems to be the recommended way to configure the Next.js linter rules (which we barely even care about...in fact we barely care about Next.js itself?!) So nore more "legacy", no more "web vitals" stuff...I guess?! The proof of the pudding is in the eating, and the proof of linter config is in the rules output vs. `main`, which I attached to the PR description.

eslint.config.mjs
```diff
@@ -244,6 +236,7 @@ const config = [
             "prefer-spread": "error",
             "preserve-caught-error": "warn",
             "react/jsx-curly-brace-presence": "error",
+            "react/no-multi-comp": ["error", {ignoreStateless: false}],
```
dsargent, line 239: Added config param to make it behave the same as `main`. Don't want to delve into the why's and wherefores, just trying to make sure we're " ≥ `main`".

eslint.config.mjs
```diff
@@ -273,11 +266,11 @@ const config = [
             ],
 
             "import/no-extraneous-dependencies": [
-                "off",
+                "error",
                 {
                     includeInternal: true,
                     includeTypes: true,
-                    devDependencies: ["__tests__/**", "jest*.*", "eslint.config.mjs"],
+                    devDependencies: ["**/__tests__/**", "jest*.*", "eslint.config.mjs", "knip.config.ts"],
```
dsargent, line 273: This was the fix -- the recursive glob for `__tests__`.

eslint.config.mjs
```diff
@@ -429,9 +422,15 @@ const config = [
...
             "react-hooks/exhaustive-deps": "off",
+            "react-hooks/preserve-manual-memoization": "off",
+            "react-hooks/purity": "off",
+            "react-hooks/refs": "off",
+            "react-hooks/set-state-in-effect": "off",
             "react/button-has-type": "off",
             "react/destructuring-assignment": "off",
             "react/forbid-component-props": "off",
+            // Causes an error: "context.getSourceCode is not a function" -- ESLint 10 issue?
+            "react/forward-ref-uses-ref": "off",
```
dsargent, line 433: Sad face...but couldn't find a fix for it. Seems like a bit of an obscure rule so not the end of the world.

eslint.config.mjs
```diff
@@ -472,6 +471,7 @@ const config = [
             "unicorn/prefer-string-slice": "off",
             "unicorn/no-named-default": "off",
             "unicorn/no-array-reduce": "off",
+            "unicorn/no-array-sort": "off",
```
dsargent, line 474: We're not quite ready for this yet. We have a few violations that require attention.

jest.setup.ts
```diff
@@ -19,7 +19,8 @@ import "@testing-library/jest-dom"
 import failOnConsole from "jest-fail-on-console"
 // eslint-disable-next-line no-shadow
 import {ReadableStream} from "node:stream/web"
-import {createElement, ReactNode} from "react"
+import type {ReactNode} from "react"
```
dsargent, line 22: New linter rule -- type-only imports have to be imported via `import type`.

knip.config.ts
```diff
@@ -33,13 +33,14 @@ const config: KnipConfig = {
     // Opt-in to all issues types
     include: [
         "binaries",
-        "classMembers",
+        "catalog",
```
dsargent, line 36: Upgraded `knip` has slightly different issues types. I don't know why they did this but it's in their doc so I fell in line.

knip.config.ts
```diff
@@ -49,26 +50,12 @@ const config: KnipConfig = {
     ],
 
     ignore: [
-        // Temporarily unused pending migration from Pinecone to pgvector
```
dsargent, line 52: New knip is a lot smarter so I was able to remove a lot of these ignores.
Though the ChatBot one makes me a little 🤨 It's only used in tests -- so why doesn't knip complain about it? Maybe only if you run knip in "production mode", which we never do?

In this very config file we have `treatConfigHintsAsErrors: true` so `knip` will not only give us "suggestions" about things it thinks are unnecessary in the config file, it will make them hard errors! I basically removed everything it said is not needed in here.

package.json
```diff
@@ -19,55 +19,50 @@
         "dev": "yarn workspace @cognizant-ai-lab/neuro-san-web run dev",
         "generate": "rm -rf packages/ui-common/generated && cd packages/ui-common && ./build_scripts/do_openapi_generate.sh",
         "lint": "eslint",
-        "lint-rules": "eslint --print-config jest.config.ts",
+        "lint-rules": "./build_scripts/eslint_list_rules.sh",
```
dsargent, line 22: Modified to call the snazzy script that aggregates rules for all the file types.

package.json
```diff
@@ -19,55 +19,50 @@
...
         "lint-summary": "eslint --format summary",
         "start": "yarn workspace @cognizant-ai-lab/neuro-san-web start",
         "test": "jest --config ./jest.config.ts"
     },
     "packageManager": "yarn@4.12.0",
     "devDependencies": {
-        "@babel/core": "7.23.7",
-        "@babel/preset-env": "7.28.3",
-        "@eslint/compat": "2.0.2",
+        "@babel/core": "7.29.0",
```
dsargent, line 29: Upgrade the heck out of the dev deps.

tsconfig.json
dsargent, line 1: Now under the purview of `prettier`. Since we've made peace with Next.js in this file and Next.js _seems_ to leave it alone nowadays, I'm tentatively un-disabling `prettier`. That means a lot of formatting changes so "disable whitespace" in the PR to see more clearly.

tsconfig.json
```diff
@@ -1,56 +1,44 @@
...
+        "checkJs": true,
+        "declaration": false,
+        "esModuleInterop": true,
+        "forceConsistentCasingInFileNames": true,
+        "isolatedModules": true,
+        "jsx": "react-jsx",
+        "lib": ["dom", "dom.iterable", "es2022"],
+        "module": "esnext",
+        "moduleDetection": "force",
+        "moduleResolution": "bundler",
```
dsargent, line 15: This is what we're using in Neuro UI, it seems to be what the LLMs and the vendor doc recommends. Open to discussion.

tsconfig.json
```diff
@@ -1,56 +1,44 @@
...
+    "compilerOptions": {
+        "allowJs": true,
+        "allowUnreachableCode": false,
+        "allowUnusedLabels": false,
+        "checkJs": true,
+        "declaration": false,
+        "esModuleInterop": true,
+        "forceConsistentCasingInFileNames": true,
+        "isolatedModules": true,
+        "jsx": "react-jsx",
```
dsargent, line 11: After lots of back and forth with contradictory Next.js doc, and hallucinating LLMs, and ancient StackOverflow posts, this _seems_ to be the modern and correct setting for JSX transformation. Open to discussion.

Also now matches what Neuro UI does.

tsconfig.json
```diff
@@ -1,56 +1,44 @@
...
-    ".next/types/**/*.ts",
-    "next-env.d.ts"
-  ],
-  "exclude": [ 
-    "node_modules",
-    "**/.next",
-    "**/coverage",
-    "**/dist"
-  ]
+    "exclude": ["node_modules", "**/.next", "**/coverage", "**/dist"]
```
dsargent, line 43: `prettier` formatting changes.

eslint.config.mjs
```diff
@@ -429,9 +422,15 @@ const config = [
             "prefer-destructuring": "off",
             radix: "off",
             "react-hooks/exhaustive-deps": "off",
+            "react-hooks/preserve-manual-memoization": "off",
```
dsargent, line 425: Would be _very_ nice to enable these various "React health" rules. I think a lot of bugs are due to misue of effects, rendering race conditions etc. and these rules might help prevent that/ferret out where we already have issues.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
```diff
@@ -776,17 +778,21 @@ describe("AgentFlow", () => {
...
+                "test-edge",
+                {
+                    edge: {
+                        id: "thought-bubble-test",
+                        source: "agent1",
+                        target: "agent2",
+                        type: "thoughtBubbleEdge",
+                        data: {text: "Test thought bubble"},
+                    },
+                    timestamp: Date.now(),
```
dsargent, line 792: Example of a type violation that was found by adding correct type annotations.

It was using `startedAt`, which does not match the definition of `thoughtBubbleEdges`:

```typescript
 readonly thoughtBubbleEdges: Map<string, {edge: ThoughtBubbleEdgeShape; timestamp: number}>
```

It should be `timestamp`, not `startedAt` (confusion with the "conversations" type, probably? Which _does_ have a member named `startedAt`.) It was not being detected because without type annotations, `tsc` had no way to know it was wrong.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
```diff
@@ -808,25 +814,29 @@ describe("AgentFlow", () => {
...
+                        source: "agent1",
+                        target: "agent2",
+                        type: "thoughtBubbleEdge",
+                        data: {
+                            text: '{"inquiry": "What is the weather?"}',
+                            showAlways: true,
+                            conversationId: "conv-1",
+                        },
+                    },
+                    timestamp: Date.now(),
```
dsargent, line 832: Another incorrect `startedAt` field.

packages/ui-common/__tests__/unit/AgentNetwork/GraphLayouts.test.ts
```diff
@@ -213,9 +213,10 @@ describe("GraphLayouts", () => {
             {origin: "b", tools: []},
         ]
 
-        const agentCounts = new Map<string, number>()
-        agentCounts.set("a", 1)
-        agentCounts.set("b", 1)
+        const agentCounts = new Map<string, number>([
```
dsargent, line 216: More initializing maps at construction time as required by the new linter rule.

eslint.config.mjs
```diff
@@ -94,7 +81,6 @@ const config = [
                 SpeechRecognitionErrorEvent: "readonly",
             },
 
-            parser: tsParser,
```
dsargent, line 97: No longer needed here. I guess the Typescript ESLint plugin sets this up automatically.

eslint.config.mjs
```diff
@@ -532,7 +532,6 @@ const config = [
             "react/no-multi-comp": "off",
         },
     },
-    eslintConfigPrettier,
```
dsargent, line 535: I don't think we need this second invocation. Works fine without.

build_scripts/eslint_list_rules.sh
```diff
@@ -0,0 +1,58 @@
...
+if ! command -v jq &> /dev/null; then
+    echo "jq is required but not installed. Please install jq to use this script."
+    exit 1
+fi
+
+echo "{"
+first=true
+for ext in "${FILE_TYPES[@]}"; do
+    # The inline function fed to jq sorts the keys of the JSON object recursively, to reduce diff noise when comparing.
+    config=$(yarn eslint --print-config "dummy.$ext" | jq '."rules" | def sortkeys_recursive:
```
dsargent, line 38: We pull out the `rules` part only because the output from `--print-config` has a ton of stuff not relevant for the job at hand.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
dsargent: (no, appending an `...as Foo` does not do the same thing _at all_. That is the type-unsafe "trust me bro" version.)

eslint.config.mjs
```diff
@@ -1,78 +1,65 @@
...
+            "typescript-eslint": typescriptEslint.plugin,
             "jest-dom": jestDom,
-            "react-hooks": reactHooks,
+            "@next/next": next,
+            // Have to configure import plugin in the "legacy" way due to:
+            // https://github.com/import-js/eslint-plugin-import/issues/3227
             import: fixupPluginRules(eslintPluginImport),
-            jest,
-            react: fixupPluginRules(eslintPluginReact),
+            jest: eslintJest,
```
dsargent, line 61: Renamed `jest` plugin to avoid shadowing violation.

eslint.config.mjs
```diff
@@ -273,11 +266,11 @@ const config = [
             ],
 
             "import/no-extraneous-dependencies": [
-                "off",
+                "error",
```
dsargent, line 269: Re-activate this one 🎉

eslint.config.mjs
swensel: In React 19 `forwardRef` is no longer necessary anyway. It will be deprecated in a future release - https://react.dev/reference/react/forwardRef

swensel (approved):
LGTM. Nice work 🎉

Issue #270, opened 2026-04-23
Enable react-hooks/exhaustive-deps and fix issues


dsargent:
Code hygiene measure.

PR #271, opened 2026-04-23
#270 Enable react-hooks/exhaustive-deps and fix issues

dsargent:
[Copilot summary](https://github.com/cognizant-ai-lab/neuro-san-ui/pull/271#pullrequestreview-4166752640).

Actually here Copilot did a good job of summarizing. The idea was to enable the linter rule and fix all resulting violations, while in general improving our "React.js hygiene". I suspect we have numerous minor and not-so-minor bugs relating to incorrect callbacks, memoization, and dependency arrays. This rule should help clean that up and ensure it _remains_ clean.

### DIFF
--- a/apps/main/pages/_app.tsx
+++ b/apps/main/pages/_app.tsx
@@ -105,7 +105,7 @@ export default function NeuroSanUI({Component, pageProps}: ExtendedAppProps): Re
         supportEmailAddress,
     } = useEnvironmentStore()
 
-    // Access NextJS router
+    // Access Next.js router
     const {pathname, query} = useRouter()
 
     // Access user info store
@@ -181,7 +181,7 @@ export default function NeuroSanUI({Component, pageProps}: ExtendedAppProps): Re
         }
 
         void getEnvironment()
-    }, [])
+    }, [setAuth0ClientId, setAuth0Domain, setBackendNeuroSanApiUrl, setLogoServiceToken, setSupportEmailAddress])
 
     useEffect(() => {
         const getUserInfo = async () => {
@@ -225,7 +225,7 @@ export default function NeuroSanUI({Component, pageProps}: ExtendedAppProps): Re
         }
 
         void getUserInfo()
-    }, [])
+    }, [setCurrentUser, setOidcProvider, setPicture])
 
     const handleSignOut = async () => {
         // Clear our state storage variables
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -300,6 +300,8 @@ export default defineConfig([
                 },
             ],
 
+            "react-hooks/exhaustive-deps": "error",
+
             // Rules we're not ready to enable yet
 
             // Requires strict type checks enabled in tsc which we're not ready for yet
@@ -421,7 +423,6 @@ export default defineConfig([
             "one-var": "off",
             "prefer-destructuring": "off",
             radix: "off",
-            "react-hooks/exhaustive-deps": "off",
             "react-hooks/preserve-manual-memoization": "off",
             "react-hooks/purity": "off",
             "react-hooks/refs": "off",
--- a/packages/ui-common/__tests__/components/AgentChat/ControlButtons.test.tsx
+++ b/packages/ui-common/__tests__/components/AgentChat/ControlButtons.test.tsx
@@ -29,7 +29,7 @@ describe("ControlButtons", () => {
     const mockHandleStop = jest.fn()
 
     const defaultProps = {
-        clearChatOnClickCallback: mockClearChat,
+        handleClearChat: mockClearChat,
         enableClearChatButton: true,
         isAwaitingLlm: false,
         handleSend: mockHandleSend,
@@ -58,7 +58,7 @@ describe("ControlButtons", () => {
         expect(screen.getByRole("button", {name: "Stop"})).toBeInTheDocument()
     })
 
-    it("calls clearChatOnClickCallback when Clear Chat button is clicked", async () => {
+    it("calls handleClearChat when Clear Chat button is clicked", async () => {
         render(<ControlButtons {...defaultProps} />)
         await user.click(screen.getByRole("button", {name: "Clear Chat"}))
         expect(mockClearChat).toHaveBeenCalledTimes(1)
--- a/packages/ui-common/components/AgentChat/ChatCommon.tsx
+++ b/packages/ui-common/components/AgentChat/ChatCommon.tsx
@@ -244,7 +244,7 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
     const [chatInput, setChatInput] = useState<string>("")
 
     // Previous user query (for "regenerate" feature)
-    const [previousUserQuery, setPreviousUserQuery] = useState<string>("")
+    const previousUserQuery = useRef<string>("")
 
     // Chat output window contents
     const [chatOutput, setChatOutput] = useState<ReactNode[]>([])
@@ -601,7 +601,10 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
         currentResponse.current = ""
     }, [setIsAwaitingLlm, setPreviousResponse, targetAgent])
 
-    const doQueryLoop = useCallback(
+    /*
+     * The main logic for sending a query to the server, with retries on errors.
+     */
+    const doRetryLoop = useCallback(
         async (query: string) => {
             succeeded.current = false
 
@@ -667,7 +670,7 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
                     }
                 }
             } while (attemptNumber < MAX_AGENT_RETRIES && !succeeded.current)
-            return {wasAborted}
+            return wasAborted
         },
         [currentUser, extraParams, handleChunk, legacyAgentEndpoint, neuroSanURL, targetAgent, updateOutput]
     )
@@ -675,14 +678,14 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
     const handleSend = useCallback(
         async (query: string) => {
             // Record user query in chat history
-            chatHistory.current = [...chatHistory.current, new HumanMessage(previousUserQuery)]
+            chatHistory.current = [...chatHistory.current, new HumanMessage(previousUserQuery.current)]
 
             // Allow parent to intercept and modify the query before sending if needed
             const queryToSend = onSend?.(query) ?? query
 
             // Save query for "regenerate" use. Again we save the real user input, not the modified query. It will again
             // get intercepted and re-modified (if applicable) on "regenerate".
-            setPreviousUserQuery(query)
+            previousUserQuery.current = query
 
             setIsAwaitingLlm(true)
 
@@ -729,7 +732,8 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
                 )
             }
             try {
-                const {wasAborted} = await doQueryLoop(queryToSend)
+                // Invoke the logic to send the request and retry as necessary
+                const wasAborted = await doRetryLoop(queryToSend)
 
                 if (!wasAborted && !succeeded.current) {
                     updateOutput(
@@ -777,11 +781,10 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
         [
             agentDisplayName,
             currentUser,
-            doQueryLoop,
+            doRetryLoop,
             onSend,
             onStreamingComplete,
             onStreamingStarted,
-            previousUserQuery,
             processLogLine,
             resetState,
             setIsAwaitingLlm,
@@ -856,10 +859,6 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
     }, [targetAgent, introduceAgent])
 
     useEffect(() => {
-        if (!targetAgent || isLegacyAgentType(targetAgent)) {
-            return
-        }
-
         const fetchAgentDetails = async () => {
             let agentFunction: FunctionResponse
 
@@ -911,10 +910,13 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
             }
         }
 
-        void fetchAgentDetails()
+        if (targetAgent && !isLegacyAgentType(targetAgent)) {
+            void fetchAgentDetails()
+        }
     }, [
         agentDisplayName,
         currentUser,
+        handleSend,
         id,
         neuroSanURL,
         renderConnectivityInfo,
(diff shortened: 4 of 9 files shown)


apps/main/pages/_app.tsx
```diff
@@ -105,7 +105,7 @@ export default function NeuroSanUI({Component, pageProps}: ExtendedAppProps): Re
         supportEmailAddress,
     } = useEnvironmentStore()
 
-    // Access NextJS router
+    // Access Next.js router
```
dsargent, line 108: Branding

apps/main/pages/_app.tsx
```diff
@@ -181,7 +181,7 @@ export default function NeuroSanUI({Component, pageProps}: ExtendedAppProps): Re
         }
 
         void getEnvironment()
-    }, [])
+    }, [setAuth0ClientId, setAuth0Domain, setBackendNeuroSanApiUrl, setLogoServiceToken, setSupportEmailAddress])
```
dsargent, line 184: Add required deps to dependency array. A lot of the changes in this PR are about this so I won't annotate all of them.

packages/ui-common/__tests__/components/AgentChat/ControlButtons.test.tsx
```diff
@@ -29,7 +29,7 @@ describe("ControlButtons", () => {
     const mockHandleStop = jest.fn()
 
     const defaultProps = {
-        clearChatOnClickCallback: mockClearChat,
+        handleClearChat: mockClearChat,
```
dsargent, line 32: Felt like an over-literal LLM name so I simplified and clarified it.

packages/ui-common/components/AgentChat/ChatCommon.tsx
dsargent, line 1: Big ugly diffs here. Two reasons --

1) Added required deps to all effects and memoizations
2) Due to (1), had to move things around so they appear in topological dependency order in the source. For example, if function `foo` now depends on `bar` where it didn't before, well, `bar` then has to appear _before_ `foo` in the source, per JS/TS scope rules. This is an improvement because now the source explicitly reflects the dependency structure and the order that things have to appear.

It does mean that the diff is a confusing mess. Disabling whitespace helps a _little_. But overall GitHub just gets super confused by moved blocks and shows everything as different. You might want to diff locally if you have a preferred diff tool that is smarter than GitHub's and especially one that is good at detecting moved blocks.

packages/ui-common/components/AgentChat/ChatCommon.tsx
```diff
@@ -197,6 +199,14 @@ const MAX_SAMPLE_QUERIES = 5
 // Maximum length of query to show in sample query chips
 const QUERY_TRUNCATE_LENGTH = 80
 
+/**
```
dsargent, line 202: This is a pure function that doesn't depend on anything, so moved it to global scope here.

packages/ui-common/components/AgentChat/ChatCommon.tsx
```diff
@@ -226,12 +236,6 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
         onClose,
         neuroSanURL,
     } = props
-
-    // Expose the handleStop method to parent components via ref for external control (e.g., to cancel chat requests)
```
dsargent, line 230: Had to move this to below `handleStop` because it depends on it!

packages/ui-common/components/AgentChat/ChatCommon.tsx
```diff
@@ -240,7 +244,7 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
     const [chatInput, setChatInput] = useState<string>("")
 
     // Previous user query (for "regenerate" feature)
-    const [previousUserQuery, setPreviousUserQuery] = useState<string>("")
+    const previousUserQuery = useRef<string>("")
```
dsargent, line 247: Changed this to a ref to ease the dependency chain. I think it's the right thing to do; we don't want to trigger rerenders just because we saved the user's previous query.

packages/ui-common/components/AgentChat/ChatCommon.tsx
```diff
@@ -354,15 +358,7 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
         if (autoScrollEnabledRef.current && chatOutputRef?.current) {
             chatOutputRef.current.scrollTop = chatOutputRef.current.scrollHeight
         }
-    }, [chatOutput])
-
-    useEffect(() => {
```
dsargent, line 359: Not needed -- everything gets refreshed/remounted when the user changes this URL due to other "effects".

packages/ui-common/components/AgentChat/ChatCommon.tsx
```diff
@@ -376,85 +372,164 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
-    }
+                            : "none",
+                    }}
+                />
+            )
+        },
+        [a11yLight, atelierDuneDark, darkMode, shadowColor, shouldWrapOutput]
+    )
 
-    const agentDisplayName = cleanUpAgentName(removeTrailingUuid(targetAgent))
```
dsargent, line 455: Confusing diffs due to blocks being moved around to respect dependency graph.

packages/ui-common/components/AgentChat/ChatCommon.tsx
```diff
@@ -376,85 +372,164 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+    /**
+     * Handles adding content to the output window.
+     * @param node A ReactNode to add to the output window -- text, spinner, etc. but could also be  simple string
+     * @returns Nothing, but updates the output window with the new content. Updates currentResponse as a side effect.
+     */
+    const updateOutput = useCallback((node: ReactNode) => {
```
dsargent, line 454: I added a lot of memoization to ensure dependences are stable, since we are now being strict about dependency arrays.

Some day this may all be invisible as `react-compiler` claims to automate most or all of this memoization.

packages/ui-common/components/AgentChat/ChatCommon.tsx
```diff
@@ -376,85 +372,164 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+                    chatMessage.origin?.length > 0
+                        ? cleanUpAgentName(chatMessage.origin[chatMessage.origin.length - 1].tool)
+                        : "Agent message"
+                updateOutput(processLogLine(chatMessage.text, agentName, chatMessage.type))
+            }
+        },
+        [onChunkReceived, processLogLine, targetAgent, updateOutput]
+    )
+
+    const agentDisplayName = useMemo(() => cleanUpAgentName(removeTrailingUuid(targetAgent)), [targetAgent])
```
dsargent, line 530: This is used in several places, so i calculate it once here and memoize it.

packages/ui-common/components/AgentChat/ChatCommon.tsx
```diff
@@ -469,43 +544,255 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+
+        // Get agent name, either from the enum (Neuro-san) or from the targetAgent string directly (legacy)
+        setPreviousResponse?.(targetAgent, currentResponse.current)
+        currentResponse.current = ""
+    }, [setIsAwaitingLlm, setPreviousResponse, targetAgent])
+
+    /*
+     * The main logic for sending a query to the server, with retries on errors.
+     */
+    const doRetryLoop = useCallback(
```
dsargent, line 607: Renamed this function -- `doQueryLoop` -> `doRetryLoop` to better express its purpose.

packages/ui-common/components/AgentChat/ChatCommon.tsx
```diff
@@ -469,43 +544,255 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+                            >
+                                {`Error occurred: ${error}`}
+                            </MUIAlert>
+                        )
+                    }
+                }
+            } while (attemptNumber < MAX_AGENT_RETRIES && !succeeded.current)
+            return wasAborted
+        },
+        [currentUser, extraParams, handleChunk, legacyAgentEndpoint, neuroSanURL, targetAgent, updateOutput]
```
dsargent, line 675: Lengthy deps array here, but, required by rules of React.js. 

All these extra deps in the deps arrays may cause more re-renders; we should keep our eyes out for any performance degradations and options for further memoization.

packages/ui-common/components/AgentChat/ChatCommon.tsx
```diff
@@ -469,43 +544,255 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+                    userQuery={query}
+                    title={currentUser}
+                    userImage={userImage}
+                />
+            )
+
+            // Add ID block for agent
+            updateOutput(
+                <UserQueryDisplay
+                    userQuery={agentDisplayName}
```
dsargent, line 707: Use memoized version

packages/ui-common/components/AgentChat/ChatCommon.tsx
```diff
@@ -469,43 +544,255 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+                    chatHistory.current = [...chatHistory.current, new AIMessage(currentResponse.current)]
+                }
+            } finally {
+                resetState()
+
+                // Allow parent components to do something when streaming is complete
+                onStreamingComplete?.()
+            }
+        },
+        [
```
dsargent, line 781: A lot of deps!! But they are all used in the function so they _should_ be in the deps array.

packages/ui-common/components/AgentChat/ControlButtons.tsx
```diff
@@ -23,7 +23,7 @@ import {SmallLlmChatButton} from "./LlmChatButton"
 
 // #region: Types
 interface ControlButtonsProps {
-    clearChatOnClickCallback: () => void
+    handleClearChat: () => void
```
dsargent, line 26: Renamed to (IMO) a more meaningful name.

packages/ui-common/components/MultiAgentAccelerator/Sidebar/Sidebar.tsx
```diff
@@ -244,18 +243,17 @@ export const Sidebar: FC<SidebarProps> = ({
         setConnectionStatus(CONNECTION_STATUS.IDLE)
     }
 
-    // Get Neuro-san version on initial load
+    // Fetch Neuro-san version on load and whenever the saved URL changes
```
dsargent, line 246: This effect was a bit broken. It was depending on urlInput, which could be stale, and not declaring it as a dep. This runs afoul of the `exhausive-deps` rule.

So now it declares the items it actually uses as a dependency. When they change from "nothing" to "something" on page load, this will fire and set the version info. When the user enters a URL manually and clicks `Test`, `handleTestConnection` will take care of setting the version info.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -97,8 +97,8 @@ export const AgentFlow: FC<AgentFlowProps> = ({
     const {fitView} = useReactFlow()
 
     const handleResize = useCallback(() => {
-        fitView() // Adjusts the view to fit after resizing
-    }, [fitView, isAwaitingLlm])
+        void fitView() // Adjusts the view to fit after resizing
```
dsargent, line 100: I like to add the `void` when we're ignoring the `Promise` (1) because my IDE nags if I don't and (2) it makes explicit our attention -- "yeah I know it returns a `Promise`, I don't care, I want to ignore it."

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -208,7 +208,7 @@ export const AgentFlow: FC<AgentFlowProps> = ({
         }, 1000)
 
         return () => clearInterval(cleanupInterval)
-    }, []) // mount/unmount only
+    }, [setThoughtBubbleEdges]) // mount/unmount only
```
dsargent, line 211: Missing dependency

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -265,13 +265,12 @@ export const AgentFlow: FC<AgentFlowProps> = ({
         [
             agentCounts,
             agentIconSuggestions,
-            coloringOption,
```
dsargent, line 268: Not used, so invalid dependency

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -265,13 +265,12 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
             currentConversations,
             isAgentNetworkDesignerMode,
             isAwaitingLlm,
+            isHeatmap,
```
dsargent, line 271: _Is_ used, so should be  a dep

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -265,13 +265,12 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
             layout,
             mergedAgentsInNetwork,
-            showThoughtBubbles,
```
dsargent, line 274: Ditto, not used. The linter rule complained.

packages/ui-common/components/MultiAgentAccelerator/ThoughtBubbleOverlay.tsx
```diff
@@ -169,15 +169,16 @@ export const ThoughtBubbleOverlay: FC<ThoughtBubbleOverlayProps> = ({
     // Handle bubble lifecycle (appear/disappear animations)
     useEffect(() => {
         const currentEdgeIds = new Set(thoughtBubbleEdges.map((e) => e.id))
-        const previousBubbleIds = new Set(bubbleStates.keys())
```
dsargent, line 172: This effect was a bit broken too. It was closing over some potentially stale variables since it wasn't declaring `bubbleStates` as a dependency; this is an example of the bugs that `exhaustive-deps` helps to surface.

Instead, we use the fact that the effect is already using the functional value of the setter which gives us the live, previous value, and get the previous IDs from there. This is faster, and prevents unnecessary rerenders since we don't have to "depend on" `bubbleStates`.

This may be a cause of some of the bugs we've seen around thought bubbles.

packages/ui-common/components/MultiAgentAccelerator/ThoughtBubbleOverlay.tsx
```diff
@@ -219,9 +220,10 @@ export const ThoughtBubbleOverlay: FC<ThoughtBubbleOverlayProps> = ({
 
     // Cleanup timeouts on unmount
     useEffect(() => {
+        const timeouts = animationTimeouts.current
```
dsargent, line 223: Extract common part

eslint.config.mjs
```diff
@@ -300,6 +300,8 @@ export default defineConfig([
                 },
             ],
 
+            "react-hooks/exhaustive-deps": "error",
```
dsargent, line 303: This enables the rule. With this PR, it is now live, and we are 100% compliant in this repo.

Neuro AI is another matter..

swensel (approved):
LGTM. Nice PR. Thanks @dsargent

PR #272, opened 2026-04-24
Check both chat_context and sly_data in AGENT_FRAMEWORK message.

swensel:
# Description

Check both chat_context and sly_data in AGENT_FRAMEWORK message.

This fixes topological editing for AND.

### DIFF
--- a/packages/ui-common/components/AgentChat/ChatCommon.tsx
+++ b/packages/ui-common/components/AgentChat/ChatCommon.tsx
@@ -717,19 +717,23 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
             return
         }
 
-        // It's a ChatMessage. Does it have chat context? Only AGENT_FRAMEWORK messages can have chat context.
-        if (chatMessage.type === ChatMessageType.AGENT_FRAMEWORK && chatMessage.chat_context) {
-            // Save the chat context, potentially overwriting any previous ones we received during this session.
-            // We only care about the last one received.
-            chatContext.current = chatMessage.chat_context
+        // It's a ChatMessage. Does it have chat context? Only AGENT_FRAMEWORK messages can have chat context and
+        // slyData.
+        if (chatMessage.type === ChatMessageType.AGENT_FRAMEWORK) {
+            if (chatMessage.chat_context) {
+                // Save the chat context, potentially overwriting any previous ones we received during this session.
+                // We only care about the last one received.
+                chatContext.current = chatMessage.chat_context
+            }
+
+            if (chatMessage.sly_data) {
+                // Save the slyData, potentially overwriting any previous ones we received during this session.
+                // We only care about the last one received.
+                slyData.current = {...slyData.current, ...chatMessage.sly_data}
+            }
 
             // Nothing more to do with this message. It's just a message to give us the chat context, so return
             return
-        }
-
-        // Merge slyData.current with incoming chatMessage.sly_data
-        if (chatMessage.sly_data) {
-            slyData.current = {...slyData.current, ...chatMessage.sly_data}
         }
 
         // Check if there is an error block in the "structure" field of the chat message.
--- a/packages/ui-common/components/AgentChat/ChatCommon.tsx
+++ b/packages/ui-common/components/AgentChat/ChatCommon.tsx
@@ -717,17 +717,16 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
             return
         }
 
-        // It's a ChatMessage. Does it have chat context? Only AGENT_FRAMEWORK messages can have chat context and
-        // slyData.
+        // Only AGENT_FRAMEWORK messages can have chat_context and sly_data.
         if (chatMessage.type === ChatMessageType.AGENT_FRAMEWORK) {
             if (chatMessage.chat_context) {
-                // Save the chat context, potentially overwriting any previous ones we received during this session.
+                // Save the chat_context, potentially overwriting any previous ones we received during this session.
                 // We only care about the last one received.
                 chatContext.current = chatMessage.chat_context
             }
 
             if (chatMessage.sly_data) {
-                // Save the slyData, potentially overwriting any previous ones we received during this session.
+                // Save the sly_data, potentially overwriting any previous ones we received during this session.
                 // We only care about the last one received.
                 slyData.current = {...slyData.current, ...chatMessage.sly_data}
             }


packages/ui-common/components/AgentChat/ChatCommon.tsx
```diff
@@ -717,21 +717,25 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+        // It's a ChatMessage. Does it have chat context? Only AGENT_FRAMEWORK messages can have chat context and
+        // slyData.
+        if (chatMessage.type === ChatMessageType.AGENT_FRAMEWORK) {
+            if (chatMessage.chat_context) {
+                // Save the chat context, potentially overwriting any previous ones we received during this session.
+                // We only care about the last one received.
+                chatContext.current = chatMessage.chat_context
+            }
+
+            if (chatMessage.sly_data) {
```
dsargent, line 728: This is another behavior change and I'm not sure if it's intentional.

With the logic you have here, we will ignore `sly_data` in all messages other than `AGENT_FRAMEWORK`.

I would guess that we actually want to _always_ send back `sly_data`, whatever the message type, after deep-merging what we have with the `sly_data` in the incoming message.

Noravee: Only the `AGENT_FRAMEWORK` will have `sly_data`.

Noravee:
I think it will work for both structural and instructions edit. Note that most structural edits will be followed by instructions edit.

packages/ui-common/components/AgentChat/ChatCommon.tsx
dsargent: Maybe for AND, but not true in general. Remember MAUI is supposed to handle _any_ kind of traffic from Neuro-san.

For example, just in a single recorded interaction with Neuro-san I see sly_data in messages of type `SYSTEM`, `HUMAN` and `AGENT`.

[santa.json.zip](https://github.com/user-attachments/files/27032747/santa.json.zip)

dsargent:
> I think it will work for both structural and instructions edit. Note that most structural edits will be followed by instructions edit.

So the user could send a prompt to AND itself to modify a single agent in the network? That's something we should try. The way @d1donlydfink said to do it was to edit the "instructions" field on the individual node.

Noravee:
You can say something like "set instructions of 'agent_A' to 'blah'" and `AND` will do it. However, it is easier if you can just manually type in what you want instead of waiting for `AND` to do it.

packages/ui-common/components/AgentChat/ChatCommon.tsx
Noravee: To clarify: each message has two main fields, `request` and `response`, both of which can contain `sly_data`.

- **`request.sly_data`** is what the client sends (in MAUI's case, the login name).
- **`response.sly_data`** is what the agent returns.

When I said only `AGENT_FRAMEWORK` has `sly_data`, I was referring specifically to `response.sly_data` — which is the one we care about. The correct behavior is to merge `response.sly_data` from the current message into `request.sly_data` to produce the `sly_data` for the next request.

On the merge strategy: I think this should be a **shallow merge**, not a deep merge. Here's why:

```
# Shallow merge (correct)
request:  {login: abc, agent_network_def: {agent_a: {...}, agent_b: {...}}}
response: {login: abc, agent_network_def: {agent_a: {...}, agent_c: {...}}}
result:   {login: abc, agent_network_def: {agent_a: {...}, agent_c: {...}}} 
✅ agent_b replaced by agent_c

# Deep merge (incorrect)
request:  {login: abc, agent_network_def: {agent_a: {...}, agent_b: {...}}}
response: {login: abc, agent_network_def: {agent_a: {...}, agent_c: {...}}}
result:   {login: abc, agent_network_def: {agent_a: {...}, agent_b: {...}, agent_c: {...}}} 
❌ agent_b incorrectly retained
```

The intent is for the response to **replace** top-level keys, not accumulate nested ones.

dsargent: This is an AI response so I'm not sure which part is AI/hallucinations and what part is actually you @Noravee. I want to know what _you_ think here, not ChatGPT.

Regarding the merging strategy, originally @d1donlydfink just said "merge the incoming sly_data with what is already there" but he didn't say whether it should be a deep or shallow merge, and if deep, "how deep". Are you sure you're not being too AND specific here? Remember there are other uses for sly_data than the one we're talking about here and different merge strategies may apply.

dsargent: And no, we don't want to have per-agent merge strategies. That would be a maintenance headache, not to mention requiring separate tests for each case.

packages/ui-common/components/AgentChat/ChatCommon.tsx
```diff
@@ -717,21 +717,24 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+        // Only AGENT_FRAMEWORK messages can have chat_context and sly_data.
+        if (chatMessage.type === ChatMessageType.AGENT_FRAMEWORK) {
+            if (chatMessage.chat_context) {
+                // Save the chat_context, potentially overwriting any previous ones we received during this session.
+                // We only care about the last one received.
+                chatContext.current = chatMessage.chat_context
+            }
+
+            if (chatMessage.sly_data) {
+                // Save the sly_data, potentially overwriting any previous ones we received during this session.
```
dsargent, line 729: This comment is inaccurate too.

packages/ui-common/components/AgentChat/ChatCommon.tsx
Noravee: I was afraid about miscommunication so I used AI to help on writing, but I always make sure to proof read the message and that the idea is correct. You can see my original message before AI modification here:

```
There are 2 main fields in each message; request and response. Both of which can contain sly_data,
where sly_data in the request is what client sends, which in MAUI case is the login name and sly_data in
the response is what the agent returns. When i says only the AGENT_FRAMEWORK has sly_data I refer to
the response one, which is the one we interested in and should be merged with the sly data in the first
request. This will become our new request sly data.

Also note that I think the merge here should be shallow. For example
shallow merge:
request sly data: {login: abc, agent_network_def:{agent_a:{...}, agent_b:{...}}}
response sly data: {login: abc, agent_network_def:{agent_a:{...}, agent_c:{...}}}
result: {login: abc, agent_network_def:{agent_a:{...}, agent_c:{...}}} 
which is what we want
deep merge:
request sly data: {login: abc, agent_network_def:{agent_a:{...}, agent_b:{...}}}
response sly data: {login: abc, agent_network_def:{agent_a:{...}, agent_c:{...}}}
result: {login: abc, agent_network_def:{agent_a:{...}, agent_b:{...}, agent_c:{...}}} 
which is incorrect because we want to replace b with c.
```

On the merging strategy, it does not matter for other networks if the merge is shallow or deep since they are simpler. For example,
`music_nerd_pro_sly`:
request sly data: {}
response sly data: {"running_cost": 3.0}
request sly data: {"running_cost": 3.0}
response sly data: {"running_cost": 6.0}

`math_guy_passthrough`:
request sly data: {"x": 2, "y": 3}
response sly data: {"equals": 5}
request sly data: {"x": 3, "y": 4, "equals": 5}
response sly data: {"equals": 7}

In general, this is similar to chat context that we take what comes out (chat history or response sly data), merge or combine with user wants to put in (new user prompt or new input sly data) and send it to agent.

dsargent: Your original explanation was perfectly understandable and I much prefer it in "your voice". The risk of miscommunication is _greater_ when filtering it through AI due to the possibility of hallucinations or the AI misunderstanding the context. I prefer it unfiltered, in your words, please.

I guess for now we will stick with shallow merge and hope that it does cause trouble down the line with some network that has deeper `sly_data` and requires deep merge semantics.

Also @swensel the tendency I'm seeing as Neuro-san evolves seems to be "don't make assumptions about message types and what they can contain". So perhaps `ChatCommon` should be agnostic about which message types have `sly_data`, which ones have `chat_context` and "use it where we find it". I wish the contract were tighter but each time we try to apply strict rules, it comes back and bites us, "yeah but in _this_ particular case...".

dsargent: > Deep merge (incorrect)
> request:  {login: abc, agent_network_def: {agent_a: {...}, agent_b: {...}}}

Aside, we _really_ need to deprecate this flawed non-solution of having the client supply the user info. That should come from the auth system (probably a JWT whose claims can be extracted on the server). Never trust the Client.

swensel:
Will merge and handle the feedback in #274.

Issue #267, opened 2026-04-22
[UN-3345] Add previous conversation support


dsargent:
**Migrated from Jira:** [UN-3345](https://leaf-ai.atlassian.net/browse/UN-3345)

## Description
ChatGPT allows users to save and return to previous conversations.  I understand that we may not want MAA UI to be a chatbot, but as we are considering things like having 1,000 nodes in a chart, that may demo well, but are there other features like this that may be more useful for end-users?  Just raising this ticket for discussion.  This would probably be a server-side change.  It also would require persisting the data somewhere.

## Jira Details
- **Issue Type:** Task
- **Priority:** Major
- **Status:** Nice to Haves
- **Reporter:** Sean Wensel
- **Assignee:** Sean Wensel
- **Created:** 2025-07-10 10:30:21
- **Updated:** 2025-10-20 10:29:43

PR #275, opened 2026-04-27
#267 #253 #278 Support persistent history per agent network

### LINKED ISSUE #253: Can click sample queries while interaction is running

dsargent:
[Copilot summary](https://github.com/cognizant-ai-lab/neuro-san-ui/pull/275#pullrequestreview-4208953051).

Retains conversation history with each agent.

Also recap history when switching between agents so user has continuity. Also fixes #253 and #278

Supports "all kinds" of history:

1) ChatContext (for Neuro-san agents)
2) ChatHistory (for user recap + legacy agents)
3) `sly_data` -- used for all kinds of things.

Not persisted by this PR:
* `structure` field: gets consumed "live" so none of our use cases currently require us to persist it. That may change.

Initial rationale is to support creation of networks using Agent Network Designer (AND). After creating a network, a user will want to switch back and forth between AND and the newly created network to iterate on their design; to do that, persistent history is required, or else context is lost when switching networks (the status quo, prior to this PR).

Demo

https://github.com/user-attachments/assets/01f58f3b-bb60-419c-891c-bd2859de7475

Chat history collapsing section

https://github.com/user-attachments/assets/1181bbad-d45b-48ae-82d1-f900b7700694

Future enhancement: consider folding ChatHistory into ChatContext as the former is basically a subset of the latter, which is still around due to legacy agents.

### DIFF
--- a/__tests__/common/strictMocks.ts
+++ b/__tests__/common/strictMocks.ts
@@ -14,6 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+import "fake-indexeddb/auto"
+
 /**
  * This function is used to clear and reset all mocks before each test.
  * This helps to ensure that each test is isolated and does not depend on the state of any previous tests, avoiding
@@ -36,5 +38,8 @@ export const withStrictMocks = (options: StrictMockOptions = {}) => {
         if (resetModules) {
             jest.resetModules()
         }
+
+        // Reset IndexedDB mock store
+        globalThis.indexedDB = new IDBFactory()
     })
 }
--- a/apps/main/__tests__/pages/MultiAgentAcceleratorPage.test.tsx
+++ b/apps/main/__tests__/pages/MultiAgentAcceleratorPage.test.tsx
@@ -290,7 +290,6 @@ describe("Multi Agent Accelerator Page", () => {
                 expect.objectContaining({
                     currentUser: MOCK_USER,
                     targetAgent: `${TEST_AGENTS_FOLDER}/${TEST_AGENT_MUSIC_NERD}`,
-                    clearChatOnNewAgent: true,
                     neuroSanURL: NEURO_SAN_SERVER_URL,
                 })
             )
--- a/apps/main/__tests__/pages/MultiAgentAcceleratorPage.test.tsx
+++ b/apps/main/__tests__/pages/MultiAgentAcceleratorPage.test.tsx
@@ -369,6 +369,14 @@ describe("Multi Agent Accelerator Page", () => {
 
         // Force Zen mode by setting isAwaitingLlm to true
         await act(async () => {
+            useSettingsStore.getState().updateSettings({
+                behavior: {
+                    enableZenMode: true,
+                },
+            })
+        })
+
+        await act(async () => {
             setIsAwaitingLlm(true)
         })
 
--- a/apps/main/__tests__/pages/MultiAgentAcceleratorPage.test.tsx
+++ b/apps/main/__tests__/pages/MultiAgentAcceleratorPage.test.tsx
@@ -369,14 +369,6 @@ describe("Multi Agent Accelerator Page", () => {
 
         // Force Zen mode by setting isAwaitingLlm to true
         await act(async () => {
-            useSettingsStore.getState().updateSettings({
-                behavior: {
-                    enableZenMode: true,
-                },
-            })
-        })
-
-        await act(async () => {
             setIsAwaitingLlm(true)
         })
 
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -170,6 +170,16 @@ export default defineConfig([
                 },
             ],
 
+            "@typescript-eslint/no-unused-vars": [
+                "error",
+                {
+                    // Useful for situations like destructuring where you want to ignore certain variables. This way
+                    // it can be made explicit in the name that the variable is intentionally unused.
+                    argsIgnorePattern: "^_",
+                    varsIgnorePattern: "^_",
+                },
+            ],
+
             "@typescript-eslint/prefer-for-of": "error",
 
             camelcase: [
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -73,10 +73,10 @@ const config: Config.InitialOptions = {
 
     coverageThreshold: {
         global: {
-            statements: -90,
-            branches: -134,
+            statements: -92,
+            branches: -137,
             functions: -23,
-            lines: -70,
+            lines: -72,
         },
     },
 
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -73,10 +73,10 @@ const config: Config.InitialOptions = {
 
     coverageThreshold: {
         global: {
-            statements: -88,
-            branches: -128,
-            functions: -22,
-            lines: -68,
+            statements: -113,
+            branches: -139,
+            functions: -36,
+            lines: -84,
         },
     },
 
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -73,10 +73,10 @@ const config: Config.InitialOptions = {
 
     coverageThreshold: {
         global: {
-            statements: -113,
+            statements: -94,
             branches: -139,
-            functions: -36,
-            lines: -84,
+            functions: -24,
+            lines: -74,
         },
     },
 
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -73,10 +73,10 @@ const config: Config.InitialOptions = {
 
     coverageThreshold: {
         global: {
-            statements: -94,
-            branches: -139,
-            functions: -24,
-            lines: -74,
+            statements: -90,
+            branches: -134,
+            functions: -23,
+            lines: -70,
         },
     },
 
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
         "eslint-plugin-react-hooks": "7.1.1",
         "eslint-plugin-testing-library": "7.16.2",
         "eslint-plugin-unicorn": "64.0.0",
+        "fake-indexeddb": "6.2.5",
         "globals": "17.4.0",
         "jest": "30.3.0",
         "jest-environment-jsdom": "30.3.0",
--- a/packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
+++ b/packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
@@ -14,6 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+// Include mock for IndexedDB
+import "fake-indexeddb/auto"
 import {createTheme, PaletteMode, ThemeProvider} from "@mui/material/styles"
 import {act, fireEvent, render, screen, waitFor, within} from "@testing-library/react"
 import {default as userEvent, UserEvent} from "@testing-library/user-event"
@@ -23,6 +25,7 @@ import {
     TEST_AGENT_MATH_GUY,
     TEST_AGENT_MATH_GUY_DISPLAY,
     TEST_AGENT_MUSIC_NERD,
+    TEST_AGENT_MUSIC_NERD_DISPLAY,
 } from "../../../../../../__tests__/common/NetworksListMock"
 import {withStrictMocks} from "../../../../../../__tests__/common/strictMocks"
 import {USER_AGENTS} from "../../../../../../__tests__/common/UserAgentTestUtils"
@@ -135,20 +138,18 @@ describe("ChatCommon", () => {
     const sendQuery = async (agent: CombinedAgentType, query: string) => {
         // locate user query input
         const userQueryInput = screen.getByPlaceholderText(`Chat with ${cleanUpAgentName(agent)}`)
-        expect(userQueryInput).toBeInTheDocument()
 
         // Type a query
         await user.type(userQueryInput, query)
 
         // Find "Send" button
         const sendButton = screen.getByRole("button", {name: "Send"})
-        expect(sendButton).toBeInTheDocument()
 
         // Click on the "Send" button
         await user.click(sendButton)
     }
 
-    it.each(["light", "dark"] as PaletteMode[])("Should render correctly with %s mode", async (darkMode) => {
+    it.each(["light"] as PaletteMode[])("Should render correctly with %s mode", async (darkMode) => {
         renderChatCommonComponent({}, darkMode)
 
         await screen.findByText(TEST_AGENT_MATH_GUY_DISPLAY)
@@ -275,7 +276,6 @@ describe("ChatCommon", () => {
         renderChatCommonComponent({onSend: mockSendFunction})
 
         const userInput = await screen.findByPlaceholderText(CHAT_WITH_MATH_GUY)
-        expect(userInput).toBeInTheDocument()
 
         // Type user input
         await user.type(userInput, strToCheck)
@@ -283,7 +283,7 @@ describe("ChatCommon", () => {
         expect(userInput).toHaveValue(strToCheck)
 
         const sendButton = await screen.findByRole("button", {name: "Send"})
-        expect(sendButton).toBeInTheDocument()
+
         // Click Send button
         await user.click(sendButton)
 
@@ -299,7 +299,6 @@ describe("ChatCommon", () => {
         renderChatCommonComponent({onSend: mockSendFunction})
 
         const userInput = screen.getByPlaceholderText(CHAT_WITH_MATH_GUY)
-        expect(userInput).toBeInTheDocument()
 
         // Type user input
         await user.type(userInput, strToCheck)
@@ -320,7 +319,6 @@ describe("ChatCommon", () => {
         renderChatCommonComponent({onSend: mockSendFunction})
 
         const userInput = screen.getByPlaceholderText(CHAT_WITH_MATH_GUY)
-        expect(userInput).toBeInTheDocument()
 
         // Type line 1
         await user.type(userInput, strToCheckLine1)
@@ -366,7 +364,7 @@ describe("ChatCommon", () => {
             const query = "Sample test query for chunk handling"
             await sendQuery(TEST_AGENT_MATH_GUY, query)
 
-            expect(await screen.findByText(testResponseText)).toBeInTheDocument()
+            await screen.findByText(testResponseText)
             expect(onChunkReceivedMock).toHaveBeenCalledTimes(1)
             expect(onChunkReceivedMock).toHaveBeenCalledWith(chunk)
         }
@@ -384,7 +382,7 @@ describe("ChatCommon", () => {
         const query = "Sample test query for chunk handling"
         await sendQuery(LegacyAgentType.DataGenerator, query)
 
-        expect(await screen.findByText(testResponseText)).toBeInTheDocument()
+        await screen.findByText(testResponseText)
         expect(onChunkReceivedMock).toHaveBeenCalledTimes(1)
         expect(onChunkReceivedMock).toHaveBeenCalledWith(testResponseText)
     })
@@ -401,12 +399,12 @@ describe("ChatCommon", () => {
         const query = "Sample test query for legacy agent final answer handling"
         await sendQuery(LegacyAgentType.DMSChat, query)
 
-        expect(await screen.findByText(testResponseText)).toBeInTheDocument()
+        await screen.findByText(testResponseText)
         expect(onChunkReceivedMock).toHaveBeenCalledTimes(1)
         expect(onChunkReceivedMock).toHaveBeenCalledWith(testResponseText)
 
         // should be a span with content "Final Answer"
-        expect(screen.getByText("Final Answer")).toBeInTheDocument()
+        screen.getByText("Final Answer")
     })
 
     it("Should correctly handle errors thrown while fetching", async () => {
@@ -547,14 +545,14 @@ describe("ChatCommon", () => {
     it("Should show agent introduction", async () => {
         renderChatCommonComponent()
 
-        expect(await screen.findByText(TEST_AGENT_MATH_GUY_DISPLAY)).toBeInTheDocument()
+        await screen.findByText(TEST_AGENT_MATH_GUY_DISPLAY)
     })
 
-    it("Should clear chat when a new agent is selected", async () => {
+    it("Should not clear chat when a new agent is selected", async () => {
         const {rerender} = render(<ChatCommon {...defaultProps} />)
 
         // Make sure first agent greeting appears
-        expect(await screen.findByText(TEST_AGENT_MATH_GUY_DISPLAY)).toBeInTheDocument()
+        await screen.findByText(TEST_AGENT_MATH_GUY_DISPLAY)
 
         rerender(
             <ChatCommon
@@ -563,11 +561,11 @@ describe("ChatCommon", () => {
             />
         )
 
-        // Previous agent output should have been cleared
-        expect(screen.queryByText(TEST_AGENT_MATH_GUY_DISPLAY)).not.toBeInTheDocument()
+        // Previous agent output should still be present
+        screen.queryByText(TEST_AGENT_MATH_GUY_DISPLAY)
 
         // New agent greeting should be present
-        expect(await screen.findByText(cleanUpAgentName(TEST_AGENT_MUSIC_NERD))).toBeInTheDocument()
+        await screen.findByText(TEST_AGENT_MUSIC_NERD_DISPLAY)
     })
 
     it("Should refuse interaction when no target agent is set", async () => {
@@ -578,7 +576,6 @@ describe("ChatCommon", () => {
         expect(screen.queryByPlaceholderText(/Chat with/u)).not.toBeInTheDocument()
 
         const overlay = document.getElementById("chat-disabled-overlay")
-        expect(overlay).toBeInTheDocument()
         expect(overlay).toHaveStyle({
             position: "absolute",
             zIndex: MODAL_Z_INDEX - 1,
@@ -592,10 +589,9 @@ describe("ChatCommon", () => {
         renderChatCommonComponent({setIsAwaitingLlm: setAwaitingLlmMock, isAwaitingLlm: true})
 
         const stopButton = await screen.findByRole("button", {name: "Stop"})
-        expect(stopButton).toBeInTheDocument()
 
         await user.click(stopButton)
-        expect(await screen.findByText("Request cancelled.")).toBeInTheDocument()
+        await screen.findByText("Request cancelled.")
         expect(setAwaitingLlmMock).toHaveBeenCalledTimes(1)
         expect(setAwaitingLlmMock).toHaveBeenCalledWith(false)
     })
@@ -612,7 +608,7 @@ describe("ChatCommon", () => {
             ref.current?.handleStop()
         })
 
-        expect(await screen.findByText("Request cancelled.")).toBeInTheDocument()
+        await screen.findByText("Request cancelled.")
         expect(setAwaitingLlmMock).toHaveBeenCalledTimes(1)
         expect(setAwaitingLlmMock).toHaveBeenCalledWith(false)
     })
@@ -621,20 +617,18 @@ describe("ChatCommon", () => {
         renderChatCommonComponent()
 
         const autoscrollButton = screen.getByRole("button", {name: "Autoscroll enabled"})
-        expect(autoscrollButton).toBeInTheDocument()
 
         await user.click(autoscrollButton)
-        expect(screen.getByRole("button", {name: "Autoscroll disabled"})).toBeInTheDocument()
+        screen.getByRole("button", {name: "Autoscroll disabled"})
     })
 
     it("Should handle text wrapping toggle correctly", async () => {
         renderChatCommonComponent()
 
         const wrapButton = screen.getByRole("button", {name: "Text wrapping enabled"})
-        expect(wrapButton).toBeInTheDocument()
 
         await user.click(wrapButton)
-        expect(screen.getByRole("button", {name: "Text wrapping disabled"})).toBeInTheDocument()
+        screen.getByRole("button", {name: "Text wrapping disabled"})
     })
 
     it("Should handle final answer from Neuro-san agents correctly", async () => {
@@ -651,7 +645,7 @@ describe("ChatCommon", () => {
 
         await sendQuery(TEST_AGENT_MATH_GUY, "Sample test query final answer test")
 
-        expect(await screen.findByText("Final Answer")).toBeInTheDocument()
+        await screen.findByText("Final Answer")
     })
 
     it.each([
@@ -699,16 +693,13 @@ describe("ChatCommon", () => {
 
         // Click "show thinking" button. It defaults to "hiding agent thinking" so we look for that
         const showThinkingButton = document.getElementById("show-thinking-button")
-        expect(showThinkingButton).toBeInTheDocument()
         await user.click(showThinkingButton)
 
         await sendQuery(TEST_AGENT_MATH_GUY, "Sample test query handle thinking button test")
 
         // All responses should be visible when "show thinking" is enabled
-        await waitFor(() => {
-            expect(screen.getAllByText(aiResponseText)).toHaveLength(2)
-        })
-        expect(await screen.findByText(agentResponseText)).toBeInTheDocument()
+        expect(await screen.findAllByText(aiResponseText)).toHaveLength(2)
+        await screen.findByText(agentResponseText)
 
         // Now click the button again to hide agent thinking
         await user.click(showThinkingButton)
@@ -899,18 +890,15 @@ describe("ChatCommon", () => {
         await sendQuery(TEST_AGENT_MATH_GUY, testMessage)
 
         // Wait for the message to appear
-        await waitFor(() => {
-            expect(screen.getByText(testMessage)).toBeInTheDocument()
-        })
+        await screen.findByText(testMessage)
 
         // Find and click Clear Chat button
         const clearButton = screen.getByRole("button", {name: "Clear Chat"})
-        expect(clearButton).toBeInTheDocument()
 
         await user.click(clearButton)
 
         // Verify chat is cleared and agent introduction appears
-        expect(await screen.findByText(TEST_AGENT_MATH_GUY_DISPLAY)).toBeInTheDocument()
+        await screen.findByText(TEST_AGENT_MATH_GUY_DISPLAY)
         expect(screen.queryByText(testMessage)).not.toBeInTheDocument()
     })
 
@@ -1182,7 +1170,7 @@ describe("ChatCommon", () => {
 
         await sendQuery(LegacyAgentType.OpportunityFinder, "test query")
 
-        expect(await screen.findByText("Legacy response with custom endpoint")).toBeInTheDocument()
+        await screen.findByText("Legacy response with custom endpoint")
     })
 
     it.each([
@@ -1199,7 +1187,7 @@ describe("ChatCommon", () => {
             expect(chatContainer).toBeInTheDocument()
         })
 
-        expect(await screen.findByText(TEST_AGENT_MATH_GUY_DISPLAY)).toBeInTheDocument()
+        await screen.findByText(TEST_AGENT_MATH_GUY_DISPLAY)
     })
 
     it("Should handle connectivity info error gracefully", async () => {
@@ -1209,7 +1197,7 @@ describe("ChatCommon", () => {
         renderChatCommonComponent()
 
         // Component should still render despite connectivity error
-        expect(await screen.findByText(TEST_AGENT_MATH_GUY_DISPLAY)).toBeInTheDocument()
+        await screen.findByText(TEST_AGENT_MATH_GUY_DISPLAY)
     })
 
     it("Should handle network request timeout", async () => {
--- a/packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
+++ b/packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
@@ -129,7 +129,7 @@ describe("ChatCommon", () => {
     }
 
     beforeEach(() => {
-        user = userEvent.setup()
+        user = userEvent.setup({delay: null})
 
         // Mock getConnectivity to return dummy connectivity info
         ;(getConnectivity as jest.Mock).mockResolvedValue(MOCK_CONNECTIVITY_INFO)
--- a/packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
+++ b/packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
@@ -551,12 +551,7 @@ describe("ChatCommon", () => {
     })
 
     it("Should clear chat when a new agent is selected", async () => {
-        const {rerender} = render(
-            <ChatCommon
-                {...defaultProps}
-                clearChatOnNewAgent={true}
-            />
-        )
+        const {rerender} = render(<ChatCommon {...defaultProps} />)
 
         // Make sure first agent greeting appears
         expect(await screen.findByText(TEST_AGENT_MATH_GUY_DISPLAY)).toBeInTheDocument()
@@ -564,7 +559,6 @@ describe("ChatCommon", () => {
         rerender(
             <ChatCommon
                 {...defaultProps}
-                clearChatOnNewAgent={true}
                 targetAgent={TEST_AGENT_MUSIC_NERD}
             />
         )
--- a/packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
+++ b/packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
@@ -14,8 +14,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Include mock for IndexedDB
-import "fake-indexeddb/auto"
 import {createTheme, PaletteMode, ThemeProvider} from "@mui/material/styles"
 import {act, fireEvent, render, screen, waitFor, within} from "@testing-library/react"
 import {default as userEvent, UserEvent} from "@testing-library/user-event"
@@ -30,11 +28,13 @@ import {
 import {withStrictMocks} from "../../../../../../__tests__/common/strictMocks"
 import {USER_AGENTS} from "../../../../../../__tests__/common/UserAgentTestUtils"
 import {ChatCommon, ChatCommonHandle} from "../../../../components/AgentChat/ChatCommon/ChatCommon"
+import {MAX_SAMPLE_QUERIES, QUERY_TRUNCATE_LENGTH} from "../../../../components/AgentChat/ChatCommon/SampleQueries"
 import {CombinedAgentType, LegacyAgentType} from "../../../../components/AgentChat/Common/Types"
 import {cleanUpAgentName} from "../../../../components/AgentChat/Common/Utils"
 import {getConnectivity, sendChatQuery} from "../../../../controller/agent/Agent"
 import {sendLlmRequest, StreamingUnit} from "../../../../controller/llm/LlmChat"
 import {ChatContext, ChatMessage, ChatMessageType, ChatResponse} from "../../../../generated/neuro-san/NeuroSanClient"
+import {useAgentChatHistoryStore} from "../../../../state/ChatHistory"
 
 // Mock agent API
 jest.mock("../../../../controller/agent/Agent")
@@ -90,6 +90,8 @@ const MOCK_CONNECTIVITY_INFO = {
 }
 
 describe("ChatCommon", () => {
+    withStrictMocks()
+
     let user: UserEvent
 
     const defaultProps = {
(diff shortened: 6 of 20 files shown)


packages/ui-common/state/ChatHistory.ts
```diff
@@ -0,0 +1,148 @@
...
+            resetHistory: (agentId: string) =>
+                set((state) => {
+                    const {[agentId]: _, ...rest} = state.history
+                    return {history: rest}
+                }),
+        }),
+        {
+            name: STORE_KEY,
+            storage: chatHistoryStorage,
+        }
```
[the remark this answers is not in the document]
dsargent: We already mock this in tests which is the only non-browser environment we care about.

packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
```diff
@@ -547,33 +545,27 @@ describe("ChatCommon", () => {
...
                 {...defaultProps}
-                clearChatOnNewAgent={true}
                 targetAgent={TEST_AGENT_MUSIC_NERD}
             />
         )
 
-        // Previous agent output should have been cleared
-        expect(screen.queryByText(TEST_AGENT_MATH_GUY_DISPLAY)).not.toBeInTheDocument()
+        // Previous agent output should still be present
+        screen.queryByText(TEST_AGENT_MATH_GUY_DISPLAY)
```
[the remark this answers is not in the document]
dsargent: Wrong, `queryBy` includes an implicit assertion.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
dsargent, line 1: A lot of changes in here, unsurprisingly. I also refactored it so the main render logic is super short, and delegates to functions to create each item. Much more readable and modular.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -17,7 +17,7 @@ limitations under the License.
 /**
  * See main function description.
  */
-import {AIMessage, BaseMessage, HumanMessage} from "@langchain/core/messages"
+import {AIMessage, HumanMessage} from "@langchain/core/messages"
```
dsargent, line 20: No longer need `BaseMessage` here as chat history persistence is "outsourced" to a new Zustand store (backed by IndexedDB).

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -48,8 +48,10 @@ import {
 } from "react"
 import ReactMarkdown from "react-markdown"
 import SyntaxHighlighter from "react-syntax-highlighter"
+import {v4 as uuid} from "uuid"
```
dsargent, line 51: Had to pull in the `uuid` library. 

"Bro -- just use `crypto.randomUUID()`!". [Nope](https://leaf-ai.atlassian.net/browse/UN-3078).

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -145,12 +147,6 @@ export interface ChatCommonProps {
      */
     readonly agentPlaceholders?: Partial<Record<CombinedAgentType, string>>
 
-    /**
```
dsargent, line 148: This feature never quite worked right, and the way I implemented it was hacky. The right way to clear things is with rendering/dependencies/changing the key of the component if we really want a full remount, so that's what we do now.

I verified that it works fine with Neuro AI (which always set this to `false` anyway. Only MAUI used the `true` path.)

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -178,6 +174,9 @@ export interface ChatCommonProps {
     readonly neuroSanURL?: string
 }
 
+// Key for the chat history, which gets special treatment; always visible even if "show thinking" is off.
```
dsargent, line 177: See comment

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -201,34 +200,34 @@ export type ChatCommonHandle = {
 const extractFinalAnswer = (response: string) =>
     /Final Answer: (?<finalAnswerText>.*)/su.exec(response)?.groups?.["finalAnswerText"]
 
+// Maximum number of items to keep in the chat output window
+const MAX_CHAT_OUTPUT_ITEMS = 50
```
dsargent, line 204: See comment

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -201,34 +200,34 @@ export type ChatCommonHandle = {
...
  * Common chat component for agent chat. This component is used by all agent chat components to provide a consistent
  * experience for users when chatting with agents. It handles user input as well as displaying and nicely formatting
  * agent responses. Customization for inputs and outputs is provided via event handlers-like props.
  */
 export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCommonHandle>}) => {
-    const slyData = useRef<Record<string, unknown>>({})
-
     const {
-        id,
+        agentPlaceholders = EMPTY,
```
dsargent, line 213: Alphabetic sort + changed props a little

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -256,7 +255,7 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
     // Controller for cancelling fetch request
     const controller = useRef<AbortController>(null)
 
-    // For tracking if we're autoscrolling. A button allows the user to enable or disable autoscrolling.
+    // For tracking if we're auto-scrolling. A button allows the user to enable or disable auto-scrolling.
```
dsargent, line 258: My IDE flags this as a typo and it was annoying me

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -271,19 +270,25 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
     // Ref for the final answer key, so we can highlight the accordion
     const finalAnswerKey = useRef<string>("")
 
-    // Use useRef here since we don't want changes in the chat history to trigger a re-render
```
dsargent, line 274: This stuff now moved to the new Zustand store.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -271,19 +270,25 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
-    Note on ChatContext vs ChatHistory:
-    "Legacy" (not Neuro-san) agents use ChatHistory, which is a collection of messages of various types, Human, AI,
-    System etc. It mimics the langchain field of the same name.
-    Neuro-san agents deal in ChatContext, which is a more complex collection of chat histories, since more agents
-    are involved.
-    Both fields fulfill the same purpose: to maintain conversation state across multiple messages.
-    */
-    const chatContext = useRef<ChatContext>(null)
+    // Persistent agent chat history store, which is where we store both kinds of chat histories
+    // (see store implementation for details)
```
dsargent, line 274: For context/chat/sly_data history we now use the Zustand store (and ultimately IndexedDB.)

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -322,6 +327,8 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
 
     const {atelierDuneDark, a11yLight} = HLJS_THEMES
 
+    const agentDisplayName = useMemo(() => cleanUpAgentName(removeTrailingUuid(targetAgent)), [targetAgent])
```
dsargent, line 330: Memoized this as it's used in several places. And we also had bugs where we forgot to remove the UUID.

Future: we should use the agent name from the reservation for temporary reservations, not the reservation ID. Then we wouldn't need all this UUID trimming business.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -344,8 +351,7 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
     useEffect(() => {
         // Scroll the final answer into view
         if (finalAnswerRef.current && !isAwaitingLlm) {
-            const offset = 50
-            chatOutputRef.current.scrollTop = finalAnswerRef.current.offsetTop - offset
+            chatOutputRef.current.scrollTop = finalAnswerRef.current.offsetTop - 50
```
dsargent, line 354: Unnecessary local variable

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -441,13 +447,16 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
     )
 
     /**
-     * Handles adding content to the output window.
+     * Handles adding content to the output window. We only store the last MAX_CHAT_OUTPUT_ITEMS items to keep
+     * memory usage down.
      * @param node A ReactNode to add to the output window -- text, spinner, etc. but could also be  simple string
-     * @returns Nothing, but updates the output window with the new content. Updates currentResponse as a side effect.
```
dsargent, line 446: Got rid of this side effect -- ugly. Now callers that want their message added to chat history (of which there were only a couple) have to do it themselves.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -441,13 +447,16 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+     * @returns Nothing, but updates the output window with the new content.
      */
     const updateOutput = useCallback((node: ReactNode) => {
-        currentResponse.current += node
-        setChatOutput((currentOutput) => [...currentOutput, node])
+        setChatOutput((current) => {
+            const next = [...current, node]
+            return next.length > MAX_CHAT_OUTPUT_ITEMS ? next.slice(-MAX_CHAT_OUTPUT_ITEMS) : next
```
dsargent, line 458: Limit length -- I've seen memory growing while using MAUI and I suspect this might be a contributor. Long conversations could go on for many screens so we have to have a "reasonable" limit. TBD if this is a reasonable limit.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -460,6 +469,7 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
             if (isLegacyAgentType(targetAgent)) {
                 // Display output as-is
                 updateOutput(chunk)
+                currentResponse.current += chunk
```
dsargent, line 472: Explicitly add it to chat history -- since we removed the side effect from `updateOutput`.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -477,19 +487,16 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
                 return
             }
 
+            // Shallow merge existing slyData with incoming chatMessage.sly_data
```
dsargent, line 490: Functionality change! Moved this here so we _always_ do it when there's `sly_data`.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -477,19 +487,16 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+                updateSlyData(targetAgent, chatMessage.sly_data)
+            }
+
             // It's a ChatMessage. Does it have chat context? Only AGENT_FRAMEWORK messages can have chat context.
             if (chatMessage.type === ChatMessageType.AGENT_FRAMEWORK && chatMessage.chat_context) {
                 // Save the chat context, potentially overwriting any previous ones we received during this session.
                 // We only care about the last one received.
-                chatContext.current = chatMessage.chat_context
-
-                // Nothing more to do with this message. It's just a message to give us the chat context, so return
```
dsargent, line 486: Funcdtionality change: we keep processing even after snagging the `chat_context`. `AGENT_FRAMEWORK` messages can have other purposes than just carrying `chat_context` nowadays; wasn't originally the case.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -516,13 +523,12 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
                         ? cleanUpAgentName(chatMessage.origin[chatMessage.origin.length - 1].tool)
                         : "Agent message"
                 updateOutput(processLogLine(chatMessage.text, agentName, chatMessage.type))
+                currentResponse.current += chatMessage.text
```
dsargent, line 526: Explicitly add to history

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -516,13 +523,12 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
             }
         },
-        [onChunkReceived, processLogLine, targetAgent, updateOutput]
+        [onChunkReceived, processLogLine, updateSlyData, targetAgent, updateChatContext, updateOutput]
```
dsargent, line 529: Update deps in line with `exhaustive-deps` rule.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -538,7 +544,8 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
         // Random greeting
         const greeting = AGENT_GREETINGS[Math.floor(Math.random() * AGENT_GREETINGS.length)]
         updateOutput(greeting)
-    }, [agentDisplayName, targetAgent, updateOutput])
+        // eslint-disable-next-line react-hooks/exhaustive-deps -- updateOutput is stable (empty useCallback deps)
+    }, [agentDisplayName, targetAgent])
```
dsargent, line 548: Remove a stable dep from the array (saves a few unnecessary renders)

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -716,15 +733,18 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
                             {processLogLine(lastAIMessage.current, "Final Answer", ChatMessageType.AI, true)}
                         </div>
                     )
+                    // Record bot answer in history.
```
dsargent, line 736: Recording stuff in chat history is now decoupled from `updateOutput` (🎉) so do this here.

I'm using the `uuid` library to add a unique ID to each message, so that we can use that to generate unique React.js keys later. We were not doing this before, resulting in occasional duplicate keys.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -744,20 +764,12 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
             setIsAwaitingLlm,
             showThinking,
             targetAgent,
+            updateChatHistory,
             updateOutput,
             userImage,
         ]
     )
 
-    useEffect(() => {
```
dsargent, line 752: This effect wasn't really needed. We handle things like this via `resetState` and remounting the component when necessary (the parent changes the key on new agent which reboots everything).

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -787,12 +799,7 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
                     />
                 )
                 const sampleQueries = (connectivity?.metadata?.["sample_queries"] || []) as string[]
-                updateOutput(
-                    <SampleQueries
-                        handleSend={handleSend}
-                        sampleQueries={sampleQueries}
-                    />
-                )
+                setAgentSampleQueries(sampleQueries)
```
dsargent, line 802: Sample queries now get handled as part of the main render function, so here we just save the input to that component in State for later rendering. Makes things more predictable and "functionally purer".

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -804,7 +811,8 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
         if (targetAgent && !isLegacyAgentType(targetAgent)) {
             void fetchAgentDetails()
         }
-    }, [agentDisplayName, currentUser, handleSend, id, neuroSanURL, targetAgent, updateOutput])
+        // eslint-disable-next-line react-hooks/exhaustive-deps -- only want to run this when targetAgent changes
+    }, [targetAgent])
```
dsargent, line 815: Really the only "unstable" value we care about here is `targetAgent`; when that changes, we definitely want to re-run the effect. The other items are either stable across renders, or not relevant to what this effect does. This saves us a lot of renders.

The downside is point-silencing `react-hooks/exhaustive-deps` which is sometimes a smell that means "your design needs some work". Maybe so, but not for this PR.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -849,303 +857,358 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+        resetHistory(targetAgent)
         previousUserQuery.current = ""
         currentResponse.current = ""
         lastAIMessage.current = ""
         introduceAgent()
-    }, [introduceAgent])
+    }, [introduceAgent, resetHistory, targetAgent])
 
-    return (
+    /**
```
dsargent, line 867: Add copious doc.

Changed implementation here: now when nodes are "hidden" (eg. they are random agent chat and user has turned off "show thinking"), we don't even add them to the DOM. Seems cleaner and more efficient.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -849,303 +857,358 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+                            showThinking || item.key === finalAnswerKey.current || item.key === CHAT_HISTORY_KEY
+                        return shouldShow ? item : null
+                    }
+                    return item
+                })
+                .filter((item) => item !== null),
+        [chatOutput, showThinking]
+    )
+
+    const getNoAgentOverlay = () => (
```
dsargent, line 892: Broke the main render into these separate functions

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -849,303 +857,358 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+                bottom: 0,
+                zIndex: theme.zIndex.modal - 1,
+                cursor: "not-allowed",
+                // Capture all pointer events to prevent interaction with the chat when no agent is selected
+                pointerEvents: "all",
+            }}
+        />
+    )
+
+    const getTitle = () => (
```
dsargent, line 909: Broken out from main render

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -849,303 +857,358 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+                    id={`close-button-${id}`}
+                    onClick={onClose}
+                >
+                    <CloseIcon id={`close-icon-${id}`} />
+                </IconButton>
+            )}
+        </Box>
+    )
+
+    const getOptionsButtons = () => (
```
dsargent, line 942: Broken out

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -849,303 +857,358 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+                    <WrapTextIcon
+                        id="wrap-icon"
+                        sx={{color: "var(--bs-white)", fontSize: "0.85rem"}}
+                    />
+                </LlmChatOptionsButton>
+            </Tooltip>
+        </>
+    )
+
+    const getResponseBox = () => (
```
dsargent, line 998: Broken out

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -849,303 +857,358 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+                isAwaitingLlm={isAwaitingLlm}
+                handleSend={handleSend}
+                handleStop={handleStop}
+                previousUserQuery={previousUserQuery.current}
+                shouldEnableRegenerateButton={shouldEnableRegenerateButton}
+            />
+        </Box>
+    )
+
+    const getUserInputBox = () => (
```
dsargent, line 1089: Broken out

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -849,303 +857,358 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
 
-                    {/* Send Button */}
-                    <SendButton
-                        enableSendButton={shouldEnableSendButton}
-                        id="submit-query-button"
-                        onClickCallback={() => handleSend(chatInput)}
-                    />
-                </Box>
-            </Box>
+    const getChatBox = () => (
```
dsargent, line 1182: Broken out

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -849,303 +857,358 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+                position: "relative",
+            }}
+        >
+            {title && getTitle()}
+            {getResponseBox()}
+            {getUserInputBox()}
+        </Box>
+    )
+
+    return (
```
dsargent, line 1201: The payoff -- the actual `return` now just returns this one `Box` with either the "no agent selected so can't do anything" overlay, or the usual chat components. Everything else is broken out into functions.

packages/ui-common/components/AgentChat/ChatCommon/ChatHistory.tsx
dsargent, line 1: New module for the chat history component. Really just a `MUIAccordion` with the history items, but cleaner design to encapsulate it in a reusable component.

Previously this was a mutable "effect" in `ChatCommon` which is uglier and harder to reason about. This way it's a nice, predictable separate component with well-defined props.

packages/ui-common/components/AgentChat/ChatCommon/SampleQueries.tsx
```diff
@@ -6,12 +6,13 @@ import {FC} from "react"
 import {hashString} from "../../../utils/text"
 
 // Maximum number of sample queries to show
-const MAX_SAMPLE_QUERIES = 5
+export const MAX_SAMPLE_QUERIES = 5
```
dsargent, line 9: Export for single point of truth in tests. I feel a bit uneasy exporting stuff just so tests can get at it, but it prevents duplication of information so it's a trade off.

packages/ui-common/components/AgentChat/ChatCommon/SampleQueries.tsx
```diff
@@ -6,12 +6,13 @@ import {FC} from "react"
...
 
 // Maximum length of query to show in sample query chips
-const QUERY_TRUNCATE_LENGTH = 80
+export const QUERY_TRUNCATE_LENGTH = 80
 
 interface SampleQueriesProps {
+    readonly disabled: boolean
```
dsargent, line 15: Add a `disabled` property so we can fix https://github.com/cognizant-ai-lab/neuro-san-ui/issues/253

packages/ui-common/components/AgentChat/ChatCommon/SampleQueries.tsx
```diff
@@ -32,11 +33,11 @@ export const SampleQueries: FC<SampleQueriesProps> = ({handleSend, sampleQueries
                 const hashedQuery = hashString(query)
                 return (
                     <Tooltip
-                        title={`Click to send query: "${hashedQuery}"`}
+                        title={`Click to send query: "${query}"`}
```
dsargent, line 36: Oops...bug. This fixes it.

packages/ui-common/components/AgentChat/ChatCommon/SampleQueries.tsx
```diff
@@ -32,11 +33,11 @@ export const SampleQueries: FC<SampleQueriesProps> = ({handleSend, sampleQueries
...
                         key={`tooltip-${hashedQuery}`}
                     >
                         <Chip
-                            key={hashedQuery}
```
dsargent, line 39: Only need a key at the top level for React.js.

packages/ui-common/components/AgentChat/ChatCommon/SampleQueries.tsx
```diff
@@ -50,9 +51,7 @@ export const SampleQueries: FC<SampleQueriesProps> = ({handleSend, sampleQueries
                                 marginRight: "1rem",
                                 marginBottom: "1rem",
                                 backgroundColor: "var(--bs-accent1-medium)",
-                                "&:hover": {
-                                    backgroundColor: "var(--bs-accent1-dark)",
-                                },
+                                "&:hover": {backgroundColor: "var(--bs-accent1-dark)"},
```
dsargent, line 54: Compactify format.

packages/ui-common/components/AgentChat/ChatCommon/UserQueryDisplay.tsx
```diff
@@ -34,18 +35,20 @@ const UserQueryContainer = styled("div")(({theme}) => ({
 // #endregion: Styled Components
 
 export const UserQueryDisplay = ({
+    sx,
```
dsargent, line 38: Added an optional `sx` prop here for overriding the style. In my particular use case, lowering `opacity` for the chat history display.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -134,6 +134,8 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
 
     const [selectedNetwork, setSelectedNetwork] = useState<string | null>(null)
 
+    const networkDisplayName = useMemo(() => cleanUpAgentName(removeTrailingUuid(selectedNetwork)), [selectedNetwork])
```
dsargent, line 137: Memoize for optimization

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -242,11 +244,10 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
                     setAgentIconSuggestions(null)
                     closeNotification()
                 } catch (e) {
-                    const networkName = cleanUpAgentName(selectedNetwork)
                     sendNotification(
                         NotificationType.error,
                         "Connection error",
-                        `Unable to get agent list for "${networkName}". Verify that ${neuroSanURL} is a valid ` +
+                        `Unable to get agent list for "${networkDisplayName}". Verify that ${neuroSanURL} is a valid ` +
```
dsargent, line 250: This was a bug -- the ugly UUID was being included in the message shown to the user.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -530,7 +531,6 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
                         onChunkReceived={onChunkReceived}
                         onStreamingComplete={onStreamingComplete}
                         onStreamingStarted={onStreamingStarted}
-                        clearChatOnNewAgent={true}
```
dsargent, line 533: Feature removed -- see related annotation in `ChatCommon`.

packages/ui-common/state/ChatHistory.ts
dsargent, line 1: New Zustand store for chat history (including `sly_data` history.) Plenty of inline doc so I won't describe everything here.

packages/ui-common/state/IndexedDBStorage.ts
dsargent, line 1: New module for accessing IndexedDB. We plug it in to the Zustand store in order to persist chat history to IndexedDB.

I was hoping Zustand would have this built in, but alas no. It only knows natively how to persist to in-memory state, or local storage, not IndexedDB.

packages/ui-common/state/Settings.ts
```diff
@@ -86,7 +86,7 @@ export const DEFAULT_SETTINGS: Settings = {
         secondary: null,
     },
     behavior: {
-        enableZenMode: true,
+        enableZenMode: false,
```
dsargent: Functionality and UI change! A bit opinionated _but_, I think most users don't want Zen mode by default. Demo-givers can turn it on via the settings dialog.

Come at me...

packages/ui-common/package.json
```diff
@@ -20,6 +20,7 @@
         "react-syntax-highlighter": "16.1.0",
         "rehype-raw": "7.0.0",
         "rehype-slug": "6.0.0",
+        "uuid": "11.1.0",
```
dsargent, line 23: Had to add this to generate unique IDs. See annotations on `ChatCommon` changes.

packages/ui-common/package.json
```diff
@@ -33,6 +34,7 @@
         "@types/react-dom": "19.2.3",
         "@types/react-syntax-highlighter": "15.5.13",
         "babel-jest": "30.2.0",
+        "fake-indexeddb": "6.2.5",
```
dsargent, line 37: Testing code that uses `IndexedDB` is hell so I pulled this in. At least it's only a dev dependency.

eslint.config.mjs
```diff
@@ -170,6 +170,16 @@ export default defineConfig([
                 },
             ],
 
+            "@typescript-eslint/no-unused-vars": [
+                "error",
+                {
+                    // Useful for situations like destructuring where you want to ignore certain variables. This way
+                    // it can be made explicit in the name that the variable is intentionally unused.
+                    argsIgnorePattern: "^_",
```
dsargent, line 178: More opinionated changes...I want to have the ability to have unused variables when we are intentional about it, like when destructuring and removing props, as the comment mentions. So this way we can have `_foo` and the linter won't complain, which is consistent with how most IDEs behave.

jest.config.ts
```diff
@@ -73,10 +73,10 @@ const config: Config.InitialOptions = {
 
     coverageThreshold: {
         global: {
-            statements: -88,
-            branches: -128,
-            functions: -22,
-            lines: -68,
+            statements: -92,
```
dsargent, line 76: Slight dip, will follow up with a "back to where we were and better" PR. I think it's a worthwhile sacrifice to get the vibe coding MVP out there.

package.json
```diff
@@ -51,6 +51,7 @@
         "eslint-plugin-react-hooks": "7.1.1",
         "eslint-plugin-testing-library": "7.16.2",
         "eslint-plugin-unicorn": "64.0.0",
+        "fake-indexeddb": "6.2.5",
```
dsargent, line 54: Had to add it here as well to keep yarn happy. I don't like having deps at multiple levels (root, individual packages) but Copilot insists that is normal and expected for yarn workspace projects, and suggests using "yarn constraints" to keep things synced and consistent. We can look into that later.

__tests__/common/strictMocks.ts
```diff
@@ -36,5 +38,8 @@ export const withStrictMocks = (options: StrictMockOptions = {}) => {
         if (resetModules) {
             jest.resetModules()
         }
+
+        // Reset IndexedDB mock store
```
dsargent, line 42: Added this to help keep tests pure and avoid cross-contamination now that some tests rely on IndexedDB.

apps/main/__tests__/pages/MultiAgentAcceleratorPage.test.tsx
```diff
@@ -369,6 +368,14 @@ describe("Multi Agent Accelerator Page", () => {
         expect(screen.queryByLabelText("Stop")).not.toBeInTheDocument()
 
         // Force Zen mode by setting isAwaitingLlm to true
+        await act(async () => {
```
dsargent: Now that the default is `false`, we need to do this for this test.

packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
```diff
@@ -87,6 +90,8 @@ const MOCK_CONNECTIVITY_INFO = {
 }
 
 describe("ChatCommon", () => {
+    withStrictMocks()
```
dsargent, line 93: Moved this as first item to ensure it's "applied everywhere".

packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
```diff
@@ -123,26 +128,27 @@ describe("ChatCommon", () => {
         )
     }
 
-    withStrictMocks()
-
     beforeEach(() => {
-        user = userEvent.setup()
+        user = userEvent.setup({delay: null})
```
dsargent, line 132: These tests were occasionally timing out in CI only (not locally) and Copilot suggested this. It does make them run a few seconds faster so I guess a win. It causes RTL to send keypresses "as fast as possible" rather than adding a 1ms delay between simulating keypresses to mimic a user (though of course no user types that fast!). With `delay: null` when you send a string to an input box, for example, the whole thing is sent in the same microtask without delay inbetween.

packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
```diff
@@ -123,26 +128,27 @@ describe("ChatCommon", () => {
...
 
         // Mock getConnectivity to return dummy connectivity info
         ;(getConnectivity as jest.Mock).mockResolvedValue(MOCK_CONNECTIVITY_INFO)
+
+        // Reset history. TODO: would be nice if withStrictMocks could also reset Zustand stores
```
dsargent, line 137: See comment

packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
```diff
@@ -206,7 +212,7 @@ describe("ChatCommon", () => {
 
         // Make sure long query chip is truncated
         const sampleQueries = MOCK_CONNECTIVITY_INFO.metadata.sample_queries
-        const expectedLongQuery = `${sampleQueries[1].slice(0, 80)}...`
+        const expectedLongQuery = `${sampleQueries[1].slice(0, QUERY_TRUNCATE_LENGTH)}...`
```
dsargent, line 215: Use consts from the module for single point of truth.

packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
```diff
@@ -215,12 +221,17 @@ describe("ChatCommon", () => {
         await screen.findByText(sampleQueries[4])
         expect(screen.queryByText(sampleQueries[5])).not.toBeInTheDocument()
 
-        // Click a sample query
-        const sampleQueryButton = await screen.findByText(sampleQueries[0])
-        await user.click(sampleQueryButton)
+        for (const query of sampleQueries.slice(0, MAX_SAMPLE_QUERIES - 1)) {
```
dsargent, line 224: Enhance test to test _all_ queries.

packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
```diff
@@ -275,15 +286,14 @@ describe("ChatCommon", () => {
         renderChatCommonComponent({onSend: mockSendFunction})
 
         const userInput = await screen.findByPlaceholderText(CHAT_WITH_MATH_GUY)
-        expect(userInput).toBeInTheDocument()
```
dsargent, line 278: Removed all the unnecessary `...toBeInTheDocument` assertions from this suite.

packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
```diff
@@ -366,7 +374,7 @@ describe("ChatCommon", () => {
             const query = "Sample test query for chunk handling"
             await sendQuery(TEST_AGENT_MATH_GUY, query)
 
-            expect(await screen.findByText(testResponseText)).toBeInTheDocument()
+            await screen.findByText(testResponseText)
```
dsargent, line 377: Simplify.

packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
```diff
@@ -351,7 +359,7 @@ describe("ChatCommon", () => {
 
             const chatResponse: ChatResponse = {
                 response: {
-                    type: ChatMessageType.AGENT_FRAMEWORK,
+                    type: ChatMessageType.AI,
```
dsargent, line 362: This test was kind of passing "by accident". MAUI only expects final answers to arrive in messages of type `AI`, so I changed the test to use that.

packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
```diff
@@ -705,27 +704,22 @@ describe("ChatCommon", () => {
...
 
         // Only the AI response should be visible
-        await waitFor(() => {
-            expect(screen.getAllByText(aiResponseText)).toHaveLength(2)
-        })
+        expect(screen.getAllByText(aiResponseText)).toHaveLength(1)
 
-        // Agent response should be in the DOM but with display: none
-        expect(await screen.findByText(agentResponseText)).not.toBeVisible()
+        // Agent response should not be in the DOM as we disabled "show thinking"
```
dsargent, line 721: Change in behavior: "hidden" items are now not even added to the DOM, which speeds things up a bit.

packages/ui-common/__tests__/components/Settings/SettingsDialog.test.tsx
```diff
@@ -167,18 +167,18 @@ describe("SettingsDialog", () => {
...
 
         const checkBoxElement = within(zenModeToggle).getByRole("checkbox")
 
-        expect(checkBoxElement).toBeChecked()
+        expect(checkBoxElement).not.toBeChecked()
 
         await user.click(checkBoxElement)
-
-        expect(useSettingsStore.getState().settings.behavior.enableZenMode).toBe(false)
+        expect(checkBoxElement).toBeChecked()
```
dsargent: Might as well `expect` this, just to make sure.

packages/ui-common/__tests__/components/Settings/SettingsDialog.test.tsx
```diff
@@ -167,18 +167,18 @@ describe("SettingsDialog", () => {
             />
         )
 
-        // Default: Zen mode should be enabled
-        expect(useSettingsStore.getState().settings.behavior.enableZenMode).toBe(true)
+        // Default: Zen mode should be disabled
```
dsargent: Reversed default for zen mode.

packages/ui-common/__tests__/unit/State/ChatHistory.test.ts
dsargent, line 1: Tests for new module.

packages/ui-common/__tests__/unit/State/IndexDBStorage.test.ts
dsargent, line 1: Tests for new IndexedDB adapter.

jest.config.ts
dsargent: @swensel Do you still have those "good old days" values for these stats before we had a fewdips?

packages/ui-common/state/Settings.ts
swensel: I think this is my one hang up in this PR. I like Zen mode on by default.

packages/ui-common/components/AgentChat/ChatCommon/ChatHistory.tsx
```diff
@@ -0,0 +1,70 @@
...
+    readonly messages: BaseMessage[]
+    readonly targetAgent: string
+    readonly userImage: string
+}
+// #endregion: Types
+
+/**
+ * Component for displaying chat history from previous interactions with the agent.
+ */
+export const ChatHistory: FC<ChatHistoryProps> = ({
```
dsargent, line 25: Spot the bug...

I have a fix but don't want to push anything because there's a 50-50 chance GitHub would nuke my comments.

__tests__/common/strictMocks.ts
```diff
@@ -14,6 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+import "fake-indexeddb/auto"
```
dsargent, line 17: The cost for having tests that exercise code that uses IndexedDB. We could mock it ourselves, but that would be a tedious, heavy mock that would basically end up replicating what this library is doing. Open to discussion.

swensel: It's OK with me to go this route.

jest.config.ts
swensel: I didn't go through every commit but here is one example [commit](https://github.com/cognizant-ai-lab/neuro-san-ui/commit/5d75d94dfca878bfad9bb115e8d07f39b7eb3d0e):

```
statements: -73,
branches: -124,
functions: -19,
lines: -56,
````

File: packages/ui-common/components/AgentChat/ChatCommon/ChatHistory.tsx (hunk 62 above)
Comment (swensel) [reply packages/ui-common/components/AgentChat/ChatCommon/ChatHistory.tsx:25]: I'm not sure. What is the bug?

File: packages/ui-common/components/AgentChat/ChatCommon/ChatHistory.tsx [hunk 64]
```diff
@@ -0,0 +1,70 @@
...
+    <MUIAccordion
+        id={id}
+        key={chatHistoryKey}
+        sx={{marginBottom: "2rem", marginTop: "1rem"}}
+        items={[
+            {
+                title: "Chat History",
+                content: messages.map((message) => {
+                    const itemKey = hashString(message.text + message.type + message.id)
+                    if (message.type.toUpperCase() === ChatMessageType.HUMAN) {
```
swensel, line 44: Do we want any other [KNOWN_MESSAGE_TYPES](https://github.com/cognizant-ai-lab/neuro-san-ui/blob/main/packages/ui-common/components/AgentChat/Common/Utils.ts#L24)?

swensel (dismissed):
LGTM. Good work.

I do like Zen Mode default on, but I'm flexible. I just think with this default change, that there are too many elements on the right side with the chat window, the thought bubbles (and their connecting lines) when `isAwaitingLlm` is `true`. However, like you said, the user can change it in Settings.

packages/ui-common/components/AgentChat/ChatCommon/ChatHistory.tsx
dsargent: `userImage={userImage}` for AI messages. Will show human avatar.

packages/ui-common/components/AgentChat/ChatCommon/ChatHistory.tsx
dsargent: I considered that. I think of the chat history recap as  a "low fidelity" reminder of the conversation. It doesn't represen the actual context that we send back to the agents, which is more extensive. So for now, I think these two types are fine for giving a recap for the user.

packages/ui-common/state/Settings.ts
dsargent: Interesting...I find it really annoying as a developer, especially when working with AND. I don't like the way everything disappears and suddenly I'm in deep space. No way to please everyone with this. But once you set it, it will be remembered for you until you clear local storage or something similar.

dsargent:
> LGTM. Good work.
> 
> I do like Zen Mode default on, but I'm flexible. I just think with this default change, that there are too many elements on the right side with the chat window, the thought bubbles (and their connecting lines) when `isAwaitingLlm` is `true`. However, like you said, the user can change it in Settings.

Maybe default thought bubbles to off, and zen mode off? I'm not sure what would be the best UX for new users. Those of us "in the know" can turn those elements on and off as we want, but other users may not realize.

swensel (approved):

PR #280, opened 2026-05-05
Fix: Resolve MUI hydration mismatch caused by server/client theme inconsistency

souravJBcognizant:
Root Cause

`createAppTheme` internally calls `cssVar()`, which reads CSS custom properties (like --bs-primary) from the browser's computed styles via` getComputedStyle(document.documentElement)`. These variables are defined in `globals.css` but the server (Node.js) has no browser, no CSS engine, and never parses that file. As a result, `cssVar()` returns a hardcoded fallback value on the server but the real CSS variable values on the client.

Due to this, the server and client produce different CSS strings, different class name hashes, and throws a hydration mismatch error.

Fix

Replaced `useMemo` with `useState + useEffect` in `_app.tsx`:

`useState(createTheme)` initializes with MUI's default theme using no arguments . Both server and client produce identical output during the initial render, so hydration succeeds.
`useEffect `runs exclusively on the client after hydration completes. At that point the browser has parsed `globals.css`, `cssVar()` returns correct values, and `createAppTheme` is called with the real branding values. This triggers a normal client-side re-render with no hydration involvement.

### DIFF
--- a/apps/main/pages/_app.tsx
+++ b/apps/main/pages/_app.tsx
@@ -127,7 +127,7 @@ export default function NeuroSanUI({Component, pageProps}: ExtendedAppProps): Re
     const background = useSettingsStore((state) => state.settings.branding.background)
 
     const [theme, setTheme] = useState(createTheme)
-    // generating the branded themes only in client-side to avoid hydration issues.
+    // Generate the branded theme only on the client to avoid hydration issues.
     useEffect(() => {
         setTheme(createAppTheme(primary, secondary, background))
     }, [primary, secondary, background])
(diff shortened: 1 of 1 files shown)


dsargent (dismissed):
I like it, if this fixes the hydration issue. Seems like a better way to handle the dynamic theme creation, and not just a Band-Aid to make the ugly warning go away.

swensel (dismissed):
LGTM.

apps/main/pages/_app.tsx
```diff
@@ -126,7 +126,11 @@ export default function NeuroSanUI({Component, pageProps}: ExtendedAppProps): Re
     const secondary = useSettingsStore((state) => state.settings.branding.secondary)
     const background = useSettingsStore((state) => state.settings.branding.background)
 
-    const theme = useMemo(() => createAppTheme(primary, secondary, background), [primary, secondary, background])
+    const [theme, setTheme] = useState(() => createTheme())
```
dsargent: Please add the generic param to `useState<...here...>`.

dsargent (approved):

PR #285, opened 2026-05-07
#276 Implement shared ESLint config

dsargent:
[Copilot summary](https://github.com/cognizant-ai-lab/neuro-san-ui/pull/285#pullrequestreview-4246913617).

Low risk PR: only affects linting infrastructure, not production code.

Adds a new package, `dev-common`, which will contain more shared "dev" items in future like configs, maybe scripts, depending on how things evolve. The idea being to avoid having to copy paste configs across projects.

First customer -- this repo itself. Changed the root ESLint config to depend on the shared one, as proof of concept and "dogfooding".

I tested it locally with Neuro UI and it worked fine. I was able to reduce Neuro UI's ESLint config to 40 lines (from 568 lines) by depending on this new package.

**Nice side effect:** after I applied and tested Copilot's recommendations for how to set up the dependencies, yarn is now _much_ happier on `install`. Only one warning remains:

`➤ YN0060: │ eslint is listed by your project with version 10.2.1 (p8b6f96), which doesn't satisfy what eslint-plugin-import and other dependencies request (but they have non-overlapping ranges!).`

This is a known issue with `es-plugin-import` -- they are dragging their feet releasing an "official" ESLint 10 version, but it works just fine as-is. It's just that their peer dependency is out of date. Related [ticket](https://github.com/import-js/eslint-plugin-import/issues/3227). We may want to switch to [`eslint-plugin-import-x](https://github.com/un-ts/eslint-plugin-import-x#why)`, a well-maintained fork that addresses many of the issues with the `eslint-plugin-import` library that the maintainers don't seem to care about. Quote from the maintainers of the one we're using:

>  there's simply no rush whatsoever to upgrade to a new version of your linter, and when all of the ecosystem packages you use have support for v10, then (and not before) you'll be able to upgrade.

Out of scope for this PR: publishing logic for the new package. Will address that in a separate PR.


packages/dev-common/Configs/eslint.config.mjs
```diff
@@ -0,0 +1,601 @@
...
+            ecmaVersion: 2022,
+            sourceType: "module",
+
+            parserOptions: {
+                ecmaFeatures: {jsx: true},
+                projectService: {
+                    allowDefaultProject: ["*.mjs", "*.ts"],
+                    defaultProject: "../../../tsconfig.json",
+                },
+                tsconfigRootDir: import.meta.dirname,
```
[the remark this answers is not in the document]
dsargent: Not touching it. This is how Copilot said to do it and after a lot of messing around, finally found this one that works. And also works from Neuro UI which does consume the package from `node_modules`.

eslint.config.mjs
```diff
@@ -1,585 +1,21 @@
...
-    // process and doesn't *need* to be in scope.
-    eslintPluginReact.configs.flat["jsx-runtime"],
-
-    eslintConfigPrettier,
+import sharedConfig from "./packages/dev-common/Configs/eslint.config.mjs"
 
+/** @type {import("eslint").Linter.Config[]} */
+export default defineConfig([
+    ...sharedConfig,
     {
```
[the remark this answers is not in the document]
dsargent: Good for future, but for now _all_ our imports work this way, importing via relative path rather than package spec. But maybe we should change them all? 🤔 And have an "imports" linter rule to enforce it. Any opinion, @swensel ?

knip.config.ts
```diff
@@ -85,6 +88,19 @@ const config: KnipConfig = {
...
+        "eslint-config-prettier",
+        "eslint-plugin-jest",
+        "eslint-plugin-jest-dom",
+        "eslint-plugin-prefer-arrow-functions",
+        "eslint-plugin-react",
+        "eslint-plugin-react-hooks",
+        "eslint-plugin-testing-library",
+        "eslint-plugin-unicorn",
+        "typescript-eslint",
     ],
```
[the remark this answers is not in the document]
dsargent: `knip` is happy, so not changing anything in this precarious edifice of configurations that I finally got to work and placate yarn, knip, tsc, eslint and jest simultaneously.

packages/dev-common/Configs/eslint.config.d.mts
dsargent, line 1: New types for the MJS config file to keep Typescript happy in the consuming project.

packages/dev-common/Configs/eslint.config.mjs
dsargent, line 1: Almost all config moved to here. This is what consuming projects will be using.

knip.config.ts
```diff
@@ -85,6 +88,19 @@ const config: KnipConfig = {
 
         // Used by Jest for TS format config file
         "ts-node",
+
+        "@eslint/compat",
```
dsargent, line 93: I do _not_ like adding ignores to Knip as they "blind" it, but according to Copilot this is the correct and necessary thing to do; apparently Knip itself has a blindspot about peer dependencies of sub packages and it's a know issue. In Copilot's words,

> Yes, this is an inherent tension with the setup and the ignoreDependencies list is the correct solution. The situation is:
dev-common correctly uses peerDependencies because it's published externally
The root must install those peers as devDependencies so ESLint works when run from the root
Knip can't trace "these are needed because they satisfy the peer deps of a local workspace package" — that's a known gap in knip's peer dep awareness across workspace boundaries
There's no cleaner alternative given the constraints.

apps/main/package.json
```diff
@@ -33,6 +33,7 @@
         "sharp": "0.33.2"
     },
     "devDependencies": {
+        "@testing-library/dom": "10.4.1",
```
dsargent, line 36: Required here as it's used in this package. That was one of the `yarn install` errors and `yarn` was correct.

packages/dev-common/package.json
dsargent, line 1: For the new package.

packages/dev-common/package.json
```diff
@@ -0,0 +1,38 @@
+{
+    "name": "@cognizant-ai-lab/dev-common",
+    "description": "Common development configurations and utilities for Cognizant AI Labs UI projects",
+    "version": "1.0.0",
+    "type": "module",
+    "peerDependencies": {
```
dsargent, line 6: Copilot assures me this is correct and idiomatic, having these all as peer dependencies with wildcarded version. I'm out of my depth with this monorepo stuff so I'm leaning heavily on Copilot.

packages/dev-common/package.json
```diff
@@ -0,0 +1,38 @@
...
+    },
+    "repository": {
+        "type": "git",
+        "url": "git+https://github.com/cognizant-ai-lab/neuro-san-ui.git"
+    },
+    "publishConfig": {
+        "registry": "https://registry.npmjs.org/",
+        "access": "public"
+    },
+    "exports": {
```
dsargent, line 32: This allows other projects to import the goodies.

packages/dev-common/README.md
dsargent, line 1: Placeholder for now.

packages/ui-common/index.ts
```diff
@@ -53,6 +53,7 @@ export * from "./state/Environment"
 export * from "./state/Settings"
 export * from "./state/UserInfo"
 export * from "./utils/Authentication"
+export * from "./utils/File"
```
dsargent, line 56: Forgot to export this previously. Unrelated to this PR but I wanted to get it done before the next `ui-common` release so we can remove the "file utils" from Neuro AI.

packages/ui-common/package.json
```diff
@@ -1,5 +1,6 @@
 {
     "name": "@cognizant-ai-lab/ui-common",
+    "description": "A shared UI component library for Cognizant AI Labs projects",
```
dsargent, line 3: Added description -- didn't have one.

packages/ui-common/package.json
```diff
@@ -26,6 +27,7 @@
     "devDependencies": {
         "@babel/core": "7.23.7",
         "@babel/preset-env": "7.28.3",
+        "@testing-library/dom": "10.4.1",
```
dsargent, line 30: Missing dep

eslint.config.mjs
dsargent, line 1: Guts of this moved to the shared config. This is how lean the ESLint of consuming projects can be.

eslint.config.mjs
```diff
@@ -1,585 +1,21 @@
...
-            "import/resolver": {
-                node: true,
-                typescript: {
-                    noWarnOnMultipleProjects: true,
-                    project: ["./tsconfig.json", "./apps/*/tsconfig.json", "./packages/*/tsconfig.json"],
-                },
-            },
             next: {
-                rootDir: ["apps/*/"],
+                rootDir: "apps/main",
```
dsargent, line 13: Consuming projects will only need to specify things like this -- "Here's where my Next.js lives" -- since obviously that can't be in the shared config. Paths, maybe file types etc. are going to differ from project to project, so consumers will have to specify those.

jest.config.ts
```diff
@@ -50,21 +50,21 @@ const config: Config.InitialOptions = {
                 tsconfig: "<rootDir>/tsconfig.test.json",
             },
         ],
-        "^.+\\.[jt]sx?$": ["babel-jest", {configFile: "./babel.jest.config.cjs"}],
+        "^.+\\.(js|jsx)$": ["babel-jest", {configFile: "./babel.jest.config.cjs"}],
```
dsargent, line 53: We already took care of the TS case in the previous pattern, so we can be explicit here: "this rule is only for js/jsx files" (even though we currently don't have many/any files like that in this project.)

jest.config.ts
```diff
@@ -50,21 +50,21 @@ const config: Config.InitialOptions = {
...
     },
     verbose: false,
     setupFilesAfterEnv: ["<rootDir>/jest.setup.ts"],
     testEnvironment: "jsdom",
     collectCoverage: true,
     collectCoverageFrom: [
         "**/*.{js,jsx,ts,tsx}",
+        "!**/.next/**",
```
dsargent, line 61: Alphabetic sort + recursive glob for Knip -- future PR will do the same centralized config refactoring for Knip, so prepare for it here.

knip.config.ts
```diff
@@ -58,6 +58,9 @@ const config: KnipConfig = {
 
         // Used by CommitCheck script
         "jest_quiet.config.ts",
+
+        // Generated type declaration for published dev-common eslint config; consumed by external projects
```
dsargent, line 62: False positive in knip; it has no way to know that external projects will be consuming this so we have to put this "trust me, bro" here.

package.json
```diff
@@ -30,9 +30,9 @@
         "@babel/preset-env": "7.29.2",
         "@eslint/compat": "2.0.5",
         "@eslint/js": "10.0.1",
+        "@jest/globals": "30.3.0",
```
dsargent, line 33: Another missing dep. Gets rid of another yarn warning.

package.json
```diff
@@ -30,9 +30,9 @@
...
         "@jest/types": "30.3.0",
         "@next/eslint-plugin-next": "16.2.4",
-        "@testing-library/dom": "10.4.1",
```
dsargent, line 35: Apparently not used in the root project, so remove it from here.

packages/dev-common/package.json
swensel: I don't know the right answer here either. May dive into that more at some point.

swensel (approved):
LGTM. Thanks for doing this.

PR #289, opened 2026-05-07
#288 Use yarn catalogs for consistent versioning

dsargent:
[Copilot summary](https://github.com/cognizant-ai-lab/neuro-san-ui/pull/289#pullrequestreview-4248077500).

Opinionated PR.

Centralizes dependency versions across the mono repo into a new yarn [catalog](https://yarnpkg.com/features/catalogs), defined in `.yarnrc.yml`.

Now, common versions are specified like this:

`"foo": "catalog:default"`

which means, "use whatever the catalog says". When upgrading deps, we upgrade in _one_ place only, `.yarnrc.yml`. 

Upsides (major):
* All packages in the monorepo use same versions of dependencies
* Dependencies all visible in one place (`.yarnrc.yml`)
* Upgrades are trivial: just upgrade in one place and all packages will pick it up
* No chance of packages within monorepo "drifting" and having inconsistent versions of dependencies. Not theoretical: some already had drifted, as seen in this PR!
* Potential to share the catalog with other repos (Neuro UI, NSFlow...) to enforce consistent versions across repos. Perhaps publishing via `dev-common`. Needs investigation.

Downsides:
* Yarn-specific. Other package managers (should we wish to migrate someday) have equivalent or close-ish mechanisms. But I've never been one to avoid _using_ the tools we have "in case we migrate".
    * npm: No direct equivalent. You'd use an .npmrc with overrides in package.json to force a single version, but there's no named catalog concept.
    * pnpm: catalogs — pnpm 9.5+ has a native catalog feature in pnpm-workspace.yaml, which is exactly what Yarn catalogs are modeled after.
* Syntax takes getting used to when looking at `package.json`. Instead of seeing `"foo": "123"` you see `"foo": "catalog:default"`
* Can't immediately see versions in `package.json`. But that's kind of the point -- the handling of versions is taken out of the `package.json` files and centralized in the yarn catalog.

Alternative solution: [yarn constraints](https://yarnpkg.com/features/constraints). I looked into it and although they are powerful and super flexible, you're basically writing more code just to ensure consistency across the monorepo.

I'm pondering if we can move the catalog to `dev-common` so that other projects (LMX, Neuro UI) can use it as well, for version syncing across repos. This current PR only addresses that _within_ the monorepo.

I already found some versions within this monorepo that had drifted apart, probably unintentionally. See comments on individual files.

Known issue: this very annoying [bug](https://github.com/yarnpkg/berry/issues/6925) in yarn catalogs prevents us from using the consolidated versions in peer dependencies. Once they fix it, we should use the catalog version there too.


apps/main/pages/_app.tsx
dsargent, line 1: Recommend you start reviewing by looking at `.yarnrc.yml`. Everything flows from there.

apps/main/pages/_app.tsx
```diff
@@ -14,8 +14,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import "@xyflow/react/dist/style.css"
-
 import "../styles/globals.css"
```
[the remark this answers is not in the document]
dsargent: Contradicts what local Copilot says. I don't get the rationale here; only `AgentFlow` requires this stylesheet, so why not import it just there?

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -13,6 +13,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
+import "@xyflow/react/dist/style.css"
```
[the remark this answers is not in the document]
dsargent: Local Copilot says false alarm, and testing/building reveals no issues, so...ignoring this.

apps/main/pages/_app.tsx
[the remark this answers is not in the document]
dsargent: @swensel thoughts? I prefer having the CSS import only in the file that uses it (I'm a minimalist and I like efficiency) but Copilot is concerned...I didn't see any issues when testing.

apps/main/package.json
dsargent, line 1: Use catalog throughout

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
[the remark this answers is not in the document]
dsargent: Again..same thing with importing CSS locally. @swensel ?

packages/ui-common/package.json
dsargent, line 1: Use catalog throughout.

.yarnrc.yml
dsargent, line 1: Home of the catalog.

I moved versions to the catalog that either (1) common across workspaces in the monorepo (or should be!) or (2) "sort of belong together" like the various jest dependencies. We can move more versions to the catalog in future to centralize them.

Some versions had already drifted apart within this single monorepo! I took the max of the ones we were using and centralized here.

Note: I implemented it as `catalogs` with `default` being the first and only one currently. Originally I had one for `default`, one for `range` which would be for peer deps like `^1.2.3`, _but_ due to the yarn bug mentioned elsewhere, we can't use the catalog in peer deps yet. But I kept it as "catalogs", anticipating that once they fix the bugs we can update to use the catalog for peer deps too.

.yarnrc.yml
```diff
@@ -1 +1,37 @@
 nodeLinker: node-modules
```
dsargent, line 1: We are "fighting yarn" by forcing it to use `node_modules` still, which they consider "legacy" as of Berry. But upgrading to the new yarn PnP model is not trivial...and some tools may break if they are hardcoded for `node_modules`. Created a [ticket](https://github.com/cognizant-ai-lab/neuro-san-ui/issues/311).

package.json
dsargent, line 1: Use catalog throughout.

yarn.lock
dsargent, line 1: For once, it might be worth skimming this to make sure it's still sane after the catalog changes.

apps/main/pages/_app.tsx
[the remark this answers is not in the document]
swensel: Sounds like the issue could happen when switching between pages, but MAUI only has one page. I'd probably leave it as is in the existing codebase, but since you didn't notice any issues, it could be changed.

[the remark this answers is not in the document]
dsargent: I'll leave it as-is in this PR since I didn't see any issues. We're probably already heavy on CSS due to all the MUI-generated classes.

yarn.lock
swensel: I think it's worth regenerating this time?

swensel (dismissed):
LGTM, although you could regen `yarn.lock`, if not already.

packages/ui-common/package.json
```diff
@@ -11,40 +11,42 @@
     },
     "dependencies": {
         "@dagrejs/dagre": "1.0.4",
-        "@langchain/core": "1.1.17",
+        "@langchain/core": "catalog:default",
```
swensel, line 14: It is good that these can live in one place now.

.yarnrc.yml
swensel: Yes I think we are going to run into a number of issues when trying that. I played with it on my local a bit, when upgrading Yarn, and it didn't go smoothly. Also even if we iron out all the issues at a given time, it opens us up to potential compatibility issues in the future if we want to use a certain library or build tool (or if a customer implementation wants to use something). I wonder if, for compatibility reasons, we should keep using `node_modules` for a while, even though I would prefer to be following the latest recommendation.

dsargent: Agreed. No huge benefits to using PnP, other than saving disk space due to using a central repo for deps (meh, disk space is cheap) though I believe PnP also helps with avoiding "ghost dependencies". Requires investigation, no immediate rush.

swensel (approved):

Issue #60, opened 2025-10-20
Optimize plasma animation


dsargent:
**Migrated from Jira:** [UN-3439](https://leaf-ai.atlassian.net/browse/UN-3439)

## Jira Details
- **Issue Type:** Bug
- **Priority:** Major
- **Status:** To Do
- **Reporter:** Darren Sargent
- **Assignee:** Sean Wensel
- **Created:** 2025-08-12 10:05:09
- **Updated:** 2025-08-18 14:16:08

PR #295, opened 2026-05-09
#60 Optimize plasma animation

dsargent:
Attempt to optimize the plasma animation, with help from Copilot.


dsargent:
@swensel tried it and didn't make a huge difference. Will revisit later.

PR #298, opened 2026-05-11
Shared knip config

dsargent:
Low risk PR, as it only affects knip.

Similarly to [ESLint](https://github.com/cognizant-ai-lab/neuro-san-ui/pull/285), this PR consolidates the common knip configuration items into a shared config in `dev-common`.

Not as huge a win as ESLint for reducing config file copy pasta, but a good foundation to build on as we evolve common rulesets.

### DIFF
--- a/apps/main/package.json
+++ b/apps/main/package.json
@@ -38,7 +38,6 @@
         "@testing-library/user-event": "14.6.1",
         "@types/lodash-es": "4.17.12",
         "@types/react": "19.2.14",
-        "eslint": "10.2.1",
         "node-mocks-http": "1.17.2",
         "typescript": "5.9.3"
     }
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -31,7 +31,6 @@ const esmPackagesToTransform = [
     "mdast.*",
     "parse-.*",
     "property-.*",
-    "react-syntax-highlighter",
     "refractor",
     "space-.*",
 ]
--- a/knip.config.ts
+++ b/knip.config.ts
@@ -23,13 +23,10 @@ import type {KnipConfig} from "knip"
 
 import {config as baseConfig} from "./packages/dev-common/Configs/knip.config"
 
-const resolvedBase = typeof baseConfig === "function" ? null : baseConfig
-
 const config: KnipConfig = {
-    // eslint-disable-next-line @typescript-eslint/no-misused-spread -- we know the baseConfig is not a function
     ...baseConfig,
     ignore: [
-        ...(resolvedBase.ignore as string[]),
+        ...(baseConfig.ignore as string[]),
 
         // Temporarily exclude for transition to monorepo (legit issue)
         "packages/ui-common/components/AgentChat/Common/Types.ts",
@@ -38,7 +35,8 @@ const config: KnipConfig = {
         "jest_quiet.config.ts",
     ],
     ignoreDependencies: [
-        ...resolvedBase.ignoreDependencies,
+        ...baseConfig.ignoreDependencies,
+
         // Used for Speech Recognition API types
         "@types/dom-speech-recognition",
 
--- a/knip.config.ts
+++ b/knip.config.ts
@@ -28,6 +28,9 @@ const config: KnipConfig = {
     ignore: [
         ...(baseConfig.ignore as string[]),
 
+        // Used in a sneaky way by jest
+        "babel.jest.config.cjs",
+
         // Temporarily exclude for transition to monorepo (legit issue)
         "packages/ui-common/components/AgentChat/Common/Types.ts",
 
@@ -37,8 +40,19 @@ const config: KnipConfig = {
     ignoreDependencies: [
         ...baseConfig.ignoreDependencies,
 
+        // Used internally by eslint
+        "globals",
+
+        // Used by jest
+        "@babel/core",
+        "@babel/preset-env",
+        "babel-jest",
+
         // Used for Speech Recognition API types
         "@types/dom-speech-recognition",
+
+        // Used internally by eslint
+        "globals",
 
         // Used by do_openapi_generate.sh
         "openapi-typescript",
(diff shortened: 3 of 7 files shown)


apps/main/package.json
```diff
@@ -38,7 +38,6 @@
         "@testing-library/user-event": "14.6.1",
         "@types/lodash-es": "4.17.12",
         "@types/react": "19.2.14",
-        "eslint": "10.2.1",
```
dsargent, line 41: Not needed here since we run knip from the monorepo root, I _think_.

packages/dev-common/Configs/eslint.config.mjs
```diff
@@ -100,10 +100,9 @@ export default defineConfig([
             parserOptions: {
                 ecmaFeatures: {jsx: true},
                 projectService: {
-                    allowDefaultProject: ["*.mjs", "*.ts"],
+                    allowDefaultProject: ["knip.config.ts"],
```
dsargent, line 103: After lots of vibing and Copilot wasting my time with hallucinations, this seems to be the incantation that works. This config stuff is waaaay more complicated than it should be. They totally overengineered it.

packages/dev-common/Configs/knip.config.ts
dsargent, line 1: New shared Knip config. I tested it with Neuro UI too and it works fine. Will be a separate PR to opt-in to it over there.

packages/dev-common/package.json
```diff
@@ -19,6 +19,7 @@
         "eslint-plugin-testing-library": "*",
         "eslint-plugin-unicorn": "*",
         "globals": "*",
+        "knip": "*",
```
dsargent, line 22: Now needed as a peer dep for the types in the shared knip config file.

packages/dev-common/package.json
```diff
@@ -33,6 +34,10 @@
         "./eslint": {
             "types": "./Configs/eslint.config.d.mts",
             "default": "./Configs/eslint.config.mjs"
+        },
+        "./knip": {
```
dsargent, line 38: Expose the shared config.

jest.config.ts
```diff
@@ -31,7 +31,6 @@ const esmPackagesToTransform = [
     "mdast.*",
     "parse-.*",
     "property-.*",
-    "react-syntax-highlighter",
```
dsargent, line 34: Doesn't seem to be needed any more and I'm all for getting rid of unnecessary transpilations. Probably we upgraded at some point to a newer version that comes with ESM exports so transpilation no longer needed.

knip.config.ts
dsargent, line 1: Opt-in to shared config. This gives an idea of how much boilerplate can be removed.

packages/dev-common/Configs/eslint.config.mjs
```diff
@@ -100,10 +100,9 @@ export default defineConfig([
...
                     defaultProject: "../../../tsconfig.json",
                 },
-                tsconfigRootDir: import.meta.dirname,
```
dsargent, line 106: Doesn't seem to be needed.

packages/dev-common/Configs/eslint.config.mjs
```diff
@@ -598,4 +597,12 @@ export default defineConfig([
             "@typescript-eslint/no-empty-interface": "error",
         },
     },
+    {
+        files: ["**/eslint.config.mjs"],
```
dsargent, line 601: Had to add this (at Copilot's suggestion) to keep ESLint happy.

packages/dev-common/Configs/knip.config.ts
```diff
@@ -0,0 +1,62 @@
+/*
+Copyright 2025 Cognizant Technology Solutions Corp, www.cognizant.com.
```
dsargent, line 2: Don't worry about the year, we will fix these en masse.

packages/dev-common/Configs/knip.config.ts
```diff
@@ -0,0 +1,62 @@
...
+
+import type {KnipConfig} from "knip"
+
+/**
+ * @fileoverview Knip configuration file to identify unused files and dependencies in the project.
+ *
+ */
+
+/**
+ * Knip's configuration type is a bit complex, but we know that the config object we are creating here is just a
```
dsargent, line 25: See comment.

Knip doesn't seem to offer a shared/layered/inherited config mechanism à la tsc `extends` or ESLint's flat config with spreading. So I had to roll my own. Would be very happy to be wrong here, but couldn't find anything in their GitHub issues about this.

Worse, their standard config object, `KnipConfig`, is a weird union that can be either normal name-value props, _or_ a function that returns a config, _or_ an async function that returns a Promise. Reminds me of how complex the typing is for React state setters/getters. Anyway, to avoid consumer projects having to deal with that, I manually narrow the type here to "just the name-value version of the config" since we know that's what we're using here. To make matters worse, Knip doesn't properly expose what we _really_ want to use, `RawConfiguration` (they expose it but no types or something?! IDE and tsc complain) so I had to use `Extract` to get something of the same shape. That way consumers can happily (and safely!) spread it without getting errors, simulating "inheritance" or layered config.

packages/dev-common/Configs/knip.config.ts
```diff
@@ -0,0 +1,62 @@
...
+    // From the doc:
+    // "By default, Knip does not report unused exports in entry files. When a repository (or workspace) is
+    // self-contained or private, you may want to include entry files when reporting unused exports:"
+    includeEntryExports: true,
+
+    // Treat hints as errors (will make exit code non-zero)
+    treatConfigHintsAsErrors: true,
+
+    // Opt-in to all issues types
+    include: [
```
dsargent, line 41: There isn't much to the shared config, but at least having these here means consumer projects will be opted-in to all the checks.

knip.config.ts
```diff
@@ -21,35 +21,13 @@ limitations under the License.
 
 import type {KnipConfig} from "knip"
 
-const config: KnipConfig = {
-    // From the doc:
```
dsargent, line 25: Nice to remove al lthe copy-pasta.

swensel (approved):
LGTM

Issue #293, opened 2026-05-09
User guide does not play nicely with Footer


dsargent:
See screenshot

<img width="1747" height="903" alt="Image" src="https://github.com/user-attachments/assets/42e6adaa-30c0-4678-af29-9ede72fd611e" />

PR #300, opened 2026-05-11
fix footer issue in User Guide

souravJBcognizant:
Issue #293 

On `/userguide`, the footer rendered in the middle of the markdown content instead of at the bottom.

Root Cause

In `pages/_app.tsx` every page is wrapped in `#unileaf`, which stacks three things vertically: `Navbar → #body-container → Footer.
`
`#unileaf` is locked at the viewport height (height: 100% in `styles/globals.css`, and `#body-container` has `flex: 1`, meaning it gets exactly the leftover space (100vh − navbar − footer).

The user guide markdown is much taller than that leftover space. Since `#body-container` can't grow (its parent is locked), the markdown spills out visually, but the box's layout size stays the same. The Footer is positioned right after the box's official end — so it lands mid-page, with the overflowing markdown rendered around it.

Fix

Made `#body-container` in `_app.tsx` a flex column so every route gets a proper flex parent. This replaces the previous approach of page-scoped `GlobalStyles`( in the same PR) workaround with an app-wide layout change.

`_app.tsx`

Updated #body-container:

```
display: "flex"
flexDirection: "column"
flex: 1
minHeight: "90%"
```
minHeight: "90%" is preserved to keep the Multi-Agent Accelerator filling the viewport and the footer below the fold(revealed on scroll).

`userguide.tsx`

Added layout properties via sx on the page's StyledMarkdownContainer:

`sx={{flex: 1, minHeight: 0, overflowY: "auto", paddingRight: "1rem"}}`
flex: 1 claims the remaining vertical space inside #body-container, so the markdown scrolls internally.

before:
image

after:
image


apps/main/pages/userguide.tsx
```diff
@@ -57,14 +66,17 @@ export default function UserGuide() {
...
-        <StyledMarkdownContainer id="user-guide-container">
-            <ReactMarkdown
-                rehypePlugins={[rehypeRaw, rehypeSlug]}
-                remarkPlugins={[[remarkToc, {heading: "Table of Contents", tight: true}]]}
-            >
-                {userGuide}
-            </ReactMarkdown>
-        </StyledMarkdownContainer>
+        <>
+            <GlobalStyles styles={USER_GUIDE_BODY_LAYOUT} />
```
dsargent: Why `GlobalStyles` here? We don't use that anywhere else.

Perhaps simpler, create a local styled component based on `StyledMarkdownContainer` (or modify `StyledMarkdownContainer` itself) and get rid of this const.

apps/main/styles/StyledMarkdownContainer.ts
```diff
@@ -20,6 +20,10 @@ import {styled} from "@mui/material/styles"
 export const StyledMarkdownContainer = styled(Box)(() => ({
     fontSize: "1rem",
     lineHeight: 1.75,
+    flex: 1,
```
dsargent: Note that this component is exported and consumed by Neuro UI as well so be careful of breaking things there. https://github.com/cognizant-ai-lab/neuro-ui

apps/main/pages/userguide.tsx
souravJBcognizant: I used the `GlobalStyles` to specifically target the `#bodycontainer` of _app.tsx.
the `StyledMarkdownContainer` is rendered as a child of `_app.tsx`'s `#body-container` so it always inherits that parent's layout. Modifying `StyledMarkdownContainer` wont solve the issue. 
We will need to either inject CSS to this particular route `/userguide` using `GlobalStyles` or change the `#bodycontainer` of `_app.tsx`(which will affect all pages).

souravJBcognizant: One other solution could be defining layout for individual pages using Per-Page Layouts ( #[link](https://nextjs.org/docs/pages/building-your-application/routing/pages-and-layouts#per-page-layouts))

apps/main/styles/StyledMarkdownContainer.ts
souravJBcognizant: moved the styles to `userguide.tsx`

dsargent (approved):
lgtm. @swensel does it look okay to you?

dsargent:
And thanks for adding screenshots. It helps.

swensel:
Hi @souravJBcognizant I'm not a huge fan of having two scrollbars on the user guide page (one to scroll through the user guide and one to scroll to the footer).

Screenshot 2026-05-13 at 7 31 32 PM

I think it could be better to show the footer as fixed on the user guide (and just have one scrollbar for the user guide), but I realize that would require a conditional case, since the footer is not fixed on other pages.

Feel free to move forward with this change. The user guide is probably not a frequently visited page. We may want to revise this in the future. Thanks.

cc: @dsargent

swensel (approved):

swensel:
@souravJBcognizant Alternatively, the user guide could also be available at the bottom of the user guide, which is essentially what we do on the other pages.

souravJBcognizant:
> @souravJBcognizant Alternatively, the user guide could also be available at the bottom of the user guide, which is essentially what we do on the other pages.

@swensel I actually tried that but it involves changing `html`'s `height:100%` to `min-height:100%` in `global.css`, and that doesn't behave well with `/multiAgentAccelerator` and` / `, both rely on definite-height ancestors and end up collapsing.

Issue #320, opened 2026-05-13
Implement "vibe-based editing" of AND networks (follow up on PR feedback)


swensel:
Follow up from https://github.com/cognizant-ai-lab/neuro-san-ui/pull/274

Issue #287, opened 2026-05-07
Atone for lapses in test coverage (Sean)


dsargent:
Handling this in #469 and #470.

PR #321, opened 2026-05-13
#320 - Implement "vibe-based editing" of AND networks (follow up on PR feedback)

swensel:
# Description

- Follow up from https://github.com/cognizant-ai-lab/neuro-san-ui/pull/274. Addressed most of the feedback. Anything remaining I have left as unresolved in that PR and I'll revisit that in a future PR.
- Fix bug where AND overwrites temp network edits.


packages/ui-common/components/MultiAgentAccelerator/AgentNodePopup.tsx
```diff
@@ -44,10 +42,7 @@ export interface AgentNodePopupProps {
     readonly initialInstructions?: string
     /** Initial description text shown in the editable field. Defaults to an empty string. */
     readonly initialDescription?: string
-    /**
-     * When true the dialog is in a saving state: the Save button is disabled and shows a spinner.
-     * Defaults to false.
-     */
+    /** When true the save API call is in-flight; shows an "Applying changes..." button and disables both actions. */
     readonly isSaving?: boolean
```
swensel, line 47: There was some feedback to use `isAwaitingLlm` instead of `isSaving`, but `isAwaitingLlm` is really more for the chat interface (and it triggers zen mode, if enabled, which we don't want, when saving vibe edits from the popup). I did try out that switch and it just didn't work well.

packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
```diff
@@ -1298,6 +1298,6 @@ describe("ChatCommon", () => {
         await sendQuery(TEST_AGENT_MATH_GUY, "test sly_data accumulation on AGENT message")
 
         // sendChatQuery was called — the component processed the chunk without error
-        expect(sendChatQuery).toHaveBeenCalled()
+        expect(sendChatQuery).toHaveBeenCalledTimes(1)
```
swensel, line 1301: There was a suggestion to check that this was only called once.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
```diff
@@ -1100,15 +1099,13 @@ describe("AgentFlow", () => {
...
             isSelectedNetworkTemporary: true,
             networkId: networkKey,
         })
 
-        // Click an agent node to open the popup. ReactFlow renders nodes with data-id attributes.
-        const agent1Node = container.querySelector('[data-id="agent1"]')
-        expect(agent1Node).toBeInTheDocument()
-        fireEvent.click(agent1Node)
+        // Click an agent node to open the popup, querying by the visible agent name.
+        fireEvent.click(screen.getByText(cleanUpAgentName("agent1")))
```
swensel, line 1108: Unfortunately if I switch these to `await user.click`, I see:

```
TypeError: Cannot read properties of null (reading 'document')

    at _e (node_modules/@xyflow/react/dist/umd/index.js:1:14514)
    at HTMLDivElement.g (node_modules/@xyflow/react/dist/umd/index.js:1:16236)
    at HTMLDivElement.<anonymous> (node_modules/@xyflow/react/dist/umd/index.js:1:6830)
```

Seems to be a Reactflow issue.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentNodePopup.test.tsx
```diff
@@ -322,18 +322,6 @@ describe("AgentNodePopup", () => {
             expect(screen.getByRole("button", {name: /close/iu})).toBeInTheDocument()
         })
 
-        it("shows a progress bar while isSaving is true", () => {
```
swensel, line 325: Removed progress bar since it was vibed and a little amateurish.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
```diff
@@ -1545,45 +1540,17 @@ describe("AgentFlow", () => {
         expect(mockSetThoughtBubbleEdges).toHaveBeenCalled()
     })
 
-    describe("popup save — async network replacement", () => {
```
swensel, line 1548: Logic for merging networks, [and the associated tests](https://github.com/cognizant-ai-lab/neuro-san-ui/pull/321/changes#diff-5cac029c8e6ff5221955c94b836b420eaa94fee0851b29d5cf90e0b2b64ff449R225), were moved to [mergeNetworks](https://github.com/cognizant-ai-lab/neuro-san-ui/pull/321/changes#diff-aa790e1f7999c0f4cef8115c3b371df8ff9f80875a8aa654c14e5cda2056a4f1R129) in `packages/ui-common/components/MultiAgentAccelerator/TemporaryNetworks.ts`.

packages/ui-common/state/TemporaryNetworks.ts
```diff
@@ -87,28 +63,32 @@ export const useTempNetworksStore = create<TempNetworksStore>()(
...
+            upsertTempNetworks: (incomingNetworks: TemporaryNetwork[]): TemporaryNetwork[] => {
                 set((state) => {
                     const updated = [...state.tempNetworks]
-                    for (const newNetwork of newNetworks) {
-                        const newName = effectiveNetworkName(newNetwork)
-                        const existingIdx = newName ? updated.findIndex((n) => effectiveNetworkName(n) === newName) : -1
-                        if (existingIdx >= 0) {
-                            updated[existingIdx] = newNetwork
+                    for (const incomingNetwork of incomingNetworks) {
+                        const existingIndex = updated.findIndex(
```
swensel, line 70: Trying to make this a little clearer with updated variable names and comments.

packages/ui-common/state/TemporaryNetworks.ts
```diff
@@ -87,28 +63,32 @@ export const useTempNetworksStore = create<TempNetworksStore>()(
...
-            updateTempNetworkDefinition: (networkName: string, definition: AgentNetworkDefinitionEntry[]) =>
-                set((state) => ({
-                    tempNetworks: state.tempNetworks.map((n) =>
-                        n.agentInfo.agent_name === networkName ? {...n, agentNetworkDefinition: definition} : n
-                    ),
-                })),
+            updateTempNetworkDefinition: (networkName: string, agentNetworkDefinition: AgentNetworkDefinitionEntry[]) =>
+                set((state) => {
+                    const updated = [...state.tempNetworks]
+                    const existingIndex = updated.findIndex((network) => network.agentInfo.agent_name === networkName)
```
swensel, line 86: Matches pattern above, although the logic on the left is more succinct. I think the right is clearer, although more verbose.

jest.config.ts
```diff
@@ -72,10 +72,10 @@ const config: Config.InitialOptions = {
 
     coverageThreshold: {
         global: {
-            statements: -102,
-            branches: -160,
-            functions: -26,
-            lines: -79,
+            statements: -128,
```
swensel: Little bit of a hit here, although there is Atone for lapses in test coverage (Sean - #287 ).

packages/ui-common/state/TemporaryNetworks.ts
```diff
@@ -43,42 +43,18 @@ interface TempNetworksStore {
...
- * Derives the canonical network name from a reservation ID.
- *
- * The backend encodes the network name as a prefix in the reservation ID, followed by a UUID suffix:
- * `{network_name}-{uuid}`, e.g. `travel_agency_ops-7876642e-fe75-4d44-a61e-300688a1a6c5`.
- *
- * Stripping the UUID suffix gives the stable name that can be used for deduplication across reservations.
- * Returns `undefined` when the reservation ID doesn't match the expected format.
- */
-export const extractNetworkNameFromReservationId = (reservationId: string): string | undefined => {
-    const stripped = reservationId.replace(UUID_SUFFIX_RE, "")
```
swensel, line 68: This is just `agentNetworkName`. Trying to reduce unnecessary code.

packages/ui-common/controller/agent/Agent.ts
```diff
@@ -284,3 +290,37 @@ export const getAgentFunction = async (url: string, agent: string, userId: strin
 
     return response.json()
 }
+
+/**
+ * Streams the Agent Network Designer endpoint with an updated agent definition.
+ * Calls `onChunk` for each line received; callers are responsible for parsing
+ * reservations / networks out of each chunk.
+ */
+export const sendNetworkDesignerUpdate = async (
```
swensel, line 299: Moved from `AgentFlow.tsx‎`.

packages/ui-common/components/MultiAgentAccelerator/TemporaryNetworks.ts
```diff
@@ -59,32 +70,73 @@ export const extractNetworkHocon = (message: ChatMessage): string | null => {
...
+ * Returns true when agentName refers to a known temporary network.
+ */
+export const isTemporaryNetwork = (agentName: string | null, networks: TemporaryNetwork[]): boolean =>
+    agentName !== null && networks.some((n) => n.agentInfo.agent_name === agentName)
+
+/**
+ * Merges incoming networks into target, keeping the entry with the highest expiration time.
+ * Returns a new array; does not mutate either argument.
+ */
+export const mergeNetworks = (target: TemporaryNetwork[], incoming: TemporaryNetwork[]): TemporaryNetwork[] =>
```
swensel, line 129: Moved from `AgentFlow.tsx‎` (and refactored).

packages/ui-common/components/MultiAgentAccelerator/TemporaryNetworks.ts
```diff
@@ -26,20 +29,28 @@ export type AgentReservation = {
...
  */
-export const extractReservations = (message: ChatMessage): AgentReservation[] => {
-    // Check for temp networks ("reservations") in sly_data
-    if (message?.type === ChatMessageType.AGENT_FRAMEWORK && message?.sly_data?.[AGENT_RESERVATIONS_KEY]) {
-        return message.sly_data[AGENT_RESERVATIONS_KEY] as AgentReservation[]
-    } else {
-        // Not the type of message that would contain reservations, or no reservations found, return empty array
-        return []
-    }
+export const convertReservationsToNetworks = (
```
swensel, line 39: Moved this up to respect functions in alphabetical order.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -507,16 +590,14 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
...
                             agentIconSuggestions={agentIconSuggestions}
-                            currentUser={userInfo.userName}
                             id="multi-agent-accelerator-agent-flow"
                             key="multi-agent-accelerator-agent-flow"
                             currentConversations={currentConversations}
                             isAwaitingLlm={isAwaitingLlm}
                             isStreaming={isStreaming}
                             isSelectedNetworkTemporary={isSelectedNetworkTemporary}
                             networkId={isSelectedNetworkTemporary ? selectedNetwork : undefined}
-                            neuroSanURL={neuroSanURL}
```
swensel, line 528: These are no longer needed. Just pass in an `onSaveAgent`.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -471,16 +564,6 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
         )
     }
 
-    const onNetworkReplaced = useCallback(
```
swensel, line 485: No longer needed.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -390,6 +426,63 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
         [isNetworkDesignerMode]
     )
 
+    /**
+     * Handles a save from the AgentFlow popup: streams the updated definition to the network designer,
+     * collects reservations from each chunk, then upserts the result and navigates if the network changed.
+     */
+    const onSaveAgent = useCallback(
```
swensel, line 442: Ugh, this is still a little long, I may want to break this up more.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -172,24 +212,37 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
...
     // For temp networks, agent_network_definition and agent_network_name live in localStorage (not IndexedDB slyData).
     // Supply them as extraSlyData so ChatCommon bounces them back to the backend on every request.
     const currentTempNetwork = isSelectedNetworkTemporary
         ? temporaryNetworks.find((n) => n.agentInfo.agent_name === selectedNetwork)
         : undefined
+
+    // For Agent Network Designer: the backend echoes agent_network_name into IndexedDB sly_data as the conversation
+    // progresses. We read it back here to find the matching temp network and override agent_network_definition in the
+    // outgoing request — leaving what's in IndexedDB untouched.
+    const designerSlyData = useAgentChatHistoryStore((state) => state.history[AGENT_NETWORK_DESIGNER_ID]?.slyData)
```
swensel, line 231: This fixes the bug where AND overwrites temp network edits.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -172,24 +212,37 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
     const isNetworkDesignerMode = selectedNetwork === AGENT_NETWORK_DESIGNER_ID
 
     // Whether the currently selected network is a temporary network (agent reservation)
-    const isSelectedNetworkTemporary =
-        selectedNetwork !== null && temporaryNetworks.some((n) => n.agentInfo.agent_name === selectedNetwork)
+    const isSelectedNetworkTemporary = isTemporaryNetwork(selectedNetwork, temporaryNetworks)
```
swensel, line 220: Use `isTemporaryNetwork` util function instead.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -73,6 +75,44 @@ const GROW_ANIMATION_TIME_MS = 800
...
+        if (converted.length === 0) return accumulated
+        return mergeNetworks(accumulated, converted)
+    } catch (e: unknown) {
+        console.warn("Failed to process chunk from network designer:", e)
+        return accumulated
+    }
+}
+
+/** Logs and notifies about a save error. Suppresses AbortError (user-cancelled). */
+const notifySaveError = (agentName: string, e: unknown): void => {
```
swensel, line 104: Used below in a try/catch in `onSaveAgent`.. Kind of ugly. Was vibed. I may revise it.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -73,6 +75,44 @@ const GROW_ANIMATION_TIME_MS = 800
 // Optimization to avoid creating a new empty map on every render
 const EMPTY_THOUGHT_BUBBLE_EDGES = new Map<string, {edge: ThoughtBubbleEdgeShape; timestamp: number}>()
 
+// #region: Agent-save helpers
+
+/**
+ * Extracts TemporaryNetworks from a single streamed chunk, merging into `accumulated`.
+ * Returns `accumulated` unchanged if the chunk yields no reservations or on parse error.
+ */
+const collectNetworksFromChunk = (
```
swensel, line 84: Used below in `onSaveAgent`.

packages/ui-common/components/MultiAgentAccelerator/const.ts
```diff
@@ -58,14 +58,9 @@ export type AgentNetworkDefinitionEntry = ConnectivityInfo & {
...
-export const DISPLAY_AS_LANGCHAIN_TOOL = "langchain_tool"
-export const DISPLAY_AS_EXTERNAL_AGENT = "external_agent"
-
-/**
- * Returns true when an agent node supports the edit popup (instructions + description).
- * Only `llm_agent` nodes (and the Frontman, whose `display_as` is undefined) are editable.
- * `coded_tool`, `langchain_tool`, `external_agent`, and any other types are read-only.
- */
-export const isEditableAgent = (displayAs: string | undefined): boolean => displayAs === DISPLAY_AS_LLM_AGENT
+export enum DisplayAs {
```
swensel, line 61: Switch to `enum` as per PR feedback from https://github.com/cognizant-ai-lab/neuro-san-ui/pull/274.

packages/ui-common/components/MultiAgentAccelerator/AgentNodePopup.tsx
```diff
@@ -214,8 +202,9 @@ export const AgentNodePopup: FC<AgentNodePopupProps> = ({
                     }}
                     placeholder="Enter instructions for this agent…"
                     rows={6}
+                    slotProps={{htmlInput: {style: {fontSize: "0.85rem"}}}}
```
swensel, line 205: Switch to the proper way to do this as per PR feedback from https://github.com/cognizant-ai-lab/neuro-san-ui/pull/274.

packages/ui-common/components/MultiAgentAccelerator/AgentNodePopup.tsx
```diff
@@ -196,8 +184,8 @@ export const AgentNodePopup: FC<AgentNodePopupProps> = ({
                     }}
                     placeholder="Enter a short description of this agent…"
                     rows={6}
+                    slotProps={{htmlInput: {style: {fontSize: "0.85rem"}}}}
```
swensel, line 187: Switch to the proper way to do this as per PR feedback from https://github.com/cognizant-ai-lab/neuro-san-ui/pull/274.

packages/ui-common/components/MultiAgentAccelerator/AgentNodePopup.tsx
```diff
@@ -139,7 +135,7 @@ export const AgentNodePopup: FC<AgentNodePopupProps> = ({
                         ) : undefined
                     }
                 >
-                    {isSaving ? "Applying changes…" : "Save"}
+                    {isSaving ? "Applying changes..." : "Save"}
```
swensel, line 138: Use normal `.`.

packages/ui-common/components/MultiAgentAccelerator/AgentNode.tsx
```diff
@@ -231,7 +215,7 @@ export const AgentNode: FC<NodeProps<RFNode<AgentNodeProps>>> = (props: NodeProp
                 sx={{
                     backgroundColor,
                     color,
-                    cursor: isClickableNode ? "pointer" : "grab",
+                    cursor: isEditable ? "pointer" : "grab",
```
swensel, line 218: Use the `isEditable` prop that is now passed in.

packages/ui-common/components/MultiAgentAccelerator/AgentNode.tsx
```diff
@@ -49,7 +43,7 @@ export interface AgentNodeProps extends Record<string, unknown> {
     readonly getConversations: () => AgentConversation[] | null
     readonly isAwaitingLlm?: boolean
     readonly agentIconSuggestion?: string
-    readonly isTemporaryNetwork?: boolean
+    readonly isEditable?: boolean
```
swensel, line 46: `isEditable` prop that is now passed in.

packages/ui-common/components/MultiAgentAccelerator/AgentNode.tsx
```diff
@@ -29,13 +29,7 @@ import type {Node as RFNode} from "@xyflow/react"
...
 
 import {AgentConversation} from "./AgentConversations"
-import {
-    DISPLAY_AS_CODED_TOOL,
-    DISPLAY_AS_EXTERNAL_AGENT,
-    DISPLAY_AS_LANGCHAIN_TOOL,
-    DISPLAY_AS_LLM_AGENT,
-    isEditableAgent,
-} from "./const"
+import {DisplayAs} from "./const"
```
swensel, line 32: Switched to an `enum`.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -403,153 +348,83 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
+            } finally {
+                cancelTimeout()
+                saveAbortControllerRef.current = null
             }
         },
-        [networkId, onNetworkReplaced]
+        [onSaveAgent]
     )
 
     const handlePopupSave = useCallback(
```
swensel, line 410: Broke up `handlePopupSave` so that it wasn't as long.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -403,153 +348,83 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
-                agentId: node.id,
-                agentName: node.data.agentName,
-                initialInstructions: found?.instructions ?? "",
-                initialDescription: found?.description ?? "",
-            })
-            setIsPopupOpen(true)
-        },
-        [tempNetworks, isTemporaryNetwork, networkId]
-    )
+    const handleNodeClick: NodeMouseHandler<RFNode<AgentNodeProps>> = useCallback((_event, node) => {
```
swensel, line 364: Shorten `handleNodeClick` and also checked the `node.data.isEditable` prop.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -115,72 +105,29 @@ const THOUGHT_BUBBLE_TIMEOUT_MS = 10_000
...
-    }
-}
-
 /**
- * Streams the Agent Network Designer endpoint with the updated definition and collects
- * the resulting reservations. Returns the deduplicated list of new networks.
+ * Returns a new definition array with the target agent's instructions and description updated.
+ * All other entries pass through unchanged.
  */
-const streamNetworkDesignerUpdate = async (
```
swensel, line 137: Moved to `packages/ui-common/components/MultiAgentAccelerator/TemporaryNetworks.ts`.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -115,72 +105,29 @@ const THOUGHT_BUBBLE_TIMEOUT_MS = 10_000
 
 // #region: Helpers
 
-/** Merges incoming networks into target, keeping the entry with the highest expiration time. */
-const mergeNetworks = (target: NetworkList, incoming: NetworkList): void => {
```
swensel, line 119: Moved to `packages/ui-common/components/MultiAgentAccelerator/TemporaryNetworks.ts`.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -102,8 +94,6 @@ export interface AgentFlowProps {
 
 type Layout = "radial" | "linear"
 
-type NetworkList = ReturnType<typeof convertReservationsToNetworks>
```
swensel, line 105: Type not needed, it was just `TemporaryNetwork[]`.

packages/ui-common/__tests__/unit/State/TemporaryNetworks.test.ts
```diff
@@ -162,18 +107,20 @@ describe("useTempNetworksStore / upsertTempNetworks", () => {
     })
 
     it("upserts multiple networks in a single call, including mixed replacements and additions", () => {
-        const existing = makeNetwork("res-old", "shared_name")
+        const existing = makeNetwork("shared_name-aaaa0001-aaaa-aaaa-aaaa-aaaaaaaaaaaa", "shared_name")
```
swensel, line 110: Bit more precise but not that important.

packages/ui-common/__tests__/unit/State/TemporaryNetworks.test.ts
```diff
@@ -52,94 +52,39 @@ describe("useTempNetworksStore / upsertTempNetworks", () => {
...
+            "temporary/network_alpha-aaaa0001-aaaa-aaaa-aaaa-aaaaaaaaaaaa"
         )
         // Beta should be replaced
         expect(stored.find((n) => n.agentNetworkName === "network_beta")?.agentInfo.agent_name).toBe(
-            "temporary/res-beta-v2"
+            "temporary/network_beta-bbbb0002-bbbb-bbbb-bbbb-bbbbbbbbbbbb"
         )
     })
 
-    it("replaces an old network that has no agentNetworkName by deriving the name from its reservation_id", () => {
```
swensel, line 88: Removed useless "old network" tests.

packages/ui-common/__tests__/unit/State/TemporaryNetworks.test.ts
```diff
@@ -52,94 +52,39 @@ describe("useTempNetworksStore / upsertTempNetworks", () => {
     })
 
     it("replaces an existing network when a new one has the same agentNetworkName", () => {
-        const original = makeNetwork("res-1", "network_alpha")
+        const original = makeNetwork("network_alpha-11111111-1111-1111-1111-111111111111", "network_alpha")
```
swensel, line 55: Bit more precise checks but not that important.

packages/ui-common/__tests__/unit/Controller/Agent.test.ts
```diff
@@ -275,3 +277,61 @@ describe("Controller/Agent/getAgentFunction", () => {
         )
     })
 })
+
+describe("Controller/Agent/sendNetworkDesignerUpdate", () => {
```
swensel, line 281: Tests for new `sendNetworkDesignerUpdate` function.

packages/ui-common/__tests__/components/MultiAgentAccelerator/TemporaryNetworks.test.ts
```diff
@@ -87,38 +96,197 @@ describe("convertReservationsToNetworks", () => {
...
+            [AGENT_NETWORK_DEFINITION_KEY]: slyDefinition,
+        })
+
+        const [network] = extractTemporaryNetworksFromMessage(message, overrideDefinition)
+        expect(network.agentNetworkDefinition).toBe(overrideDefinition)
     })
 })
 
-describe("extractNetworkNameFromReservationId", () => {
+describe("mergeNetworks", () => {
```
swensel, line 225: Tests for new `mergeNetworks` function.

packages/ui-common/__tests__/components/MultiAgentAccelerator/TemporaryNetworks.test.ts
```diff
@@ -87,38 +96,197 @@ describe("convertReservationsToNetworks", () => {
...
+    it("returns false for external_agent", () => {
+        expect(isEditableAgent("external_agent")).toBe(false)
+    })
+
+    it("returns false when displayAs is undefined", () => {
+        expect(isEditableAgent(undefined)).toBe(false)
+    })
+})
+
+describe("extractTemporaryNetworksFromMessage", () => {
```
swensel, line 151: Tests for new `extractTemporaryNetworksFromMessage` function.

packages/ui-common/__tests__/components/MultiAgentAccelerator/TemporaryNetworks.test.ts
```diff
@@ -87,38 +96,197 @@ describe("convertReservationsToNetworks", () => {
...
+    it("returns false when no network matches the agentName", () => {
+        expect(isTemporaryNetwork("temporary/other", [makeNetwork("temporary/abc")])).toBe(false)
+    })
+
+    it("returns true when a network matches the agentName", () => {
+        expect(isTemporaryNetwork("temporary/abc", [makeNetwork("temporary/abc")])).toBe(true)
+    })
+})
+
+describe("isEditableAgent", () => {
```
swensel, line 127: Tests for new `isEditableAgent` function.

packages/ui-common/__tests__/components/MultiAgentAccelerator/TemporaryNetworks.test.ts
```diff
@@ -87,38 +96,197 @@ describe("convertReservationsToNetworks", () => {
 
         expect(network.agentNetworkName).toBe("travel_agency_ops")
     })
+})
 
-    it("prefers the explicit agentNetworkName over the derived one", () => {
-        const res = makeReservation("travel_agency_ops-7876642e-fe75-4d44-a61e-300688a1a6c5")
-        const [network] = convertReservationsToNetworks([res], null, undefined, "override_name")
+describe("isTemporaryNetwork", () => {
```
swensel, line 101: Tests for new `isTemporaryNetwork` function.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
```diff
@@ -1963,17 +1656,15 @@ describe("AgentFlow", () => {
...
                 expect(screen.queryByRole("button", {name: /applying changes/iu})).not.toBeInTheDocument()
             })
 
-            expect(consoleErrorSpy).toHaveBeenCalledWith(expect.stringContaining("Failed to submit"), expect.any(Error))
             consoleErrorSpy.mockRestore()
-            consoleDebugSpy.mockRestore()
         })
 
-        it("closes popup immediately (no API call) when neuroSanURL is not provided", async () => {
+        it("closes popup immediately without calling onSaveAgent when it is not provided", async () => {
```
swensel, line 1667: I may look into this more. I suppose it's fine the popup closes in this case though.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
```diff
@@ -1963,17 +1656,15 @@ describe("AgentFlow", () => {
             await user.type(instructionsField, "Updated instructions")
             await user.click(screen.getByRole("button", {name: "Save"}))
 
-            // Popup should close even on error (finally block)
+            // Popup closes immediately on save; onSaveAgent error is handled silently in the background
```
swensel, line 1659: I'll also look into this more.

apps/main/__tests__/pages/MultiAgentAcceleratorPage.test.tsx
```diff
@@ -1034,4 +1036,90 @@ describe("Multi Agent Accelerator Page", () => {
             expect(lastCall[0]).toBeNull()
         })
     })
+
+    describe("extraSlyData", () => {
```
swensel, line 1120: Tests for `extraSlyData`. Added as part of the fix for AND overwriting temp network "manual edits" when doing topological edits on AND.

packages/ui-common/state/TemporaryNetworks.ts
```diff
@@ -87,28 +63,32 @@ export const useTempNetworksStore = create<TempNetworksStore>()(
...
                     const updated = [...state.tempNetworks]
-                    for (const newNetwork of newNetworks) {
-                        const newName = effectiveNetworkName(newNetwork)
-                        const existingIdx = newName ? updated.findIndex((n) => effectiveNetworkName(n) === newName) : -1
-                        if (existingIdx >= 0) {
-                            updated[existingIdx] = newNetwork
+                    for (const incomingNetwork of incomingNetworks) {
+                        const existingIndex = updated.findIndex(
+                            (network) => network.agentNetworkName === incomingNetwork.agentNetworkName
+                        )
```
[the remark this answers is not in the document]
swensel: `agentNetworkName` has always been required since `temp-networks` was added. I made some changes in this PR to make that more clear and not worry about "old networks". MVP v1 (#274) was always like this, it was maybe only an issue when I was doing development prior to v1.

packages/ui-common/components/MultiAgentAccelerator/TemporaryNetworks.ts
```diff
@@ -26,20 +29,28 @@ export type AgentReservation = {
...
+    networkHocon: string | null,
+    agentNetworkDefinition?: AgentNetworkDefinitionEntry[]
+): TemporaryNetwork[] => {
+    return agentReservations.map((reservation) => ({
+        reservation,
+        agentInfo: {
+            agent_name: `${TEMPORARY_NETWORK_FOLDER}/${reservation.reservation_id}`,
+        },
+        // agentNetworkName is the network name "prefix", with the UUID-stripped
+        agentNetworkName: removeTrailingUuid(reservation.reservation_id),
```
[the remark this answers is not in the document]
swensel: `agentNetworkName` and `removeTrailingUuid(reservation.reservation_id)` are the same. We may switch to `reservation_id` later, but not for now.

__tests__/common/NetworksListMock.ts
```diff
@@ -69,5 +69,6 @@ export const TEMPORARY_NETWORK: TemporaryNetwork = {
         expiration_time_in_seconds: Math.floor(Date.now() / 1000) + 3600,
     },
     agentInfo: TEMP_NETWORK_AGENT_INFO,
+    agentNetworkName: TEMPORARY_NETWORK_NAME,
```
swensel, line 72: There was some logic that assumed `agentNetworkName` was optional. It shouldn't be optional. Making sure that I'm passing it for tests now, too.

apps/main/__tests__/pages/MultiAgentAcceleratorPage.test.tsx
```diff
@@ -554,6 +555,7 @@ describe("Multi Agent Accelerator Page", () => {
                 agentInfo: expect.objectContaining({
                     agent_name: agentName,
                 }),
+                agentNetworkName: RESERVATION.reservation_id,
```
swensel, line 581: There was some logic that assumed `agentNetworkName` was optional. It shouldn't be optional. Making sure that I'm passing it for tests now, too.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
```diff
@@ -1597,202 +1564,36 @@ describe("AgentFlow", () => {
...
-                expect(screen.queryByRole("button", {name: /applying changes/iu})).not.toBeInTheDocument()
-            })
-
-            const stored = useTempNetworksStore.getState().tempNetworks
-            // Old network should be gone, new one should be present
-            expect(stored.some((n) => n.agentInfo.agent_name === OLD_NETWORK_ID)).toBe(false)
-            expect(stored.some((n) => n.agentInfo.agent_name === NEW_NETWORK_ID)).toBe(true)
-        })
-
-        it("keeps the newest reservation when the backend streams an old echo then the new one", async () => {
```
swensel, line 1666: Still about merging networks. Moved to `packages/ui-common/components/MultiAgentAccelerator/TemporaryNetworks.ts`.

packages/ui-common/controller/agent/Agent.ts
swensel: This is the "manual edit" case from temp networks.

packages/ui-common/components/MultiAgentAccelerator/AgentNodePopup.tsx
dsargent: I was wondering about that. Seemed strange as we humans don't normally reach for the UNICODE ellipsis character, but nothing inherently bad about it?

__tests__/common/NetworksListMock.ts
dsargent: It's defined as mandatory in the `TemporaryNetwork` interface?

apps/main/__tests__/pages/MultiAgentAcceleratorPage.test.tsx
dsargent: I don't get it? If you don't include it here `tsc` will complain due to typing:

`const expectedTemporaryNetwork: TemporaryNetwork`

apps/main/__tests__/pages/MultiAgentAcceleratorPage.test.tsx
dsargent: Good that you put them in a separate `describe` suite.

packages/ui-common/components/MultiAgentAccelerator/AgentNodePopup.tsx
dsargent: The trouble is now we have `isAwaitingLlm`, `isStreaming`, `isSaving` and it's getting confusing and hard to manage.

packages/ui-common/components/MultiAgentAccelerator/const.ts
dsargent: Small nit, but please start sorting items like these, properties etc. alphabetically. Eventually I want to enable linter rules to enforce, so might as well get started preparing.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
dsargent: Elsewhere we use "extract" though. Should be consistent.

dsargent: Also, this should be in the "controller" hierarchy somewhere, not in the presentation layer (this module).

packages/ui-common/components/MultiAgentAccelerator/GraphLayouts.ts
```diff
@@ -327,7 +330,8 @@ export const layoutLinear = (
                 isAwaitingLlm,
                 depth: undefined, // Depth will be computed later,
                 agentIconSuggestion: agentIconSuggestions?.[nodeId],
-                isTemporaryNetwork,
+                isEditable:
```
dsargent, line 333: Style call, but in situations like this I prefer "introduce explanatory variable" refactorings:

`const isEditable = ...`

Then you can also add a comment for the variable, or often it's "obvious" and doesn't even need one. "An agent is editable if it belongs to a temporary network, and is of a particular type."

Not to mention that the logic is copy-pasted from the previous function that tests for the same thing.

packages/ui-common/state/TemporaryNetworks.ts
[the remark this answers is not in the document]
dsargent: It's keying off your comments in the PR saying that previously it was optional (which it never was as you point out here). No wonder the AI is confused.

packages/ui-common/components/MultiAgentAccelerator/TemporaryNetworks.ts
dsargent: Could probably replace this with lodash [mergeWith](https://lodash.info/doc/mergeWith) using the `customizer` for your "latest one wins" logic.

packages/ui-common/components/MultiAgentAccelerator/AgentNode.tsx
dsargent: This is the thing that doesn't work for me. could be a Firefox or Linux issue, ymmv. But in general, I think "custom cursors" don't place nicely with react-flow because _it_ wants to control the cursor for cueing the user to node moves & edits.

packages/ui-common/components/MultiAgentAccelerator/AgentNode.tsx
```diff
@@ -182,23 +168,23 @@ export const AgentNode: FC<NodeProps<RFNode<AgentNodeProps>>> = (props: NodeProp
                 )
             } else {
                 switch (displayAs) {
-                    case DISPLAY_AS_EXTERNAL_AGENT:
+                    case DisplayAs.EXTERNAL_AGENT:
```
dsargent, line 171: Better as an enum 👍🏻

packages/ui-common/components/MultiAgentAccelerator/AgentNode.tsx
dsargent: I prefer it this way. Nodes shouldn't know about things like temporary networks. All they need to know is "you are editable; don't ask why."

packages/ui-common/components/MultiAgentAccelerator/AgentNodePopup.tsx
```diff
@@ -168,21 +164,13 @@ export const AgentNodePopup: FC<AgentNodePopupProps> = ({
...
             <MUIDialog
-                dialogSx={isSaving ? {"& *, & *::before, & *::after": {cursor: "wait !important"}} : undefined}
                 footer={footer}
                 id="agent-node-popup"
                 isOpen={isOpen}
                 onClose={handleClose}
                 paperProps={{minWidth: "480px", maxWidth: "600px", width: "100%"}}
                 title={agentName}
             >
-                {/* Progress bar shown while saving — makes it clear the dialog is busy */}
```
dsargent, line 179: I agree with removing this. We still have plenty of visual cues to inform the user that "something is happening, please wait".

packages/ui-common/components/MultiAgentAccelerator/TemporaryNetworks.ts
```diff
@@ -59,32 +70,73 @@ export const extractNetworkHocon = (message: ChatMessage): string | null => {
...
+
+/**
+ * Extracts TemporaryNetwork objects from a chat message by reading reservations, HOCON,
+ * agent network definition, and agent network name from the message / sly_data.
+ * @param message The chat message to extract from.
+ * @param agentNetworkDefinitionOverride Optional definition override (e.g. a locally-edited version) that
+ *   takes precedence over whatever the backend echoes in sly_data.
+ * @returns A (possibly empty) array of TemporaryNetwork objects ready for the store.
+ */
+export const extractTemporaryNetworksFromMessage = (
```
dsargent, line 97: No need to say "from" message (repeats what is obvious from the signature)

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -115,72 +105,29 @@ const THOUGHT_BUBBLE_TIMEOUT_MS = 10_000
...
+const buildUpdatedDefinition = (
+    currentDefinitions: AgentNetworkDefinitionEntry[],
+    agentId: string,
+    instructions: string,
+    description: string
+): AgentNetworkDefinitionEntry[] =>
+    currentDefinitions.map((entry) => (entry.origin === agentId ? {...entry, instructions, description} : entry))
+
+/** Finds the temp network entry for the given network ID. */
+const findTempNetwork = <T extends {agentInfo: {agent_name?: string}}>(
```
dsargent, line 121: Doesn't belong in here

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -115,72 +105,29 @@ const THOUGHT_BUBBLE_TIMEOUT_MS = 10_000
...
+    tempNetworks: T[],
+    networkId: string | undefined
+): T | undefined => (networkId ? tempNetworks.find((n) => n.agentInfo.agent_name === networkId) : undefined)
+
+/** Creates an AbortController with a built-in timeout that fires a TimeoutError. */
+const createTimedAbortController = (timeoutMs: number): {controller: AbortController; cancelTimeout: () => void} => {
```
dsargent, line 127: Doesn't belong in here either, probably in the Controller layer.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -403,153 +348,84 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
+            agentName: node.data.agentName,
+        })
+        setIsPopupOpen(true)
+    }, [])
 
     const handlePopupClose = useCallback(() => {
-        // If a save is in-flight, abort it immediately so the stream doesn't hang.
-        saveAbortControllerRef.current?.abort()
+        // This was added for API instability, however handlePopupClose is not called if the save is in progress
+        saveAbortControllerRef.current?.abort() // Could remove
```
dsargent, line 376: "could remove" is sort of a teaser. Yoda principle applies..."remove, or remove not...there is no 'could'."

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -403,153 +348,84 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
+            if (!onSaveAgent || updated.length === 0) return
+
+            const {controller, cancelTimeout} = createTimedAbortController(60_000)
+            saveAbortControllerRef.current = controller
+            try {
+                await onSaveAgent(agentName, updated, currentAgentNetworkName, controller.signal)
+            } catch (e) {
+                // onSaveAgent is responsible for user-facing error handling and notification.
+                // We catch here only to ensure the finally block (and popup close) always runs.
+                console.error(`Error saving agent ${agentName}. See onSaveAgent implementation for details.`, e)
```
dsargent, line 401: `catch` is optional btw. You can still have `try...finally` without `catch`.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -403,153 +348,84 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
-                sendNotification(
-                    NotificationType.error,
-                    `Failed to update agent "${agentName}".`,
-                    "A reservation was returned but did not match the current network. Please try again."
-                )
+    /**
+     * Manages the AbortController, timeout, and cleanup for a single agent-save attempt.
+     * Delegates the actual streaming and result application to the `onSaveAgent` prop.
+     */
+    const executeAgentSave = useCallback(
```
dsargent, line 386: Belongs in controller layer probably

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
dsargent: Where did all the deleted code go? Was it not necessary? (Not complaining. Deleted code is my favorite kind of code. Just curious.)

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -172,24 +212,37 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
...
     // Supply them as extraSlyData so ChatCommon bounces them back to the backend on every request.
     const currentTempNetwork = isSelectedNetworkTemporary
         ? temporaryNetworks.find((n) => n.agentInfo.agent_name === selectedNetwork)
         : undefined
+
+    // For Agent Network Designer: the backend echoes agent_network_name into IndexedDB sly_data as the conversation
+    // progresses. We read it back here to find the matching temp network and override agent_network_definition in the
+    // outgoing request — leaving what's in IndexedDB untouched.
+    const designerSlyData = useAgentChatHistoryStore((state) => state.history[AGENT_NETWORK_DESIGNER_ID]?.slyData)
+    const designerNetworkName = isNetworkDesignerMode
```
dsargent, line 232: You can probably simplify this since you want `undefined`/something "falsy" anyway.

```typescript
const designerNetworkName = isNetworkDesignerMode && designerSlyData?.[AGENT_NETWORK_NAME_KEY]
```

Will return `undefined` for the various "missing" cases. To be verified (ask Copilot).

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -172,24 +212,37 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
...
+        : undefined
+    const designerTempNetwork = designerNetworkName
+        ? temporaryNetworks.find((n) => n.agentNetworkName === designerNetworkName)
+        : undefined
+
     const extraSlyData: Record<string, unknown> | undefined = currentTempNetwork
         ? {
               [AGENT_NETWORK_DEFINITION_KEY]: currentTempNetwork.agentNetworkDefinition,
-              // Use the name the backend originally sent, not the local UUID-based key.
+              // Use the agentNetworkName, not reservation_id
```
dsargent, line 242: I really think we need to move to `reservation_id` identifiers for temp networks. Use the friendly name is going to bite us and confuse users. Please create a ticket if you haven't already.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -390,6 +426,63 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
...
+                let newNetworks: TemporaryNetwork[] = []
+                await sendNetworkDesignerUpdate(
+                    neuroSanURL,
+                    signal,
+                    agentName,
+                    updated,
+                    agentNetworkName,
+                    userInfo.userName,
+                    (chunk) => {
+                        newNetworks = collectNetworksFromChunk(chunk, updated, newNetworks)
```
dsargent, line 459: Not keen on the side effect here, assigning to the mutable `newNetworks`. Very impure and un-React-like.

Also, each chunk will overwrite `newNetworks` if there are multiple chunks with networks. Is that what you want here, "last one wins"? It may work now, but what about in future if we are dealing with multiple in a single interaction?

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -390,6 +426,63 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
...
+                    }
+                )
+
+                if (newNetworks.length === 0) {
+                    sendNotification(
```
dsargent, line 464: UX thought: starting to think we should show errors via inline accordions in the chat, rather than annoying popup toasts. Not 100% sure. We do that in some cases already. Thoughts?

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -390,6 +426,63 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
...
+                        useAgentChatHistoryStore
+                            .getState()
+                            .copyHistory(selectedNetwork, replacement.agentInfo.agent_name)
+                        setSelectedNetwork(replacement.agentInfo.agent_name)
+                    }
+                } else {
+                    sendNotification(
+                        NotificationType.error,
+                        `Failed to update agent "${agentName}".`,
+                        "A reservation was returned but did not match the current network. Please try again."
```
dsargent, line 485: Users will have no idea what this means. "Reservation?! Didn't match the current network?!"

Please take a step back and think how to present this from the user's point of view, and perhaps tell them what they can do about it (if anything). If you really want to log the grungy technical details, that can be `console.error()` in addition to the toast.

packages/ui-common/components/MultiAgentAccelerator/TemporaryNetworks.ts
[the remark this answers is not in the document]
dsargent: s/may/absolutely should/

packages/ui-common/components/MultiAgentAccelerator/TemporaryNetworks.ts
```diff
@@ -59,32 +70,73 @@ export const extractNetworkHocon = (message: ChatMessage): string | null => {
...
+ * type AGENT_FRAMEWORK only.
+ * @return An array of AgentReservation objects if reservations are found, or an empty array if not found or
+ * if the message is not of the expected type.
  */
-export const convertReservationsToNetworks = (
-    agentReservations: AgentReservation[],
-    networkHocon: string | null,
-    agentNetworkDefinition?: AgentNetworkDefinitionEntry[],
-    agentNetworkName?: string
+const extractReservations = (message: ChatMessage): AgentReservation[] => {
```
dsargent, line 79: Couldn't tell what's going on here -- no annotations and the diffs are noisy.

packages/ui-common/components/MultiAgentAccelerator/TemporaryNetworks.ts
```diff
@@ -59,32 +70,73 @@ export const extractNetworkHocon = (message: ChatMessage): string | null => {
...
-        // networks are always deduplicated by name even when the backend omits AGENT_NETWORK_NAME_KEY.
-        agentNetworkName: agentNetworkName ?? extractNetworkNameFromReservationId(reservation.reservation_id),
-        networkHocon,
-        agentNetworkDefinition,
-    }))
+    const reservations = extractReservations(message)
+    if (reservations.length === 0) return []
+
+    const networkHocon = extractNetworkHocon(message)
+    const agentNetworkDefinition =
```
dsargent, line 105: Again, the overcautious checking and manually returning `undefined` is unnecessary here. Smacks of LLM paranoia and verbosity?

When you ask for `foo.bar` and `foo` doesn't have `bar`, you get `undefined` anyway. That's what `undefined` is for. You don't need to manually fall back to it.

```typescript
> const a = {answer: 42}
>a.b
undefined 
```

I feel like we need a whole separate set of "annoying stuff that AI coding agents generate" linter rules for things like this.

packages/ui-common/components/MultiAgentAccelerator/TemporaryNetworks.ts
```diff
@@ -59,32 +70,73 @@ export const extractNetworkHocon = (message: ChatMessage): string | null => {
...
+ */
+export const isTemporaryNetwork = (agentName: string | null, networks: TemporaryNetwork[]): boolean =>
+    agentName !== null && networks.some((n) => n.agentInfo.agent_name === agentName)
+
+/**
+ * Merges incoming networks into target, keeping the entry with the highest expiration time.
+ * Returns a new array; does not mutate either argument.
+ */
+export const mergeNetworks = (target: TemporaryNetwork[], incoming: TemporaryNetwork[]): TemporaryNetwork[] =>
+    incoming.reduce<TemporaryNetwork[]>(
```
dsargent, line 130: LLM generating single-letter variables again because it thinks they're loop indices?

`i` is fine -- that one _is_ a loop index and is standard. But `n`, `e` -- not so much.

packages/ui-common/controller/agent/Agent.ts
```diff
@@ -284,3 +290,37 @@ export const getAgentFunction = async (url: string, agent: string, userId: strin
...
+    agentName: string,
+    updated: AgentNetworkDefinitionEntry[],
+    agentNetworkName: string | undefined,
+    currentUser: string,
+    onChunk: (chunk: string) => void
+): Promise<void> => {
+    await sendChatQuery(
+        url,
+        signal,
+        // Shouldn't have to pass a user message, but API behaves different without it
```
dsargent, line 311: Sounds like a ticket for Neuro-san or against AND?

packages/ui-common/state/TemporaryNetworks.ts
dsargent: Could probably use lodash `merge` here too.

jest.config.ts
```diff
@@ -72,10 +72,10 @@ const config: Config.InitialOptions = {
 
     coverageThreshold: {
         global: {
-            statements: -101,
-            branches: -155,
-            functions: -26,
-            lines: -78,
+            statements: -127,
```
dsargent: As soon as things settle down, we absolutely have to hit those test coverage tickets and get back to the "good times". Coverage is getting lower which adds risk.

packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
dsargent: Yes, always good to make the strictest assertion you can without it being too fragile. It's an art. But if you can "say more", then it's good to do so. The narrower you make the "gaps", the harder for bugs to creep in.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
dsargent: https://github.com/xyflow/xyflow/issues/2587
https://github.com/xyflow/xyflow/issues/2461#issuecomment-1263189140 [solution]

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
```diff
@@ -1597,202 +1564,36 @@ describe("AgentFlow", () => {
...
-                chunkCallback(makeChunk(OLD_NETWORK_ID.replace("temporary/", ""), OLD_EXPIRY))
-            })
-
-            act(() => {
-                useTempNetworksStore
-                    .getState()
-                    .setTempNetworks([
-                        makeTempNetwork(OLD_NETWORK_ID, [{origin: "agent1", tools: []}], OLD_NETWORK_NAME),
-                    ])
+                expect(screen.getByRole("button", {name: /applying changes/iu})).toBeInTheDocument()
```
dsargent, line 1579: Gotta love that pointless /i that the LLMs love to add. It's not like we don't know the actual case of the text.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
```diff
@@ -1597,202 +1564,36 @@ describe("AgentFlow", () => {
...
+                expect(screen.queryByRole("button", {name: /^save$/iu})).not.toBeInTheDocument()
```
dsargent, line 1580: /^save$/ is just "Save" but with an added, unnecessary regex.

I'm going to write those "anti LLM linter rules" myself some day.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
```diff
@@ -1597,202 +1564,36 @@ describe("AgentFlow", () => {
...
         })
 
-        it("new network carries the edited definition when backend does not echo it back", async () => {
-            // Backend returns a reservation but no agent_network_definition in sly_data
-            ;(sendChatQuery as jest.Mock).mockImplementation(async (_url, _signal, _query, _agent, chunkCallback) => {
-                chunkCallback(makeReservationChunk(NEW_RESERVATION_ID, OLD_NETWORK_NAME))
-            })
+        it("calls onSaveAgent with the correct agentName, updated definition, networkName and a signal", async () => {
+            const onSaveAgent = jest.fn().mockResolvedValue(undefined)
+            const EDITED_INSTRUCTIONS = "Updated instructions for agent1"
```
dsargent, line 1595: naming -- SCREAMING_SNAKE_CASE is for module-level constants. Either make it such, or use camel.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
```diff
@@ -1818,131 +1618,25 @@ describe("AgentFlow", () => {
             await user.type(instructionsField, EDITED_INSTRUCTIONS)
             await user.click(screen.getByRole("button", {name: "Save"}))
```
dsargent, line 1619: In this case, no pointless regex for `Save` (as it should be). So why is there one in the previous case?

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
```diff
@@ -1818,131 +1618,25 @@ describe("AgentFlow", () => {
...
-            expect(enqueueSnackbar).toHaveBeenCalledWith(
-                expect.anything(),
-                expect.objectContaining({
-                    variant: "error",
-                    description: expect.stringContaining("did not return a reservation"),
-                })
+            expect(onSaveAgent).toHaveBeenCalledTimes(1)
+            const [calledAgentName, calledUpdated, calledNetworkName, calledSignal] = onSaveAgent.mock.calls[0]
+            // agentName is the display name (cleaned up from the raw ID)
+            expect(typeof calledAgentName).toBe("string")
```
dsargent, line 1628: Can we just assert what we expect `calledAgentName` to be, its value rather than this game of "animal, vegetable, mineral"? Or is it truly unpredictable? Also comment says `agentName` but the variable is called `calledAgentName`.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
dsargent: Probably lack of an `await` somewhere?

packages/ui-common/__tests__/components/MultiAgentAccelerator/TemporaryNetworks.test.ts
```diff
@@ -87,38 +96,197 @@ describe("convertReservationsToNetworks", () => {
...
+
+    it("returns false when the networks array is empty", () => {
+        expect(isTemporaryNetwork("temporary/abc", [])).toBe(false)
+    })
+
+    it("returns false when no network matches the agentName", () => {
+        expect(isTemporaryNetwork("temporary/other", [makeNetwork("temporary/abc")])).toBe(false)
+    })
+
+    it("returns true when a network matches the agentName", () => {
```
dsargent, line 122: Might want to add a test to see how it behaves when differ only in case, `temporary/abc` vs `temporary/Abc`. Whatever you hope would happen, assert it.

dsargent (dismissed):
This was a lot of work. Thanks for doing this.

I still have quite a few comments but no actual blockers. And this is an improvement over what we have, so I'm approving it.

dsargent (approved):

PR #326, opened 2026-05-14
Enable more ESLint rules and fix issues

dsargent:
Enables almost all¹ the very useful and important `react-hooks` rules and fixes the few issues that resulted.
These linter rules come straight from the React Dev team and help us use React properly. Violating them can lead to subtle bugs that will haunt us, so let's have the linter keep us honest and following React best practices.

¹Unfortunately not `react-hooks/set-state-in-effect`. Too many violations and will need to be a separate PR. 

Can't add Copilot here because,

```
You have reached your monthly limit for [premium requests](https://docs.github.com/en/copilot/concepts/copilot-billing/understanding-and-managing-requests-in-copilot) for Copilot code review. Limit resets on June 1, 2026, at 12:00 am.
```

Local IDE Copilot has looked at it and helped with it though.

### DIFF
--- a/packages/dev-common/Configs/eslint.config.mjs
+++ b/packages/dev-common/Configs/eslint.config.mjs
@@ -325,8 +325,6 @@ export default defineConfig([
                 },
             ],
 
-            "react-hooks/exhaustive-deps": "error",
-
             // Rules we're not ready to enable yet
 
             // Requires strict type checks enabled in tsc which we're not ready for yet
@@ -448,11 +446,7 @@ export default defineConfig([
             "one-var": "off",
             "prefer-destructuring": "off",
             radix: "off",
-            "react-hooks/preserve-manual-memoization": "off",
-            "react-hooks/purity": "off",
-            "react-hooks/refs": "off",
             "react-hooks/set-state-in-effect": "off",
-            "react/button-has-type": "off",
             "react/destructuring-assignment": "off",
             "react/forbid-component-props": "off",
             // Causes an error: "context.getSourceCode is not a function" -- ESLint 10 issue?
--- a/packages/ui-common/__tests__/components/AgentChat/VoiceChat/MicrophoneButton.test.tsx
+++ b/packages/ui-common/__tests__/components/AgentChat/VoiceChat/MicrophoneButton.test.tsx
@@ -52,6 +52,7 @@ jest.mock("../../../../components/AgentChat/Common/LlmChatButton", () => ({
             onClick={onClick}
             disabled={disabled}
             data-testid="microphone-button"
+            type="button"
         >
             {children}
         </button>
--- a/packages/ui-common/__tests__/components/ChatBot/ChatBot.test.tsx
+++ b/packages/ui-common/__tests__/components/ChatBot/ChatBot.test.tsx
@@ -22,6 +22,7 @@ jest.mock("../../../components/AgentChat/ChatCommon/ChatCommon", () => ({
             <button
                 onClick={props.onClose}
                 data-testid="close-button"
+                type="button"
             >
                 Close
             </button>
--- a/packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
+++ b/packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
@@ -127,7 +127,8 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
         customURLLocalStorage?.replaceAll('"', "") || backendNeuroSanApiUrl
     )
 
-    const agentCountsRef = useRef<Map<string, number>>(new Map())
+    // Tracks how many times each agent has been involved in the conversation
+    const [agentCounts, setAgentCounts] = useState<Map<string, number>>(new Map())
 
     const conversationsRef = useRef<AgentConversation[] | null>(null)
 
@@ -322,7 +323,7 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
             // If the selected network is one of the expired ones, deselect it
             if (expiredNetwork.some((n) => n.agentInfo.agent_name === selectedNetwork)) {
                 setSelectedNetwork(null)
-                agentCountsRef.current = new Map()
+                setAgentCounts(new Map())
             }
         }, EXPIRED_NETWORKS_CHECK_INTERVAL_MS)
 
@@ -344,8 +345,10 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
                 setCurrentConversations(result)
             }
 
-            // Agent hit counts
-            agentCountsRef.current = getUpdatedAgentCounts(agentCountsRef.current, chatMessage?.origin)
+            // Update agent hit counts
+            setAgentCounts((prevCounts) => {
+                return getUpdatedAgentCounts(prevCounts, chatMessage.origin)
+            })
 
             // Agent network designer progress messages
             if (isNetworkDesignerMode) {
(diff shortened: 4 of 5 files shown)


packages/dev-common/Configs/eslint.config.mjs
```diff
@@ -448,11 +446,7 @@ export default defineConfig([
             "one-var": "off",
             "prefer-destructuring": "off",
             radix: "off",
-            "react-hooks/preserve-manual-memoization": "off",
```
dsargent, line 451: We enable all rules by default, then disable the ones we don't want here. So, deleting them from here = enabling them.

packages/dev-common/Configs/eslint.config.mjs
```diff
@@ -448,11 +446,7 @@ export default defineConfig([
...
-            "react-hooks/purity": "off",
-            "react-hooks/refs": "off",
             "react-hooks/set-state-in-effect": "off",
-            "react/button-has-type": "off",
```
dsargent, line 455: I accidentally enabled this one, realized there were only a couple of violations (in tests, in fact) so I just fixed them. It's a good rule to have -- the default type for "button" is "submit" which causes all kinds of issues unless that's what you intend.

packages/ui-common/__tests__/components/AgentChat/VoiceChat/MicrophoneButton.test.tsx
```diff
@@ -52,6 +52,7 @@ jest.mock("../../../../components/AgentChat/Common/LlmChatButton", () => ({
             onClick={onClick}
             disabled={disabled}
             data-testid="microphone-button"
+            type="button"
```
dsargent, line 55: Satisfy `button-has-type` rule.

packages/ui-common/__tests__/components/ChatBot/ChatBot.test.tsx
```diff
@@ -22,6 +22,7 @@ jest.mock("../../../components/AgentChat/ChatCommon/ChatCommon", () => ({
             <button
                 onClick={props.onClose}
                 data-testid="close-button"
+                type="button"
```
dsargent, line 25: Satisfy `button-has-type` rule.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -127,7 +127,8 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
         customURLLocalStorage?.replaceAll('"', "") || backendNeuroSanApiUrl
     )
 
-    const agentCountsRef = useRef<Map<string, number>>(new Map())
+    // Tracks how many times each agent has been involved in the conversation
```
dsargent, line 130: I originally made this a "ref" to prevent re-renders, but copilot convinced me it should be State, which also placates the linter rule `react-hooks/refs`.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -322,7 +323,7 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
             // If the selected network is one of the expired ones, deselect it
             if (expiredNetwork.some((n) => n.agentInfo.agent_name === selectedNetwork)) {
                 setSelectedNetwork(null)
-                agentCountsRef.current = new Map()
+                setAgentCounts(new Map())
```
dsargent, line 326: Change counts to be state

packages/ui-common/components/MultiAgentAccelerator/ThoughtBubbleOverlay.tsx
```diff
@@ -127,6 +127,14 @@ export const ThoughtBubbleOverlay: FC<ThoughtBubbleOverlayProps> = ({
     isStreaming = false,
     onBubbleHoverChange,
 }) => {
+    /**
```
dsargent, line 130: See comment. This is to placate the `react-hooks/purity` rule; in React you're not supposed to have impure functions like `Date.now()` in the render loop, so this (recommend by the React dev team on the [page](https://react.dev/reference/eslint-plugin-react-hooks/lints/purity) for the same rule) is a compromise. I'm not 100% sure it's right, but I think we want to redo thought bubbles anyway. Open to discussion as always.

packages/ui-common/components/MultiAgentAccelerator/ThoughtBubbleOverlay.tsx
```diff
@@ -150,6 +158,15 @@ export const ThoughtBubbleOverlay: FC<ThoughtBubbleOverlayProps> = ({
     const rafRef = useRef<ReturnType<typeof requestAnimationFrame> | null>(null)
     const mountedRef = useRef(true)
 
+    // Effect to update `time` every second to trigger re-renders for animation timing and line updates.
```
dsargent, line 161: The idea is we capture a "fixed" version of "now" every (some interval, set here to 250ms) and use that in the render loop.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -468,6 +471,16 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
         )
     }
 
+    const onNetworkReplaced = useCallback(
```
dsargent, line 474: Memoized this -- it's passed as a prop to the child component, which means unnecessary re-renders unless we stabilize it as a callback.

packages/ui-common/components/MultiAgentAccelerator/ThoughtBubbleOverlay.tsx
```diff
@@ -216,7 +232,7 @@ export const ThoughtBubbleOverlay: FC<ThoughtBubbleOverlayProps> = ({
 
             return newState
         })
-    }, [thoughtBubbleEdges])
+    }, [thoughtBubbleEdges, time])
```
dsargent, line 235: Have to add it as dep now, which will cause a rerender every 250ms but I think that's okay and expected.

packages/ui-common/components/MultiAgentAccelerator/ThoughtBubbleOverlay.tsx
```diff
@@ -248,9 +264,9 @@ export const ThoughtBubbleOverlay: FC<ThoughtBubbleOverlayProps> = ({
         for (const node of nodes) {
             const getConversations = node.data?.["getConversations"]
             if (typeof getConversations === "function") {
-                const convs = getConversations()
-                if (Array.isArray(convs)) {
-                    const hasSelf = convs.some((conv) => conv?.agents?.has?.(node.id))
+                const conversations = getConversations()
```
dsargent, line 267: Not a fan of unnecessary abbreviations!

swensel: I'll keep that in mind.

swensel (approved):
LGTM

packages/ui-common/components/MultiAgentAccelerator/ThoughtBubbleOverlay.tsx
dsargent: Dan gets annoyed by cryptic abbreviations and acronyms too. Text is cheap, unless we really need to abbreviate or something is egregiously long, let's use the full name. Aside, helps with grepping too -- if I was looking for all code related to `conversations`, it wouldn't have found this previously.

Issue #259, opened 2026-04-17
Bug in how env vars get set up in no-auth mode


dsargent:
See `_app.tsx`, if (!authenticationEnabled()) {`  test causes some vars not to be set, like `logoServiceToken`.

PR #330, opened 2026-05-15
#259 Disable all next-auth functionality when authentication is off

dsargent:
Properly degrades gracefully to "anonymous" when authentication is disabled, with no ugly `next-auth` errors in the log.

Also addresses https://github.com/cognizant-ai-lab/neuro-san-ui/issues/259 and paves the way for https://github.com/cognizant-ai-lab/neuro-san-studio/issues/773

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 # Node/NextJS/dotenv things
-.env*
+.env
 .env.local
 **/.next
 **/next-env.d.ts
--- a/apps/main/.env.sample
+++ b/apps/main/.env.sample
@@ -1,0 +1,14 @@
+# Determines which backend neuro-san server to access. This one is for the Dev environment -- change as necessary.
+NEURO_SAN_SERVER_URL=<URL for your neuro-san server including port number>
+
+# Set this to false for testing or authentication via load balancer (recommended)
+NEXT_PUBLIC_ENABLE_AUTHENTICATION=false
+
+# Set this to your OpenAI API key (required)
+OPENAI_API_KEY=<Your OpenAI API key here>
+
+# Set this to your support email address
+SUPPORT_EMAIL_ADDRESS=test@example.com
+
+# Set this to your Logo.dev token (for automated logo suggestions in the UI)
+LOGO_SERVICE_TOKEN=<Your logo service token here>
--- a/apps/main/__tests__/pages/App.test.tsx
+++ b/apps/main/__tests__/pages/App.test.tsx
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+import type {EnvironmentResponse} from "../../pages/api/environment/Types"
 import {render, screen, waitFor} from "@testing-library/react"
 import {default as userEvent, UserEvent} from "@testing-library/user-event"
 import {useRouter} from "next/router"
@@ -21,7 +22,7 @@ import {ReactNode} from "react"
 
 import {withStrictMocks} from "../../../../__tests__/common/strictMocks"
 import {mockFetch} from "../../../../__tests__/common/TestUtils"
-import {authenticationEnabled, DEFAULT_NEURO_SAN_SERVER_URL} from "../../../../packages/ui-common/const"
+import {authenticationEnabled} from "../../../../packages/ui-common/const"
 import {useEnvironmentStore} from "../../../../packages/ui-common/state/Environment"
 import {useAuthentication} from "../../../../packages/ui-common/utils/Authentication"
 import NeuroSanUI from "../../pages/_app"
@@ -59,6 +60,8 @@ describe("Main App Component", () => {
     const testClientId = "testClientId"
     const testDomain = "testDomain"
     const testSupportEmailAddress = "test@example.com"
+    const testLogoServiceToken = "testLogoServiceToken"
+
     beforeEach(() => {
         ;(useAuthentication as jest.Mock).mockReturnValue({
             data: {user: {name: "mock-user", image: "mock-image-url"}},
@@ -71,18 +74,20 @@ describe("Main App Component", () => {
             auth0ClientId: null,
             auth0Domain: null,
             supportEmailAddress: null,
+            logoServiceToken: null,
         })
 
         user = userEvent.setup()
 
-        window.fetch = mockFetch({
+        const mockEnvironmentResponse: EnvironmentResponse = {
             backendNeuroSanApiUrl: testNeuroSanURL,
             auth0ClientId: testClientId,
             auth0Domain: testDomain,
             supportEmailAddress: testSupportEmailAddress,
-            oidcHeaderFound: true,
-            username: "testUser",
-        })
+            logoServiceToken: testLogoServiceToken,
+        }
+
+        window.fetch = mockFetch(mockEnvironmentResponse as Record<string, unknown>)
         ;(useRouter as jest.Mock).mockReturnValue({
             pathname: "/projects",
         })
@@ -113,6 +118,7 @@ describe("Main App Component", () => {
         expect(state.auth0ClientId).toBe(testClientId)
         expect(state.auth0Domain).toBe(testDomain)
         expect(state.supportEmailAddress).toBe(testSupportEmailAddress)
+        expect(state.logoServiceToken).toBe(testLogoServiceToken)
     })
 
     it("Should render correctly when authentication is disabled", async () => {
@@ -124,10 +130,11 @@ describe("Main App Component", () => {
 
         // No authentication so values should be defaults
         const state = useEnvironmentStore.getState()
-        expect(state.backendNeuroSanApiUrl).toBe(DEFAULT_NEURO_SAN_SERVER_URL)
+        expect(state.backendNeuroSanApiUrl).toBe(testNeuroSanURL)
         expect(state.auth0ClientId).toBe("")
         expect(state.auth0Domain).toBe("")
-        expect(state.supportEmailAddress).toBe("")
+        expect(state.supportEmailAddress).toBe(testSupportEmailAddress)
+        expect(state.logoServiceToken).toBe(testLogoServiceToken)
     })
 
     it("Should handle failure to fetch environment variables", async () => {
@@ -148,6 +155,7 @@ describe("Main App Component", () => {
         expect(state.auth0ClientId).toBe(null)
         expect(state.auth0Domain).toBe(null)
         expect(state.supportEmailAddress).toBe(null)
+        expect(state.logoServiceToken).toBe(null)
 
         expect(consoleSpy).toHaveBeenCalledTimes(2)
 
--- a/apps/main/instrumentation.ts
+++ b/apps/main/instrumentation.ts
@@ -15,7 +15,7 @@ limitations under the License.
 */
 
 /**
- * This file executes at NextJS startup. It is intended for injecting instrumentation and monitoring but here
+ * This file executes at Next.js startup. It is intended for injecting instrumentation and monitoring but here
  * we are using it to check that the environment variables are set correctly.
  *
  * @see https://nextjs.org/docs/app/guides/instrumentation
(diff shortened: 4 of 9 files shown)


apps/main/__tests__/pages/App.test.tsx
```diff
@@ -14,14 +14,15 @@ See the License for the specific language governing permissions and
...
 
+import type {EnvironmentResponse} from "../../pages/api/environment/Types"
 import {render, screen, waitFor} from "@testing-library/react"
 import {default as userEvent, UserEvent} from "@testing-library/user-event"
 import {useRouter} from "next/router"
 import {ReactNode} from "react"
 
 import {withStrictMocks} from "../../../../__tests__/common/strictMocks"
 import {mockFetch} from "../../../../__tests__/common/TestUtils"
-import {authenticationEnabled, DEFAULT_NEURO_SAN_SERVER_URL} from "../../../../packages/ui-common/const"
```
dsargent, line 24: Removed default Neuro-san URL. It's a trade off between "works out of the box, zero config" vs. "we probably don't want our dev server baked into the code". We've grown and matured to the point where I favor the latter, though it saddens me that you can't just run MAUI "out of the box" without configuring stuff.

apps/main/__tests__/pages/App.test.tsx
```diff
@@ -59,6 +60,8 @@ describe("Main App Component", () => {
     const testClientId = "testClientId"
     const testDomain = "testDomain"
     const testSupportEmailAddress = "test@example.com"
+    const testLogoServiceToken = "testLogoServiceToken"
```
dsargent, line 63: Added extra tests around this new-ish field.

apps/main/__tests__/pages/App.test.tsx
```diff
@@ -71,18 +74,20 @@ describe("Main App Component", () => {
             auth0ClientId: null,
             auth0Domain: null,
             supportEmailAddress: null,
+            logoServiceToken: null,
         })
 
         user = userEvent.setup()
 
-        window.fetch = mockFetch({
+        const mockEnvironmentResponse: EnvironmentResponse = {
```
dsargent, line 82: Add strong typing. Just by doing that it found a couple of invalid fields in the test data, which I've removed.

apps/main/__tests__/pages/App.test.tsx
```diff
@@ -71,18 +74,20 @@ describe("Main App Component", () => {
...
             auth0ClientId: testClientId,
             auth0Domain: testDomain,
             supportEmailAddress: testSupportEmailAddress,
-            oidcHeaderFound: true,
-            username: "testUser",
-        })
+            logoServiceToken: testLogoServiceToken,
+        }
+
+        window.fetch = mockFetch(mockEnvironmentResponse as Record<string, unknown>)
```
dsargent, line 90: Ugly `as` cast, but it's hard to avoid here since `mockFetch` is super general but here we want it to return a particular shape of response.

One way to make it type safe and avoid the cast: make `mockFetch` take a generic param for the return type. Felt like too much upheaval for this PR.

apps/main/__tests__/pages/App.test.tsx
```diff
@@ -113,6 +118,7 @@ describe("Main App Component", () => {
         expect(state.auth0ClientId).toBe(testClientId)
         expect(state.auth0Domain).toBe(testDomain)
         expect(state.supportEmailAddress).toBe(testSupportEmailAddress)
+        expect(state.logoServiceToken).toBe(testLogoServiceToken)
```
dsargent, line 121: Extra asserts. More asserts = better tests.

apps/main/__tests__/pages/App.test.tsx
```diff
@@ -124,10 +130,11 @@ describe("Main App Component", () => {
 
         // No authentication so values should be defaults
         const state = useEnvironmentStore.getState()
-        expect(state.backendNeuroSanApiUrl).toBe(DEFAULT_NEURO_SAN_SERVER_URL)
```
dsargent, line 127: We no longer default to the hardcoded default URL.

apps/main/pages/_app.tsx
```diff
@@ -87,6 +82,14 @@ const NavbarWrapper = (props: Omit<NavbarProps, "userInfo">): ReactElement => {
     )
 }
 
+/**
```
dsargent, line 85: This is part of compartmentalizing `next-auth`. See comment.

apps/main/pages/_app.tsx
```diff
@@ -145,17 +148,6 @@ export default function NeuroSanUI({Component, pageProps}: ExtendedAppProps): Re
 
     useEffect(() => {
         const getEnvironment = async () => {
-            if (!authenticationEnabled()) {
```
dsargent, line 148: Refactored the logic here to fix https://github.com/cognizant-ai-lab/neuro-san-ui/issues/259

apps/main/pages/_app.tsx
```diff
@@ -172,14 +164,22 @@ export default function NeuroSanUI({Component, pageProps}: ExtendedAppProps): Re
...
 
-            // Save env vars in zustand store
+            // We should always have these whether authentication is enabled or not
             setBackendNeuroSanApiUrl(data.backendNeuroSanApiUrl)
-            setAuth0ClientId(data.auth0ClientId)
-            setAuth0Domain(data.auth0Domain)
             setSupportEmailAddress(data.supportEmailAddress)
             setLogoServiceToken(data.logoServiceToken)
+
+            if (authenticationEnabled()) {
```
dsargent, line 175: The logic is much cleaner and more solid now. We set up the "common" items, and only if authentication is enabled, do we grab the relevant items for that.

apps/main/pages/_app.tsx
```diff
@@ -302,10 +302,7 @@ export default function NeuroSanUI({Component, pageProps}: ExtendedAppProps): Re
         body = (
             <>
                 <CssBaseline />
-                {/*Note: Still need the NextAuth SessionProvider even in ALB case since we have to use useSession
```
dsargent, line 305: Comment no longer applies as I cleaned this up.

apps/main/pages/_app.tsx
```diff
@@ -302,10 +302,7 @@ export default function NeuroSanUI({Component, pageProps}: ExtendedAppProps): Re
...
-                unconditionally due to React hooks rules. But it doesn't interfere with ALB log on and will be
-                removed when we fully switch to ALB auth.*/}
-                <SessionProvider>
+                <NullableSessionProvider>
```
dsargent, line 305: Use new `NullableSessionProvider` (couldn't think of a better name) to conditionally wrap with a `next-auth` `SessionProvider` _only_ when authentication is enabled.

apps/main/.env.sample
```diff
@@ -0,0 +1,14 @@
+# Determines which backend neuro-san server to access. This one is for the Dev environment -- change as necessary.
```
dsargent, line 1: Preparing for the next PR which will be "how to set up MAUI" instructions. This is a minimal-but-complete sample of what is required for an authentication-free MAUI set-up. Note the lack of any "auth"-related variables. Without this PR, if you omitted all those `next-auth` env vars, even with authentication turned off it would spew errors into the server log.

Now, it runs clean, no errors with or without authentication as long as you have set up the env vars it needs for whatever case you're using.

apps/main/instrumentation.ts
```diff
@@ -15,7 +15,7 @@ limitations under the License.
 */
 
 /**
- * This file executes at NextJS startup. It is intended for injecting instrumentation and monitoring but here
+ * This file executes at Next.js startup. It is intended for injecting instrumentation and monitoring but here
```
dsargent, line 18: Branding

apps/main/instrumentation.ts
```diff
@@ -27,15 +27,22 @@ import {authenticationEnabled} from "@cognizant-ai-lab/ui-common/const"
  * List of environment variables that are required for the app to run. If any of these are not set, the app will
  * exit with an error message.
  */
-export const REQUIRED_ENV_VARS = ["NEURO_SAN_SERVER_URL", "SUPPORT_EMAIL_ADDRESS"]
+export const REQUIRED_ENV_VARS = ["NEURO_SAN_SERVER_URL", "OPENAI_API_KEY", "SUPPORT_EMAIL_ADDRESS"]
 
-export const REQUIRED_FOR_AUTH_ENV_VARS = ["AUTH0_CLIENT_ID", "AUTH0_CLIENT_SECRET", "AUTH0_DOMAIN", "AUTH0_ISSUER"]
+export const REQUIRED_FOR_AUTH_ENV_VARS = [
```
dsargent, line 32: Added `next-auth` vars as they required when authentication is enabled.

apps/main/instrumentation.ts
```diff
@@ -27,15 +27,22 @@ import {authenticationEnabled} from "@cognizant-ai-lab/ui-common/const"
  * List of environment variables that are required for the app to run. If any of these are not set, the app will
  * exit with an error message.
  */
-export const REQUIRED_ENV_VARS = ["NEURO_SAN_SERVER_URL", "SUPPORT_EMAIL_ADDRESS"]
+export const REQUIRED_ENV_VARS = ["NEURO_SAN_SERVER_URL", "OPENAI_API_KEY", "SUPPORT_EMAIL_ADDRESS"]
```
dsargent, line 30: Made `OPENAI_API_KEY` mandatory -- we need it for icon suggestions. Arguable whether it's actually mandatory. I think without it we degrade to the generic icons, but you'll probably get errors in the server log. Plus, we want users to have the "full MAUI experience" when they depoy it.

packages/ui-common/__tests__/unit/Utils/Authentication.test.ts
```diff
@@ -39,42 +41,56 @@ describe("useAuthentication", () => {
...
+        process.env["NEXT_PUBLIC_ENABLE_AUTHENTICATION"] = "true"
     })
 
     afterEach(() => {
         window.fetch = oldFetch
     })
 
-    it("Does nothing if currentUser not available", async () => {
-        ;(signOut as jest.Mock).mockResolvedValueOnce(undefined)
+    describe("useAuthentication", () => {
```
dsargent, line 52: Add extra tests for the auth logic, which wasn't properly covered.

Also move tests into sub-suites based on which function they're testing. More organized.

packages/ui-common/utils/Authentication.ts
```diff
@@ -33,11 +34,16 @@ export const AD_TENANT_ID = "de08c407-19b9-427d-9fe8-edf254300ca7"
  * of NextAuth's useSession hook.
  */
 export const useAuthentication = () => {
+    if (!authenticationEnabled()) {
```
dsargent, line 37: New logic to degrade gracefully when authentication is disabled, so consumers can blindly use this hook without worrying.

packages/ui-common/utils/Authentication.ts
```diff
@@ -33,11 +34,16 @@ export const AD_TENANT_ID = "de08c407-19b9-427d-9fe8-edf254300ca7"
...
+        // Auth disabled: return a stub
+        return {data: {user: {name: DEFAULT_USERNAME, image: DEFAULT_USER_IMAGE}}}
+    }
+    // Auth enabled: safe to call hooks. Despite the conditional test above, we are guaranteed that
```
dsargent, line 41: See comment. Technically we're "breaking rules of hooks" here due to the conditional statement, but "we know something the linter doesn't". Namely that `authenticationEnabled` is set at build/process launch time (based on an env var) so the order of hook execution will be constant for any particular server instance/Docker build, as required by React.js.

packages/ui-common/const.ts
```diff
@@ -25,12 +25,15 @@ export const getContactUsConfirmationText = (email: string): string =>
...
+export const DEFAULT_USERNAME = "Guest"
+
 /**
  * The default user image to use when the user does not have a profile picture.
  */
 export const DEFAULT_USER_IMAGE = "https://www.gravatar.com/avatar/?d=mp"
 
 export const authenticationEnabled = (): boolean => process.env["NEXT_PUBLIC_ENABLE_AUTHENTICATION"] !== "false"
-
-// Default "dev URL" for NeuroSan server, to allow for "zero config" execution.
```
dsargent, line 35: Removed.

packages/ui-common/const.ts
```diff
@@ -25,12 +25,15 @@ export const getContactUsConfirmationText = (email: string): string =>
     "to continue. If you don't have an email client, you can still contact us at " +
     `${email} using a web based email client.`
 
+/**
```
dsargent, line 28: For "auth disabled" case.

.gitignore
```diff
@@ -1,5 +1,5 @@
 # Node/NextJS/dotenv things
-.env*
+.env
```
dsargent, line 2: Had to narrow this pattern to allow for `.env.sample`.

jest.config.ts
```diff
@@ -72,10 +72,10 @@ const config: Config.InitialOptions = {
 
     coverageThreshold: {
         global: {
-            statements: -102,
-            branches: -160,
+            statements: -101,
```
dsargent, line 75: Slight bump 🎉

apps/main/__tests__/pages/App.test.tsx
swensel: Removing the default may help with token consumption too, although we really should have auth on the neuro-san side (at least through the ALB, if not through the library itself), so it's not just security through obscurity (or in this case, preventing overusage through obscurity).

packages/ui-common/utils/Authentication.ts
dsargent: The only risk is, if some day, some miscreant were to modify `authenticationEnabled` so that it depends on something dynamic that could change at runtime, _then_ we'd be breaking the "rules of hooks", the linter wouldn't detect it and React.js would be super upset.

apps/main/pages/_app.tsx
swensel: `OptionalSessionProvider` or `OptInSessionProvider`, maybe? `NullableSessionProvider` is fine though.

apps/main/instrumentation.ts
swensel: Icon suggestions are a nice to have. I'm not sure I'd force this as required. Perhaps a different ticket to take care of any errors so this feature is truly optional?

swensel: FWIW the icon suggestion API calls (`/api/agentIconSuggestions` and `/api/networkIconSuggestions`) were also the hang up for ZScaler SSL cert inspection, although that's a separate security issue that needs to be handled through policy (although also to unblock LMX on local).

packages/ui-common/utils/Authentication.ts
swensel: Thanks for adding a comment.

swensel (approved):
LGTM. 

Since this is authentication related, it does bring up overall authentication considerations like: 

1) Neuro-san shouldn't just trust the user name we send.

2) The other point I had with removing the default Neuro-san server. Sure that helps with token usage (although that's not the main reason it was removed here), but really we need more mature authentication in Neuro-san. 

However, neither should be part of this PR (or repo).

Thanks for making the no auth case easier for users in MAUI. 🎉

PR #333, opened 2026-05-16
Make editable node more obvious - #301

swensel:
# Description

Make editable node more obvious - #301

# Screenshot 
### This is more subtle, but on hover the node becomes larger. May need another indicator that's not tied to hover.

https://github.com/user-attachments/assets/b93a819f-e852-4b70-8bcf-c2f4515e091d

# Screenshot with edit icon
### Another idea, currently commented out. I don't quite like it as much, but it could be refined.

Screenshot 2026-05-15 at 5 01 38 PM

### DIFF
--- a/packages/ui-common/components/MultiAgentAccelerator/AgentNode.tsx
+++ b/packages/ui-common/components/MultiAgentAccelerator/AgentNode.tsx
@@ -18,6 +18,7 @@ limitations under the License.
 // eslint-disable-next-line no-restricted-imports
 import * as MuiIcons from "@mui/icons-material"
 import AutoAwesomeIcon from "@mui/icons-material/AutoAwesome"
+// import EditOutlinedIcon from "@mui/icons-material/EditOutlined"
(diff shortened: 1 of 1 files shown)


swensel:
Here's some ideas from Claude Design.

I think `Single small mark - D - Corner dot` is a good idea. Although usually a UX element like that indicates something already edited, or that needs attention (like a notification of the number of messages on the Slack icon). Perhaps incorporating something like that would be good, when a user already has edited fields, instead of it being a perpetual "you can edit this" indicator. We could also add a "return to default" or something along those lines, if we kept track of the edits that the user has done that are different than the default prompts.

`Flip the default - H - Dim the read-only` also could be good. Since most of the nodes are editable, this could be a decent option that's not tied to hover or existing user edits, and it could be perpetually present.

Perhaps one of the `Dashed rings` or `Dotted orbits`, can work, as well, although I'm not sure if I like those as much or not. I'll think about it.

Screenshot 2026-05-15 at 7 05 07 PM

Screenshot 2026-05-15 at 7 05 13 PM

Issue #335, opened 2026-05-16
New user tour


dsargent:
When a user launches MAUI for the first time, show them around. We have a lot more features now so discoverability is an issue.

Important: skippable! and only nags you once.

Optional: "what's new" only?

<img width="800" height="388" alt="Image" src="https://github.com/user-attachments/assets/365ab9e1-5c60-4a0c-bd9c-a22044cdef0e" />

PR #336, opened 2026-05-18
#335 Guided tour for new users

dsargent:
Implements a "welcome tour" for users new to the application. After a while (5 seconds) the tour is shown automatically if we detect we haven't showed it before (via local storage persistence) but the tour can also be launched manually from the relevant `Help` menu item at any time.

Demo:

https://github.com/user-attachments/assets/5e1a1797-a05e-419b-8764-7fc682db227f


apps/main/pages/_app.tsx
```diff
@@ -73,11 +74,17 @@ const DEFAULT_APP_NAME = `Cognizant ${LOGO}`
 const NavbarWrapper = (props: Omit<NavbarProps, "userInfo">): ReactElement => {
     const {data} = useAuthentication()
     const userInfo = data?.user
+    const onStartTour = () => {
```
dsargent, line 77: Plumbing to launch the tour from the help menu. After a lot of discussion with the AIs, I ended up using a custom event. The other options were things like: shared components or shared zustand state. This seemed like the lesser evil.

Another thing I considered is having the app own the tour, because it's the common parent of the items that are in the tour. But I didn't like that either, because the app is meant to be a general container that knows as little as possible about the things it contains.

It would also make it harder to do what we do currently, which is select a network if the tour is launched but the user hasn't selected one. That would be very awkward to do from the app component and would probably require lifting a ton of things we don't want to.

As a compromise, the tour is now owned by the MAA component. We can revisit that later.

apps/main/pages/_app.tsx
```diff
@@ -73,11 +74,17 @@ const DEFAULT_APP_NAME = `Cognizant ${LOGO}`
...
+        // Dispatches the "start tour" event
+        window.dispatchEvent(new CustomEvent(TRIGGER_APP_TOUR_EVENT_NAME))
+    }
+
     return (
         <Navbar
             {...props}
             id="nav-bar"
             userInfo={userInfo}
+            onStartTour={onStartTour}
```
dsargent, line 87: New handler for the Help menu item.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -905,20 +905,25 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
-                left: 0,
-                right: 0,
-                bottom: 0,
-                zIndex: theme.zIndex.modal - 1,
-                cursor: "not-allowed",
-                // Capture all pointer events to prevent interaction with the chat when no agent is selected
-                pointerEvents: "all",
-            }}
-        />
+        <Tooltip
```
dsargent, line 908: Not strictly related to the tour, but usability: show a tooltip when chat is disable to tell the user what's going on.

packages/ui-common/components/Common/Navbar.tsx
```diff
@@ -332,6 +336,17 @@ export const Navbar = ({
                     >
                         User guide
                     </MenuItem>
+                    <MenuItem
```
dsargent, line 339: New menu item

packages/ui-common/components/MultiAgentAccelerator/Tour/MainTourSteps.ts
dsargent, line 1: These are the steps of the tour. Having them in a separate module makes it easier to update. But see one 🙁 thing below.

packages/ui-common/components/MultiAgentAccelerator/Tour/MainTourSteps.ts
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * @see https://react-joyride.com/docs/step
+ */
+export const MAIN_TOUR_STEPS: Step[] = [
+    {
+        // First step doesn't refer to any particular UI element, just a welcome message
+        content:
+            "Welcome to Cognizant AI Lab Multi-Agent Accelerator! This tour will give you a quick overview of " +
+            "the application.",
+        target: () => document.querySelector("#multi-agent-accelerator-grid"),
```
dsargent, line 14: This is the ugly part. 

There's several ways to target elements for the tour. See `target` in the [doc](https://react-joyride.com/docs/step).

By far the easiest is DOM queries like these. This also means the tour steps have no compile-time connection to the targeted elements. "They're just strings". But that's a double-edged sword: if we make a typo int he ID, or someone changes the ID without searching for whatever reason, 💥 that item is "no longer a stop on the tour". So this puts us in a "breaking Vince's tests" type situation with the IDs.

Another option is using React RefObjects, à la `<Foo ref={bar}>`, then we would have `target: bar` in the corresponding step. The immediate, huge issue with that is that then we'd have to pass all those ref objects around, in particularly feeding them to these steps. So this module would suddenly have to be a function with 20 or so parameters, all the refs.

The massive upside would be that if we get something wrong, `tsc` will tell us. If someone refactors, `tsc` will scream and we'll know the tour would be broken at compile time.

The jury is still a bit out on this. I'm not sure, but this works for now, and is super simple.

packages/ui-common/components/MultiAgentAccelerator/const.ts
```diff
@@ -48,6 +48,9 @@ export const AGENT_NETWORK_DEFINITION_KEY = "agent_network_definition"
 // The key in sly_data where the agent network name is stored
 export const AGENT_NETWORK_NAME_KEY = "agent_network_name"
 
+// The event name that needs to be fired for the app tour to start
+export const TRIGGER_APP_TOUR_EVENT_NAME = "trigger-app-tour"
```
dsargent, line 52: Single source of truth for the event name.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -205,6 +217,36 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
         [agentsInNetwork]
     )
 
+    // Introductory tour
```
dsargent, line 220: Tour-related stuff.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -205,6 +217,36 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
...
+
+    // Track that the user requested the tour, and we should start it once network data is available
+    const [tourRequested, setTourRequested] = useState<boolean>(false)
+
+    // Tour persisted status
+    const tourStatus = useTourStore((s) => s.status)
+    const setTourStatus = useTourStore((s) => s.setStatus)
+
+    const {controls, Tour} = useJoyride({
```
dsargent, line 229: There's a purely declarative version with `<Tour />` but using the hook gives us more flexibility and for our case, I prefer the more imperative version.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -392,6 +434,27 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
         return () => clearInterval(interval)
     }, [changeSelectedNetwork, temporaryNetworks, selectedNetwork])
 
+    const dismissTourModal = () => {
+        setTourModalOpen(false)
+        setHaveShownTourModal(true)
+    }
+
+    const handleExternalTourRequest = useCallback(() => {
```
dsargent, line 442: This is required to handle the case where the user selects `Help | Tour`. The issue with that is, that item lives in `Navbar`, and we are in `MAA`; the common parent is the App module, so it has to adjudicate for us. C'est la vie with React.js.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -392,6 +434,27 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
...
+        if (selectedNetwork == null && networks?.length > 0) {
+            setSelectedNetwork(networks[0].agent_name)
+        }
+
+        // Close the modal if open
+        setTourModalOpen(false)
+        setTourRequested(true)
+    }, [networks, selectedNetwork])
+
+    useEffect(() => {
```
dsargent, line 453: Listen for the event from the Navbar menu.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -539,6 +600,40 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
         }
     }
 
+    useEffect(() => {
```
dsargent, line 603: These effects handle the tour state.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -695,7 +790,7 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
         )
     }
 
-    const getConfirmationModal = () =>
+    const getDeleteNetworkConfirmationModal = () =>
```
dsargent, line 793: Make name more explicit now that we have a couple of modals we're juggling.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -724,6 +819,60 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
             />
         ) : null
 
+    const getTourModal = () =>
```
dsargent, line 822: Nag modal to see if the user wants to take the tour.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -787,8 +936,10 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
 
     return (
         <>
+            {Tour}
```
dsargent, line 939: This is where insert the Tour item (and the modal just below)

packages/ui-common/state/Tour.ts
dsargent, line 1: For remembering when the user has responded to the nag dialog

packages/ui-common/state/UserInfo.ts
```diff
@@ -21,7 +21,7 @@ import {OidcProvider} from "../utils/types"
 /**
  * Zustand state store for "feature flags" like, "generic branding".
  */
-export interface UserInfoStore {
+interface UserInfoStore {
```
dsargent, line 24: I cleaned up some stuff so this no longer needs to be exported, and therefore Knip _inists_ that it not be exported.

packages/ui-common/utils/zIndexLayers.ts
```diff
@@ -29,6 +29,9 @@ export const getZIndex = (layer: number, appTheme: Theme): number => {
             // an MUI Drawer, however, we also want to ensure that LAYER_2 is below the modal, which currently has a
             // default z-index of 1300.
             return (appTheme.zIndex.modal + appTheme.zIndex.drawer) / 2
+        case 3:
+            // Initially for the "guided tour" which should be on top of everything
```
dsargent, line 33: As part of the never-ending z-index arms race, added this layer for the tour, which needs to be on top of everything.

packages/ui-common/package.json
```diff
@@ -17,6 +17,7 @@
         "jsonrepair": "3.8.0",
         "lodash-es": "catalog:default",
         "notistack": "catalog:default",
+        "react-joyride": "catalog:default",
```
dsargent, line 20: Add dep

.yarnrc.yml
```diff
@@ -29,6 +29,7 @@ catalogs:
         "lodash-es": "4.18.1"
         "next-auth": "5.0.0-beta.30"
         "notistack": "3.0.2"
+        "react-joyride": "3.1.0"
```
dsargent, line 32: Add dep to catalog. I think I'm moving to a place where "everything should be in the catalog unless there's a very good reason not to". It does make it easier than digging through the various `package.json` files in the monoreo.

jest.config.ts
```diff
@@ -67,16 +67,17 @@ const config: Config.InitialOptions = {
...
 
     coverageReporters: ["text-summary"],
 
     coverageThreshold: {
         global: {
-            statements: -125,
-            branches: -159,
-            functions: -30,
-            lines: -99,
+            statements: -120,
```
dsargent, line 77: I'll take it. Major new feature, 3 out of 4 improved, one slightly down.

This was some _hard won_ coverage.

apps/main/__tests__/pages/App.test.tsx
```diff
@@ -174,4 +175,23 @@ describe("Main App Component", () => {
         await screen.findByText(COMPONENT_BODY)
         expect(document.getElementById("body-div")).toBeInTheDocument()
     })
+
+    it("Should launch the tour when the items is selected from the Help menu", async () => {
```
dsargent, line 179: Add test for launching Tour from `Help` menu.

apps/main/__tests__/pages/MultiAgentAcceleratorPage.test.tsx
dsargent, line 1: I've been wanting to do this refactoring for a while. This test suite was named like it was testing the MAA page, but it was actually testing the MAA _component_ which lives within that page.

As part of this PR, I:

1) Moved all the component related tests to a new suite, in the appropriate place
2) Hollowed out this test and made it so it _only_ tests the MAA page itself, as its name implies.

apps/main/__tests__/pages/MultiAgentAcceleratorPage.test.tsx
```diff
@@ -14,1192 +14,57 @@ See the License for the specific language governing permissions and
...
 
-const conversationMock = jest.fn()
-const temporaryNetworksMock = jest.fn()
-const networkIconSuggestionsMock = jest.fn()
-let onDeleteNetwork: (a: string, b: boolean) => void
-let setSelectedNetwork: (network: string) => void
+const mockMultiAgentAcceleratorSpy = jest.fn()
 
-jest.mock("../../../../packages/ui-common/components/MultiAgentAccelerator/AgentFlow", () => ({
+jest.mock("../../../../packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator", () => ({
```
dsargent, line 40: The page itself doesn't really do much; it's just a host for the MMA component. So we just make sure it renders reasonably, and passes on the right props to the MAA component which does all the heavy lifting.

packages/ui-common/__tests__/components/MultiAgentAccelerator/MultiAgentAccelerator.test.tsx
dsargent, line 1: This is where all the stuff that _was_ in `MultiAgentAcceleratorPage.test.tsx` was moved to. It rightfully belongs here since it's testing the component, not the page.

If you want to see what changed:

`git difftool origin/main:apps/main/__tests__/pages/MultiAgentAcceleratorPage.test.tsx origin/DS-guided-tour:packages/ui-common/__tests__/components/MultiAgentAccelerator/MultiAgentAccelerator.test.tsx`

packages/ui-common/__tests__/components/MultiAgentAccelerator/MultiAgentAccelerator.test.tsx
```diff
@@ -0,0 +1,1290 @@
...
+                        extraSlyData: {
+                            [AGENT_NETWORK_DEFINITION_KEY]: agentNetworkDefinition,
+                        },
+                    })
+                )
+            })
+        })
+    })
+
+    describe("Tour", () => {
```
dsargent, line 1200: Biggest change in this file is this new test suite for tours.

packages/ui-common/components/Common/ConfirmationModal.tsx
```diff
@@ -22,7 +22,7 @@ import {FC, JSX as ReactJSX, ReactNode, useState} from "react"
 import {MUIDialog} from "./MUIDialog"
 
 // #region: Styled Components
-const StyledButton = styled(Button)({
+export const StyledButton = styled(Button)({
```
dsargent, line 25: Export this so we can reuse it. Didn't seem worth moving to a separate "styled components" module for this PR, but we could.

packages/ui-common/components/MultiAgentAccelerator/Tour/MainTourSteps.ts
swensel: I may prefer `refs` here, but this is OK for now. I understand what you're saying about a lot of parameters passed into the function.

swensel (approved):
Cool feature. Thanks for putting this together.

packages/ui-common/components/MultiAgentAccelerator/Tour/MainTourSteps.ts
dsargent: It's not just that. The Tour visits a lot of places, so the refs would have to all be collected from child components and passed over to the steps. Even if we moved the steps into MAA, we'd still be needing to prop drill for the refs. Might be a way to do it with Zustand. Next version maybe.

swensel: Another option could be React Context instead of prop drilling.

PR #338, opened 2026-05-18
Fix stale state after temporary network deletion

souravJBcognizant:
Issue #264 
Deleting a temporary network only removes its entry from the sidebar. Other state tied to that network sticks around:

The graph view keeps rendering the deleted network if it was the active one. The user has to manually click another network (or refresh) to clear it.

`Chat history` & `sly_data` in `IndexedDB` stay indefinitely under the deleted network's agent_name. They'd accumulate over time and could leak into a re-created network with the same name.

`agentCounts` keeps the per-agent message counts from the deleted network, which would briefly color-leak into the next selected network in heatmap mode.

This affects all three deletion paths:

1. User clicks trash on a non-expired network and confirms the modal.
2. User clicks trash on an already-expired network (immediate delete, no modal).
3. Auto-reaper purges a network whose 5-minute grace period has elapsed.

Fix
Make all three paths perform the same full cleanup:

If the deleted network was the currently-selected one, clear `selectedNetwork` and reset `agentCounts`. The graph clears in the next render.

Call `useAgentChatHistoryStore.getState().resetHistory(networkId)` so the `IndexedDB` entry under` agent-chat-history.state.history[networkId]` is gone.


packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -622,6 +635,14 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
                         .setTempNetworks(
                             temporaryNetworks.filter((network) => network.agentInfo.agent_name !== networkToBeDeleted)
                         )
+                    /* istanbul ignore else -- @preserve: the modal cannot open without networkToBeDeleted set */
```
dsargent: IDK what this is about, but it reeks of the LLM taking short cuts. AI coding assistants love to disable stuff rather than fix it. 

> the modal cannot open without networkToBeDeleted set

Then the `if` is serving no purpose and should be _removed_, not just be disabled from code coverage to hit a metric.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -306,14 +307,21 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
 
             const currentTemporaryNetworks = useTempNetworksStore.getState().tempNetworks
 
-            // Remove networks that have been expired for more than GRACE_PERIOD_MS
+            // Networks past the grace period get fully purged: removed from the store AND have their
+            // chat history / sly_data cleared from IndexedDB
+            const reapedNetworks = currentTemporaryNetworks.filter(
```
dsargent: This feels duplicative with the code below. We're scanning the array twice, with the same condition inverted. There has to be a more elegant way.

Note: not a performance thing. These arrays are tiny. Just, feels duplicative and inelegant.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -306,14 +307,21 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
...
+                (n) => n.reservation.expiration_time_in_seconds <= now - GRACE_PERIOD_MS / 1000
+            )
             useTempNetworksStore
                 .getState()
                 .setTempNetworks(
                     currentTemporaryNetworks.filter(
                         (n) => n.reservation.expiration_time_in_seconds > now - GRACE_PERIOD_MS / 1000
                     )
                 )
+            for (const n of reapedNetworks) {
```
dsargent: `n` is a pretty terrible variable name; why not `network`?

If the AI generated this, it's probably getting confused with loop counter variables like `i` and `j` where single letters are "generally considered okay". But it's not a loop counter here.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -622,6 +635,14 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
                         .setTempNetworks(
                             temporaryNetworks.filter((network) => network.agentInfo.agent_name !== networkToBeDeleted)
                         )
+                    /* istanbul ignore else -- @preserve: the modal cannot open without networkToBeDeleted set */
+                    if (networkToBeDeleted) {
+                        useAgentChatHistoryStore.getState().resetHistory(networkToBeDeleted)
+                    }
+                    if (selectedNetwork === networkToBeDeleted) {
+                        setSelectedNetwork(null)
```
swensel: @souravJBcognizant There are 3 instances of 

```
setSelectedNetwork(null)
setAgentCounts(new Map())
```

in this file now. I think making that a common function would be good.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -306,14 +307,21 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
 
             const currentTemporaryNetworks = useTempNetworksStore.getState().tempNetworks
 
-            // Remove networks that have been expired for more than GRACE_PERIOD_MS
+            // Networks past the grace period get fully purged: removed from the store AND have their
+            // chat history / sly_data cleared from IndexedDB
+            const reapedNetworks = currentTemporaryNetworks.filter(
+                (n) => n.reservation.expiration_time_in_seconds <= now - GRACE_PERIOD_MS / 1000
```
swensel: It could be a good idea to make `now - GRACE_PERIOD_MS / 1000` a `const` at the top of this `setInterval`, under `const now = Date.now() / 1000 // convert to seconds since epoch`, since you are doing that in multiple places now.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -622,6 +635,14 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
                         .setTempNetworks(
                             temporaryNetworks.filter((network) => network.agentInfo.agent_name !== networkToBeDeleted)
                         )
+                    /* istanbul ignore else -- @preserve: the modal cannot open without networkToBeDeleted set */
+                    if (networkToBeDeleted) {
```
swensel: Based on the comment `"the modal cannot open without networkToBeDeleted set"`, why does there need to be an `if` block here, then? Also if you need to keep the `if` block, why can't `if (selectedNetwork === networkToBeDeleted) {` be inside of this block?

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -427,6 +435,11 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
                 (network) => network.agentInfo.agent_name !== networkId
             )
             useTempNetworksStore.getState().setTempNetworks(tempNetworksWithoutThisOne)
+            useAgentChatHistoryStore.getState().resetHistory(networkId)
+            if (selectedNetwork === networkId) {
+                setSelectedNetwork(null)
```
dsargent: This is copy pasta from line 642.

More generally, resetting state is getting messy in this module: we now do this (set selected to null/reset counts) in two places, then there's the `resetState` function itself, then there's this in in `onStreamingStarted`:

```
        // Reset agent counts
        setAgentCounts(new Map())

        // Reset newly added temporary networks
        setNewlyAddedTemporaryNetworks(new Set())

        // Reset Agent Network Designer preview
        setAgentsInNetworkDesigner([])
```

It feels rather arbitrary and a bit of a patchwork; each time a bug is discovered, we apply a local patch to fix it for that one case without looking at the bigger picture.

apps/main/__tests__/pages/MultiAgentAcceleratorPage.test.tsx
```diff
@@ -830,6 +861,116 @@ describe("Multi Agent Accelerator Page", () => {
...
+
+            // Active-network deselection should have run (targetAgent becomes null).
+            expect(chatCommonMock).toHaveBeenLastCalledWith(
+                expect.objectContaining({
+                    targetAgent: null,
+                })
+            )
+        })
+
+        it("Should delete an unselected temporary network without touching the active selection", async () => {
```
dsargent, line 860: These vibed tests are very very copy pasted -- a maintenance burden. Please factor out common code cleanly.

Image

dsargent (changes requested):
Please see comments.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
swensel: If you combine the `.filters` then this isn't needed.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
swensel: Similar to what I was saying with https://github.com/cognizant-ai-lab/neuro-san-ui/pull/338/changes#r3263641992

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
swensel: Same thing I was saying with https://github.com/cognizant-ai-lab/neuro-san-ui/pull/338/changes#r3263662834

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
dsargent: I'm not sure you can combine the filters as they want the opposite value (but for the same condition). But there's gotta be a more elegant way than it is now.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
souravJBcognizant: I've removed the if block entirely; since the modal can't open without networkToBeDeleted being set, the null check was never reachable.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
souravJBcognizant: Combined it into one loop that separates networks into `reapedNetworks` and `survivingNetworks`, and moved the cutoff value `now - GRACE_PERIOD_MS / 1000` into a separate variable called `reapCutoff` to make the comparison easier to read.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
souravJBcognizant: Fixed ,renamed the variable to use clearer and more appropriate naming.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
souravJBcognizant: Extracted the duplicated `setSelectedNetwork(...) + setAgentCounts(new Map())` pairing into a single `changeSelectedNetwork` helper and used it at all three call sites

apps/main/__tests__/pages/MultiAgentAcceleratorPage.test.tsx
souravJBcognizant: I have create a common function `seedTemporaryNetworkFromDefaultReservation` to reduce the redundent lines

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
souravJBcognizant: this is similar to the above comment:
Extracted the duplicated `setSelectedNetwork(...) + setAgentCounts(new Map())` pairing into a single `changeSelectedNetwork` helper and used it at all three call sites

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
souravJBcognizant: same as above comment:
removed the if block entirely

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
souravJBcognizant: same as above comment:
Combined it into one loop that separates networks into `reapedNetworks` and `survivingNetworks`, and moved the cutoff value `now - GRACE_PERIOD_MS / 1000` into a separate variable called `reapCutoff` to make the comparison easier to read.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -303,32 +309,39 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
...
-            useTempNetworksStore
-                .getState()
-                .setTempNetworks(
-                    currentTemporaryNetworks.filter(
-                        (n) => n.reservation.expiration_time_in_seconds > now - GRACE_PERIOD_MS / 1000
-                    )
-                )
+            // Networks past the grace period get fully purged: removed from the store AND have their
+            // chat history / sly_data cleared from IndexedDB
+            const reapedNetworks: typeof currentTemporaryNetworks = []
```
dsargent: Could be just `TemporaryNetwork[]` for the type. IMO this is less readable.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -303,32 +309,39 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
...
+            const survivingNetworks: typeof currentTemporaryNetworks = []
+            for (const network of currentTemporaryNetworks) {
+                if (network.reservation.expiration_time_in_seconds <= reapCutoff) {
+                    reapedNetworks.push(network)
```
dsargent: Can just clear the history here as each expired network is discovered, instead of collecting the `reapedNetworks` only to iterate through them again after.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
dsargent: Probably should use `changeSelectedNetwork` here too for consistency.

dsargent (dismissed):
Couple of small comments, but much better than it was. Thank you.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
souravJBcognizant: fixed , use the TemporaryNetwork[] type

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
souravJBcognizant: fixed, removing the chat history of reapedNetwork in one pass

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
souravJBcognizant: fixed, used changeSelectedNetwork

souravJBcognizant:
> Couple of small comments, but much better than it was. Thank you.

I’ve incorporated the suggested changes. Kindly take another look and re-approve the PR if everything appears satisfactory now.

dsargent (approved):

Issue #317, opened 2026-05-13
Show node contents while dragging


dsargent:
Looks better in general, and may fix #315 

Sample fix (to be verified/tested):

```typescript
❯ git diff packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
diff --git i/packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx w/packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
index 0ff13864..6c9dad89 100644
--- i/packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
+++ w/packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
@@ -865,6 +865,10 @@ export const AgentFlow: FC<AgentFlowProps> = ({
                     color: theme.palette.text.primary,
                     fill: theme.palette.text.primary,
                 },
+                "& .react-flow__node.dragging > div": {
+                    opacity: 1,
+                    visibility: "visible",
+                },
             }}
         >
             <ReactFlow
```

Issue #315, opened 2026-05-12
When you try to drag a node in MAUI, it disappears


swensel:
# Description

When you try to drag a node in MAUI, it disappears. Sometimes you have to try to drag it a few times to reproduce the issue. This is once you are done dragging the node, and it does not reappear until you refresh.

There's also a warning in the console: `[React Flow]: It seems that you are trying to drag a node that is not initialized. Please use onNodesChange as explained in the docs. Help: https://reactflow.dev/error#015`

# Before screenshot

<img width="1528" height="815" alt="Image" src="https://github.com/user-attachments/assets/f4523842-bd2d-4ee8-b7d3-0c9e7c57e254" />

# After screenshot

<img width="1570" height="804" alt="Image" src="https://github.com/user-attachments/assets/7950f908-ae1d-4f3a-b589-a24a29f87360" />

Issue #206, opened 2026-03-12
[UN-2935] Neuro-san UI: Any changes user makes when dragging nodes around are lost when chat runs


dsargent:
**Migrated from Jira:** [UN-2935](https://leaf-ai.atlassian.net/browse/UN-2935)

## Jira Details
- **Issue Type:** Task
- **Priority:** Major
- **Status:** Nice to Haves
- **Reporter:** Darren Sargent
- **Assignee:** Sean Wensel
- **Created:** 2025-02-27 16:34:43
- **Updated:** 2025-03-03 14:13:10

PR #339, opened 2026-05-18
Fix node disappearance while dragging

souravJBcognizant:
Issue: #315 #317 

Root cause
`onNodesChange` was only allowing `position`related updates and ignoring all other node changes outside agent designer mode. But React Flow needs every type of change, especially the size/dimension updates that happen when a node first appears — to properly track nodes internally.

Because those dimension updates were being ignored, React Flow didn’t have complete information about the node’s size and bounds. So when the user tried to drag the node, React Flow treated it as invalid/missing, causing the node to disappear from the canvas.

Fix
Forward all change types to `applyNodeChanges` in normal mode. In agent network designer mode (preview-only), continue to suppress position changes so users can't drag nodes around a network they aren't editing, but allow every other change type through.

Before:

`changes.filter((c) => c.type === "position" && !isAgentNetworkDesignerMode)`
After:

`isAgentNetworkDesignerMode ? changes.filter((c) => c.type !== "position") : changes`

similar issue: [link](https://github.com/xyflow/xyflow/issues/4287)


packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -569,14 +569,9 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
-            setNodes((currentNodes) =>
-                applyNodeChanges<RFNode<AgentNodeProps>>(
-                    // For now, we only allow dragging, no updates. In agent network designer mode, doesn't make sense
-                    // to allow position changes since the user isn't actually manipulating a real network
-                    changes.filter((c) => c.type === "position" && !isAgentNetworkDesignerMode),
-                    currentNodes
-                )
-            )
+            // Block node moves in the agent designer preview; pass all changes through otherwise.
+            const allowedChanges = isAgentNetworkDesignerMode ? changes.filter((c) => c.type !== "position") : changes
```
dsargent: Won't this allow deleting nodes as well? We don't want that

swensel: Hi @souravJBcognizant ,

In addition to what Darren said, I don't think we would want to `"block node moves"` (as per the comment) even when `isAgentNetworkDesignerMode` is `true`.

Also, were you able to test this before and after? I can't reproduce the issue, even on `main` now. 

I do still see the warning in the console though (`[React Flow]: It seems that you are trying to drag a node that is not initialized. Please use onNodesChange as explained in the docs. Help: https://reactflow.dev/error#015`):

Image

Being able to reproduce this would be important in determining if it's fixed.

dsargent: To be fair, no, we don't want the user moving nodes around in the AND preview -- that doesn't make sense. (Well, not that it makes much sense to allow the user to drag the nodes around in the regular view either. Maybe they want to move stuff out of the way to better see certain nodes 🤷🏻 There's also #206 but oh well.)

So yes, given our current worldview, blocking the user doing _anything_ with the nodes in the AND preview makes sense.

dsargent: And yes, I was able to vandalize the graph and delete nodes at will with the changes in this PR. I managed to turn "Six Thinking Hats" into "Three thinking hats".

image

👎🏻 I don't think Claude is doing a good job.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
```diff
@@ -1999,4 +2025,55 @@ describe("AgentFlow", () => {
...
+            dimensions: {width: 100, height: 100},
+        }
+        const selectChange: NodeChange<RFNode> = {id: "agent1", type: "select", selected: true}
+
+        beforeEach(() => {
+            capturedReactFlowProps.onNodesChange = null
+            applyNodeChangesSpy.mockClear()
+        })
+
+        it("forwards every change type — including position — to applyNodeChanges in normal mode", () => {
```
dsargent: Asserting incorrect behavior.

dsargent (changes requested):
This fix is incorrect. Please try again.

dsargent:
Also general comment: please put the ticket number in PR titles to help with cross-referencing.

`#123 Fix y2k bug`

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
souravJBcognizant: Hi @dsargent ,

I was following this issue: [4287](https://github.com/xyflow/xyflow/issues/4287)

From the React Flow discussions, it seems that when `onNodesChange` is customized, certain internal updates, especially `dimensions` ([NodeChange Object](https://reactflow.dev/api-reference/types/node-change) ) used for node measurement/initialization, need to pass through. Otherwise React Flow can get out of sync and show warnings like “trying to drag a node that is not initialized”.

My mistake here was using:
`isAgentNetworkDesignerMode? changes.filter((c) => c.type !== "position"): changes`

In my change, instead of forwarding only the internal updates (mostly `dimensions`) needed for synchronization, I forwarded all node change types (`remove`, `select`, `add`, etc.). Because of that, delete actions were also allowed through, which is why nodes could be removed.

Should I try narrowing this down further so that only the required internal updates are forwarded while still blocking position changes in AND preview mode?

dsargent: I don't have time to research the solution in such details -- at that point, I might as well just work on this ticket myself as it would take less time. I just know that forwarding all event types is not correct.

souravJBcognizant: Understood. I’ll apply these changes and spend some time reproducing the node disappearing issue while monitoring the console logs before updating the PR again, so I can verify the fix properly instead of just relying on the discussion thread from React Flow.

swensel:
Hi @souravJBcognizant , what's going on with this PR? Are you still working on it?

souravJBcognizant:
> Hi @souravJBcognizant , what's going on with this PR? Are you still working on it?

Hi @swensel , I am currently blocked on this due to the new Zscaler policy. I’ll test and push the changes once the MAUI API issue is resolved with the DLP team. We had a meeting regarding this today.

dsargent:
Way out of date and superseded by #438.

Issue #329, opened 2026-05-14
Blackduck reports vulnerability with next.js v16.2.1


sidcog:
The latest Blackduck scan reports a vulnerability in the next.js v16.2.1 library. There are four high issues reported, the most egregious was reported on May 12, 2026.

Blackduck suggests updating next.js to v16.2.5 to fix all four issues.

"Next.js is vulnerable to stored cross-site scripting (XSS) due to improper HTML escaping of serialized content in beforeInteractive scripts. This could allow an attacker to execute arbitrary JavaScript code in the victims browser by injecting a maliciously crafted input that breaks out of the intended script context."

PR #345, opened 2026-05-19
#329 Update dependencies to address Next.js and uuid vulnerabilities

dsargent:
Also fix a CodeQL issue regarding regex performance.


swensel (approved):
LGTM. Thanks for doing this.

Issue #216, opened 2026-03-12
[UN-946] Enable React strict mode in UI build


dsargent:
**Migrated from Jira:** [UN-946](https://leaf-ai.atlassian.net/browse/UN-946)

## Jira Details
- **Issue Type:** Task
- **Priority:** Major
- **Status:** Nice to Haves
- **Reporter:** Darren Sargent
- **Assignee:** Darren Sargent
- **Created:** 2023-02-01 12:09:54
- **Updated:** 2025-02-16 20:43:24

PR #351, opened 2026-05-20
#216 Enable React Strict mode

dsargent:
Explanation [here](https://react.dev/reference/react/StrictMode).

Supposedly a very good feature to enable to catch React.js related bugs in development. Note: disappears in prod builds, by design.

I ran the app and a few tests, didn't see any issues. This is insurance against such things creeping in in future.

### DIFF
--- a/apps/main/pages/_app.tsx
+++ b/apps/main/pages/_app.tsx
@@ -26,7 +26,7 @@ import Head from "next/head"
 import {useRouter} from "next/router"
 import {SessionProvider} from "next-auth/react"
 import {SnackbarProvider} from "notistack"
-import {FC, ReactElement, JSX as ReactJSX, ReactNode, useEffect, useState} from "react"
+import {FC, ReactElement, JSX as ReactJSX, ReactNode, StrictMode, useEffect, useState} from "react"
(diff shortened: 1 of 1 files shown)


apps/main/pages/_app.tsx
```diff
@@ -355,20 +355,22 @@ export default function NeuroSanUI({Component, pageProps}: ExtendedAppProps): Re
...
-                <SnackbarProvider
-                    Components={{
-                        info: Snackbar,
-                        success: Snackbar,
-                        warning: Snackbar,
-                        error: Snackbar,
-                    }}
-                />
-            </ThemeProvider>
+            <StrictMode>
```
dsargent, line 358: (Disable whitespace). This is it -- sandwich the whole app in `StrictMode` to enable the checks. Like an extra linter.

swensel (approved):
Approved. I think we might run into some unknown issue, but we can fix it then.

dsargent:
> Approved. I think we might run into some unknown issue, but we can fix it then.

If we do, that's good! It means we had a bug lurking, and this surfaced it.

PR #354, opened 2026-05-21
POC: Editing UI

dsargent:


dsargent:
Was just a POC. @swensel has incorporated the useful elements into #357

Issue #348, opened 2026-05-20
Sometimes an old network name is used when creating a new network for AND


swensel:
# Description

Sometimes an old network name is used when creating a new network for AND

From Babak:

I ran into a bug demoing MAUI just now. I had built an agent network in the morning and it showed up under temporary. When I built a new one a little later, it did not add it to temporary and just highlighted the name of the agent network I had created in the earlier session, but when I clicked on it, the new agent network was loaded.

Issue #319, opened 2026-05-13
Add title to flow (show which network is being used)


dsargent:
Currently flow has no title to show which network is being used:

<img width="762" height="670" alt="Image" src="https://github.com/user-attachments/assets/b198368e-8fdd-4ca0-b0fa-7f445ea58beb" />

Would be good UX to add one:

<img width="766" height="644" alt="Image" src="https://github.com/user-attachments/assets/7a029271-daaf-4a49-b3ee-d2d92ed919ef" />

Patch for that one:

```typescript
diff --git i/packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx w/packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
index eb16bd97..f6b779e0 100644
--- i/packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
+++ w/packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
@@ -84,6 +84,7 @@ export interface AgentFlowProps {
     readonly isAgentNetworkDesignerMode?: boolean
     readonly isStreaming?: boolean
     readonly isSelectedNetworkTemporary?: boolean
+    networkDisplayName?: string
     /** The history key for the currently selected network (used to scope sly_data reads/writes per network). */
     readonly networkId?: string
     readonly neuroSanURL?: string
@@ -195,6 +196,7 @@ export const AgentFlow: FC<AgentFlowProps> = ({
     isAwaitingLlm,
     isStreaming,
     isSelectedNetworkTemporary: isTemporaryNetwork,
+    networkDisplayName,
     networkId,
     neuroSanURL,
     onNetworkReplaced,
@@ -847,6 +849,7 @@ export const AgentFlow: FC<AgentFlowProps> = ({
         <Box
             id={`${id}-outer-box`}
             sx={{
+                position: "relative",
                 height: "100%",
                 width: "100%",
                 backgroundColor: theme.palette.background.default,
@@ -866,6 +869,28 @@ export const AgentFlow: FC<AgentFlowProps> = ({
                 },
             }}
         >
+            <Typography
+                id={`${id}-title-bar`}
+                variant="h6"
+                sx={{
+                    position: "absolute",
+                    top: 8,
+                    left: "50%",
+                    transform: "translateX(-50%)",
+                    zIndex: getZIndex(2, theme),
+                    color: theme.palette.text.primary,
+                    pointerEvents: "none",
[...]

Issue #313, opened 2026-05-12
Make it obvious to users in AND interaction whether they are about to edit an existing network or design a new one


dsargent:
Violates "principle of least surprise" currently. If you design a network with AND, then revisit AND to design a new one, it will always attempt to apply whatever you prompt to "edits" for the existing network, rather than creating a new one.

This is fine and good, but the user may want to design a new network and they don't know they have to clear chat history with AND to do that.

Provide some cue or indication of what will happen. Something like,

`Editing your_previous_network. Enter editing instructions, or clear chat to start fresh.`

Of course, having to clear chat to start a fresh network is a bit of a hack in itself. That could be improved -- "I want to retain context, but I also want to design a new network".

swensel:
One idea that Darren mentioned - add a banner, if on AND, and the user is editing a network already. Until the user clears history on AND, they will be editing a previous network, and the banner will continue to show.

Issue #309, opened 2026-05-11
"Agent Network Designer" window shows stale preview


dsargent:
When vibing networks in series, you get a stale preview of the previously created network before the preview window updates.

PR #357, opened 2026-05-22
#319, #303, #301, #309, #313 - Improve editing experience. Add title.

swensel:
# Description

- #319 - Add title to flow (show which network is being used) and add edit next to that.

- #303 - Allow editing any previous network (topological edits for multiple networks).

Screenshot 2026-05-26 at 10 37 36 PM
Screenshot 2026-05-26 at 10 37 42 PM
Screenshot 2026-05-26 at 10 37 57 PM
Screenshot 2026-05-26 at 10 38 05 PM

- #301 - Also add on hover pencil for nodes.

Also has on hover pencil (see frontman)

Also:

- Disable other user actions during topological edits (and show popup while topological editing is in progress).
- Hide network editor when LLM is in progress.
- Clicking on pencil icons in sidebar should also load the network and enter edit mode.
- Clear chat whenever the user navigates to the Agent Network Designer, so they start fresh each time. This also addresses:
   - #309 - "Agent Network Designer" window shows stale preview
   - #313 - Make it obvious to users in AND interaction whether they are about to edit an existing network or design a new one
   - Also perhaps #348 -Sometimes an old network name is used when creating a new network for AND (we'll have to test that).
 - Trash cans are different color than other action icons (although trash icon should be red on hover).


packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -879,6 +871,16 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
         introduceAgent()
     }, [introduceAgent, resetHistory, targetAgent])
 
+    // Expose the handleStop and handleClearChat methods to parent components via ref for external control
+    useImperativeHandle(
```
swensel, line 687: Moved this `useImperativeHandle` below `handleClearChat` since that function is referenced here.

packages/ui-common/components/MultiAgentAccelerator/TemporaryNetworks.ts
```diff
@@ -70,22 +62,36 @@ const extractNetworkHocon = (message: ChatMessage): string | null => {
...
- * type AGENT_FRAMEWORK only.
- * @return An array of AgentReservation objects if reservations are found, or an empty array if not found or
- * if the message is not of the expected type.
+ * Converts a list of agent reservations received from the backend into TemporaryNetwork objects that can be
+ * displayed in the UI.
+ * @param agentReservations List of "agent reservations" (temporary networks) received from the backend
+ * @param networkHocon Optional network HOCON string associated with the reservations.
+ * @param agentNetworkDefinition Optional agent network definition entries.
+ * @param agentNetworkName Optional backend canonical network name used to match / deduplicate networks.
+ *   When omitted, the name is derived from the reservation_id via {@link extractNetworkNameFromReservationId}.
```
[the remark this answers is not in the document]
swensel: Will remove.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -743,40 +890,242 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
+                        <Typography
+                            variant="overline"
+                            sx={{fontWeight: "bold", letterSpacing: 1, lineHeight: 1.8}}
+                        >
+                            Network Editor
+                        </Typography>
+                        <IconButton
+                            size="small"
+                            aria-label="close edit mode"
+                            onClick={onExitEditMode}
```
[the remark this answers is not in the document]
swensel: I will make an update so that the request is aborted if the user exits out of the Network Editor.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -698,10 +705,17 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
                             id="multi-agent-accelerator-agent-flow"
                             key="multi-agent-accelerator-agent-flow"
                             currentConversations={currentConversations}
+                            currentUser={userInfo.userName}
                             isAwaitingLlm={isAwaitingLlm}
+                            isEditMode={isEditingNetwork}
                             isStreaming={isStreaming}
                             isSelectedNetworkTemporary={isSelectedNetworkTemporary}
```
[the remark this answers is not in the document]
swensel: I'm not sure if we want to reset it once a user goes into that mode.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -348,84 +405,170 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
+                currentTempNetwork?.agentNetworkName,
+                currentUser
+            )
+            applyNetworkSaveResult(dockPrompt, newNetworks, currentTempNetwork?.agentNetworkName)
+            setDockPrompt("")
+        } catch (e: unknown) {
+            const isAbort = e instanceof DOMException && e.name === "AbortError"
+            const isTimeout = e instanceof DOMException && e.name === "TimeoutError"
+            if (!isAbort) {
+                const detail = isTimeout ? "The request timed out. Please try again." : String(e)
```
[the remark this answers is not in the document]
swensel: Updating this.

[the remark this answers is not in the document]
swensel: Pushed an update for this.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
```diff
@@ -173,6 +174,65 @@ describe("AgentFlow", () => {
         return {mockSetThoughtBubbleEdges, getThoughtBubbleEdgesMap: () => map}
     }
 
+    it("Should show the network title when networkDisplayName is provided", async () => {
+        renderAgentFlowComponent({networkDisplayName: "My Network"})
+        expect(await screen.findByText("My Network")).toBeInTheDocument()
+    })
+
+    it("Should show the network title in dark mode", async () => {
```
swensel, line 182: Vibed and just for coverage. Probably not a helpful test, could remove.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
```diff
@@ -1688,4 +1748,439 @@ describe("AgentFlow", () => {
             })
         })
     })
+
+    describe("topology editor dock", () => {
```
swensel, line 1754: New Dock tests.

packages/ui-common/components/MultiAgentAccelerator/Sidebar/AgentNetworkTreeItem.tsx
```diff
@@ -183,21 +186,48 @@ export const AgentNetworkTreeItem: FC<AgentNetworkNodeProps> = ({
                                     <BookmarkIcon sx={{fontSize: "0.75rem", color: "var(--bs-accent1-medium)"}} />
                                 </Tooltip>
                             ) : null}
-                            {isTemporaryNetwork && networkHocon && (
-                                <Tooltip title={isExpired ? "Expired" : "Download network definition"}>
+                        </Box>
+                        {isChild && isTemporaryNetwork && (
```
swensel, line 190: `{isChild && isTemporaryNetwork && (` was below. Moved this up, there are more icons now in the temp network sidebar items.

packages/ui-common/components/MultiAgentAccelerator/Sidebar/AgentNetworkTreeItem.tsx
```diff
@@ -183,21 +186,48 @@ export const AgentNetworkTreeItem: FC<AgentNetworkNodeProps> = ({
...
+                            <Box sx={{display: "flex", alignItems: "center", gap: "0.25rem", marginLeft: "auto"}}>
+                                {networkHocon && (
+                                    <Tooltip title={isExpired ? "Expired" : "Download network definition"}>
```
swensel, line 193: Download network.

packages/ui-common/components/MultiAgentAccelerator/Sidebar/AgentNetworkTreeItem.tsx
```diff
@@ -183,21 +186,48 @@ export const AgentNetworkTreeItem: FC<AgentNetworkNodeProps> = ({
...
+                                                        opacity: 0.3,
+                                                    },
+                                                }}
+                                            >
+                                                <DownloadIcon sx={{fontSize: "0.75rem"}} />
+                                            </IconButton>
+                                        </span>
+                                    </Tooltip>
+                                )}
+                                <Tooltip title="Edit this network">
```
swensel, line 223: Edit network.

packages/ui-common/components/MultiAgentAccelerator/Sidebar/AgentNetworkTreeItem.tsx
```diff
@@ -209,26 +239,24 @@ export const AgentNetworkTreeItem: FC<AgentNetworkNodeProps> = ({
...
-                                        onDeleteNetwork?.(itemId, isExpired)
-                                    }}
-                                    sx={{
-                                        cursor: "pointer",
-                                        fontSize: "1rem",
-                                        "&:hover": {color: theme.palette.warning.main},
-                                    }}
-                                />
-                            </Tooltip>
+                                <Tooltip title="Delete network">
```
swensel, line 246: Delete network.

packages/ui-common/components/MultiAgentAccelerator/Sidebar/Sidebar.tsx
```diff
@@ -259,6 +261,16 @@ export const Sidebar: FC<SidebarProps> = ({
         void fetchVersion()
     }, [customURLLocalStorage, backendNeuroSanApiUrl])
 
+    // When the edit pencil is clicked, select the network first (if not already selected) so the network loads
```
swensel, line 264: This is nice quality of life thing for users. Instead of having to select an network and then click edit, if they click edit in the sidebar, the network is selected, and they enter edit mode.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -103,31 +132,52 @@ const THOUGHT_BUBBLE_TIMEOUT_MS = 10_000
...
-    tempNetworks: T[],
-    networkId: string | undefined
-): T | undefined => (networkId ? tempNetworks.find((n) => n.agentInfo.agent_name === networkId) : undefined)
-
-/** Creates an AbortController with a built-in timeout that fires a TimeoutError. */
-const createTimedAbortController = (timeoutMs: number): {controller: AbortController; cancelTimeout: () => void} => {
-    const controller = new AbortController()
-    const id = setTimeout(() => controller.abort(new DOMException("Save timed out", "TimeoutError")), timeoutMs)
-    return {controller, cancelTimeout: () => clearTimeout(id)}
+const streamNetworkDesignerPrompt = async (
```
swensel, line 142: `streamNetworkDesignerPrompt` is called below in `handleDockApply`.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -348,84 +405,184 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
-        if (!node.data.isEditable) return
-
-        setSelectedAgent({
-            agentId: node.id,
-            agentName: node.data.agentName,
-        })
-        setIsPopupOpen(true)
-    }, [])
+    // Dock (edit-mode prompt bar) state
+    const [dockPrompt, setDockPrompt] = useState<string>("")
```
swensel, line 429: Stores edit mode state.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -348,84 +405,184 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
+    const [isDockStreaming, setIsDockStreaming] = useState<boolean>(false)
+    const dockAbortControllerRef = useRef<AbortController | null>(null)
+
+    const handleNodeClick: NodeMouseHandler<RFNode<AgentNodeProps>> = useCallback(
```
swensel, line 433: `handleNodeClick` is also on the left, it's just moved down a bit.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -348,84 +405,184 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
-                await onSaveAgent(agentName, updated, currentAgentNetworkName, controller.signal)
-            } catch (e) {
-                // onSaveAgent is responsible for user-facing error handling and notification.
-                // We catch here only to ensure the finally block (and popup close) always runs.
-                console.error(`Error saving agent ${agentName}. See onSaveAgent implementation for details.`, e)
-            } finally {
-                cancelTimeout()
-                saveAbortControllerRef.current = null
+    /** Applies the networks returned by the designer: upserts them and triggers navigation if needed. */
+    const applyNetworkSaveResult = useCallback(
```
swensel, line 467: Used for Network Editor saves (topological edits).

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -348,84 +405,184 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
+                    `Failed to update agent "${agentName}".`,
+                    "A reservation was returned but did not match the current network. Please try again."
+                )
             }
         },
-        [onSaveAgent]
+        [networkId, onNetworkReplaced]
     )
 
+    const handleDockApply = useCallback(async () => {
```
swensel, line 497: Apply the changes in the Dock (Network Editor / Topological Edits).

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -348,84 +405,184 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
+        const currentDefinition = currentTempNetwork?.agentNetworkDefinition ?? []
+
+        setIsDockStreaming(true)
+        const controller = new AbortController()
+        dockAbortControllerRef.current = controller
+        let hasTimedOut = false
+        const timeoutId = setTimeout(() => {
+            hasTimedOut = true
+            controller.abort()
+        }, 120_000) // 2 min timeout
```
swensel, line 512: 2 min timeout. We've talked about moving to something like TanStack Query.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -348,84 +405,184 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
+        try {
+            const newNetworks = await streamNetworkDesignerPrompt(
+                neuroSanURL,
+                controller.signal,
+                dockPrompt,
+                currentDefinition,
+                currentTempNetwork?.agentNetworkName,
+                currentUser
+            )
+            applyNetworkSaveResult(dockPrompt, newNetworks, currentTempNetwork?.agentNetworkName)
```
swensel, line 522: Apply save.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -348,84 +405,184 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
+            setDockPrompt("")
+        } catch (e: unknown) {
+            const isAbort = e instanceof DOMException && e.name === "AbortError"
```
swensel, line 525: Error handling.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -348,84 +405,184 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
+                )
+            }
+        } finally {
+            clearTimeout(timeoutId)
+            dockAbortControllerRef.current = null
+            setIsDockStreaming(false)
+        }
+    }, [applyNetworkSaveResult, currentUser, dockPrompt, networkId, neuroSanURL, tempNetworks])
+
+    const handleExitEditMode = useCallback(() => {
```
swensel, line 544: This is the escape hatch for users if they enter topological editing and want to get out (it does take a while). I considered not letting them do this but that felt wrong.

We could have a confirmation such as "Are you sure you want to abort this request?". That might be a good idea.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -348,84 +405,184 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
+        if (isDockStreaming) {
+            dockAbortControllerRef.current?.abort()
+            dockAbortControllerRef.current = null
+            setIsDockStreaming(false)
+        }
+        onExitEditMode?.()
+    }, [isDockStreaming, onExitEditMode])
+
     const handlePopupSave = useCallback(
```
swensel, line 553: Saves for manual (node) edits.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -348,84 +405,184 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
+                setIsPopupOpen(false)
+                return
+            }
+
+            setIsSavingAgent(true)
+            const saveController = new AbortController()
+            saveAbortControllerRef.current = saveController
+            const saveTimeoutId = setTimeout(
+                () => saveController.abort(new DOMException("Save timed out", "TimeoutError")),
+                60_000
```
swensel: These requests are faster than topological edits. 1 min timeout here. There wasn't a timeout here before.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -743,40 +904,242 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
+                            transform: "translateX(-50%)",
+                            zIndex: getZIndex(1, theme),
+                            display: "flex",
+                            alignItems: "center",
+                            gap: 1,
+                            pointerEvents: "none",
+                        }}
+                    >
+                        <Typography
+                            id={`${id}-network-title`}
```
swensel, line 945: Add network title.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -743,40 +904,242 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
+                                variant="contained"
+                                size="small"
+                                onClick={onEnterEditMode}
+                                startIcon={<EditIcon />}
+                                sx={{
+                                    pointerEvents: "auto",
+                                    "&:hover": {backgroundColor: theme.palette.primary.main},
+                                }}
+                            >
+                                Edit
```
swensel, line 977: Edit button next to network title, to enter edit more (shows Network Editor dock).

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -743,40 +904,242 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
+                    onNodeClick={handleNodeClick}
+                    fitView={true}
+                    nodeTypes={nodeTypes}
+                    edgeTypes={edgeTypes}
+                    connectionMode={ConnectionMode.Loose}
+                >
+                    {!isAwaitingLlm && (
+                        <>
+                            {agentsInNetwork?.length && !isAgentNetworkDesignerMode && !isEditMode ? getLegend() : null}
+                            <Background id={`${id}-background`} />
```
swensel, line 996: This logic is also on the left. It was just moved down.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -743,40 +904,242 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
+                            {!isAgentNetworkDesignerMode && !isEditMode && getControls()}
+                            {shouldShowRadialGuides ? getRadialGuides() : null}
+                        </>
+                    )}
+                </ReactFlow>
+                {isDockStreaming && (
+                    <Box
+                        id={`${id}-dock-applying-overlay`}
```
swensel, line 1004: This is an overlay shown when the user is applying topological changes. The user is disabled from taking other actions unless they use the [escape hatch](https://github.com/cognizant-ai-lab/neuro-san-ui/pull/357/changes#r3315468742).

There were issues when trying to do topological changes and manual edits at the same time.

It's also not clear if multiple topological edits to different networks at the same time will work in the same browser tab.

We can probably revise this in the future.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -743,40 +904,242 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
+                    nodes={nodes}
+                    edges={thoughtBubbleEdgesForOverlay}
+                    showThoughtBubbles={showThoughtBubbles}
+                    isStreaming={isStreaming}
+                    onBubbleHoverChange={handleBubbleHoverChange}
+                />
+            </Box>
+            {isEditMode && isTemporaryNetwork && !isAwaitingLlm && (
+                <Box
+                    id={`${id}-topology-editor-dock`}
```
swensel, line 1063: Network Editor dock (for applying topological changes).

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -743,40 +904,242 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
+                        variant="caption"
+                        sx={{
+                            px: 2,
+                            pt: 0.5,
+                            pb: 0,
+                            color: theme.palette.text.secondary,
+                            display: "block",
+                        }}
+                    >
+                        Changes apply only to this Temporary network
```
swensel, line 1105: Since changes take a long time, I wonder if I should also add a note that it will take a few minutes, but we may improve performance in the future. Babak has also said that with reasoning models, users may already also be used to waiting a few minutes for requests to complete.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -743,40 +904,242 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
+                            startIcon={
+                                isDockStreaming ? (
+                                    <CircularProgress
+                                        size={16}
+                                        color="inherit"
+                                    />
+                                ) : undefined
+                            }
+                        >
+                            {isDockStreaming ? "Applying…" : "Apply"}
```
swensel, line 1149: Different states for Apply or "Applying...".

packages/ui-common/components/MultiAgentAccelerator/AgentNode.tsx
```diff
@@ -205,13 +207,19 @@ export const AgentNode: FC<NodeProps<RFNode<AgentNodeProps>>> = (props: NodeProp
         ? theme.palette.common.black
         : theme.palette.common.white
 
+    const isClickableNode = isEditable
```
swensel, line 210: This is kind of useless. I could remove it.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -255,6 +257,13 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
     // Reference to the ChatCommon component to allow external stop button to call its handleStop method
     const chatRef = useRef<ChatCommonHandle | null>(null)
 
+    // Clear chat whenever the user navigates to the Agent Network Designer, so they start fresh each time
```
swensel, line 260: Clear chat whenever the user navigates to the Agent Network Designer, so they start fresh each time.

This fixes:

- #309 - "Agent Network Designer" window shows stale preview
- #313 - Make it obvious to users in AND interaction whether they are about to edit an existing network or design a new one
- Also perhaps #348 -Sometimes an old network name is used when creating a new network for AND (we'll have to test that).

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -600,6 +609,10 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
         }
     }
 
+    const handleEditNetwork = (_networkId: string) => {
```
swensel, line 612: Various changes for entering edit mode.

packages/ui-common/components/MultiAgentAccelerator/TemporaryNetworks.ts
```diff
@@ -29,28 +29,20 @@ export type AgentReservation = {
...
-        reservation,
-        agentInfo: {
-            agent_name: `${TEMPORARY_NETWORK_FOLDER}/${reservation.reservation_id}`,
-        },
-        // agentNetworkName is the network name "prefix", with the UUID-stripped
-        agentNetworkName: removeTrailingUuid(reservation.reservation_id),
-        networkHocon,
-        agentNetworkDefinition,
-    }))
+export const extractReservations = (message: ChatMessage): AgentReservation[] => {
```
swensel, line 38: Moved up. Was already present in this file.

packages/ui-common/components/MultiAgentAccelerator/TemporaryNetworks.ts
```diff
@@ -70,22 +62,35 @@ const extractNetworkHocon = (message: ChatMessage): string | null => {
...
  */
-const extractReservations = (message: ChatMessage): AgentReservation[] => {
-    // Check for temp networks ("reservations") in sly_data
-    if (message?.type === ChatMessageType.AGENT_FRAMEWORK && message?.sly_data?.[AGENT_RESERVATIONS_KEY]) {
-        return message.sly_data[AGENT_RESERVATIONS_KEY] as AgentReservation[]
-    } else {
-        // Not the type of message that would contain reservations, or no reservations found, return empty array
-        return []
-    }
+export const convertReservationsToNetworks = (
```
swensel, line 73: I think an LLM moved this. Maybe I should just move it back up.

packages/ui-common/components/MultiAgentAccelerator/TemporaryNetworks.ts
```diff
@@ -70,22 +62,35 @@ const extractNetworkHocon = (message: ChatMessage): string | null => {
...
+        },
+        // Use the explicit name when provided; fall back to extracting it from the reservation_id so that
+        // networks are always deduplicated by name even when the backend omits AGENT_NETWORK_NAME_KEY.
+        agentNetworkName: agentNetworkName ?? removeTrailingUuid(reservation.reservation_id),
+        networkHocon,
+        agentNetworkDefinition,
+    }))
 }
 
+export const isEditableAgent = (displayAs: string | undefined): boolean => displayAs === DisplayAs.LLM_AGENT
```
swensel, line 92: I don't get why this was moved either 🤷

jest.config.ts
```diff
@@ -74,10 +74,10 @@ const config: Config.InitialOptions = {
 
     coverageThreshold: {
         global: {
-            statements: -120,
-            branches: -161,
-            functions: -29,
-            lines: -94,
+            statements: -128,
```
swensel: Not too bad. Little bit of a hit.

dsargent:
First general point: that title is a beast. I understand (and applaud) the urge to be descriptive and include "all relevant information", but that's not what titles are for. That info belongs in the _description_ (where you also have it). Titles should be short, snappy, high level. Think: headline, release notes. For this PR, something like, 

`<list of ticket numbers> Improve editing experience. Add Title.`

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
dsargent: We are now doubling down on the ugly `imperativeHandle` pattern (which is my original sin). Feels anti-React. Is there a better way?

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
dsargent: Setting expectations is always good. "This may take up to 5 minutes" (or some reasonable value matching what you've observed). We eventually may want to move to a batch model, where you save your edit then get some kind of notification when it's done so you can go do other things.

In my dream system we'd be collecting metrics on things like how long edits take, then we could tell the user, "The average time for edits similar to yours is: _whatever_ minutes". Think how much better the UX is when you call a 1-800 number and the recording says, "The average wait time is _8_ minutes" vs. "We are experiencing higher than usual call volume. Please be patient."

> Changes apply only to this Temporary network

I think this is just noise and more confusing to the user than anything. "Why would it affect any other network than the one I'm looking at?!" I think this disclaimer has more to do with our various discussions and development iterations & history than something that would be meaningful to the user, since we used to have issues with "which network am I editing?" I recommend removing it.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
dsargent: Okay, but why roll your own when there's [MUI Backdrop](https://mui.com/material-ui/react-backdrop/) designed exactly for this scenario?

packages/ui-common/components/MultiAgentAccelerator/Sidebar/AgentNetworkTreeItem.tsx
dsargent: ? nothing new here ?

packages/ui-common/components/MultiAgentAccelerator/Sidebar/AgentNetworkTreeItem.tsx
dsargent, line 237: Same color for disabled look vs regular, so no need to specify twice.

Also, consider using a styled component for the increasing number of "action buttons". At some point we may need to go with a hotdog menu to avoid the profliferation of icons. I think 3 is about the limit.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -103,31 +132,52 @@ const THOUGHT_BUBBLE_TIMEOUT_MS = 10_000
...
+    let newNetworks: TemporaryNetwork[] = []
+
+    await sendChatQuery(
+        neuroSanURL,
+        signal,
+        userPrompt,
+        AGENT_NETWORK_DESIGNER_ID,
+        (chunk: string) => {
+            const chatMessage = chatMessageFromChunk(chunk)
+            if (!chatMessage) return
```
dsargent, line 159: LLMs love to not put braces for one-liners, but I'm not keen on that style. It's inconsistent with non-one-liners, and makes it harder to add items to the `if` condition. I want a prettier/linter rule for this but didn't look into one yet.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -103,31 +132,52 @@ const THOUGHT_BUBBLE_TIMEOUT_MS = 10_000
...
+
+            const reservations = extractReservations(chatMessage)
+            if (reservations.length === 0) return
+
+            const networkHocon = extractNetworkHocon(chatMessage)
+            const agentNetworkNameFromMessage = chatMessage.sly_data?.[AGENT_NETWORK_NAME_KEY] as string | undefined
```
dsargent, line 165: I'm pondering what to do about our "blind faith" casts, like this one (but we do it in a lot of places). Because we don't know until runtime what we'll get from the API, we can't rely on `tsc`, so we have to do ugly `as string` type "trust me, bro" casts.

A better but artisanal way would be checking `typeof <thing> === "string"` and responding appropriately if not. Ultimately something like `zod` is probably the answer.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -348,84 +405,184 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
+            if (!isTemporaryNetwork) return
+
+            // Only llm_agent nodes support instructions/description editing.
+            if (!isEditableAgent(node.data.displayAs)) return
+
+            // Find the clicked agent's existing instructions and description from the temp network definition.
+            const currentTempNetwork = networkId
+                ? tempNetworks.find((n) => n.agentInfo.agent_name === networkId)
+                : undefined
+            const found = (currentTempNetwork?.agentNetworkDefinition ?? []).find((e) => e.origin === node.id)
```
dsargent, line 445: `found` feels like a super literal LLM suggestion. Poor name. Something like `currentNetworkDefinition` to align with `currentTempNetwork`.

Also the null object pattern `?? []` is redundant.

```typescript
            const found = (currentTempNetwork?.agentNetworkDefinition?.find((e) => e.origin === node.id)
```

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -348,84 +405,184 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
-                cancelTimeout()
-                saveAbortControllerRef.current = null
+    /** Applies the networks returned by the designer: upserts them and triggers navigation if needed. */
+    const applyNetworkSaveResult = useCallback(
+        (agentName: string, newNetworksFromSave: TemporaryNetwork[], currentAgentNetworkName: string | undefined) => {
+            if (newNetworksFromSave.length === 0) {
+                sendNotification(
+                    NotificationType.error,
+                    `Failed to update agent "${agentName}".`,
+                    "The network designer did not return a reservation. Please try again."
```
dsargent, line 473: Again, this message is too technical. Users: "what's a 'reservation'?!". Please try to view things from the user's point of view, and not a technical user like one of our engineers.

Probably the best compromise is:

- nice, fluffy, non-technical-user friendly message in the popup.
- `console.error` for the gory technical details.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -348,84 +405,184 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
+                )
+                return
+            }
+
+            const replacement = newNetworksFromSave.find((n) => n.agentNetworkName === currentAgentNetworkName)
```
dsargent, line 478: This block needs a comment.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -348,84 +405,184 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
+                if (networkId && onNetworkReplaced) {
+                    useAgentChatHistoryStore.getState().copyHistory(networkId, replacement.agentInfo.agent_name)
+                    onNetworkReplaced(networkId, replacement.agentInfo.agent_name)
+                }
+            } else {
+                // Reservations came back but none matched the current network — surface this to the user.
+                sendNotification(
+                    NotificationType.error,
+                    `Failed to update agent "${agentName}".`,
+                    "A reservation was returned but did not match the current network. Please try again."
```
dsargent, line 490: Again, this is incomprehensible to the average user. It's very much written from a dev's point of view. Think of that site I sent where the guy has his mom test UI's. What would she make of this message? What would be a better one written from the _user's_ perspective?

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -348,84 +405,184 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
+                )
             }
         },
-        [onSaveAgent]
+        [networkId, onNetworkReplaced]
     )
 
+    const handleDockApply = useCallback(async () => {
+        if (!dockPrompt.trim() || !neuroSanURL || !currentUser) return
```
dsargent, line 498: This is a prime case for either a comment, or better, an "introduce explanatory variable" refactoring.

```typescript
const readyToApplyEdit = dockPrompt.trim() && neuroSanURL && currentUser
if (!readyToApplyEdit) {
  return
}
```

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
dsargent: Kudos for underscore.

I'd make this a module-level `const`. Also, lose the LLM over-explaining comment that hardcodes the value. Those inevitably get forgotten to be changed when the value itself changes, leading to confusion.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -348,84 +405,184 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
+        setIsDockStreaming(true)
+        const controller = new AbortController()
+        dockAbortControllerRef.current = controller
+        let hasTimedOut = false
+        const timeoutId = setTimeout(() => {
+            hasTimedOut = true
+            controller.abort()
+        }, 120_000) // 2 min timeout
+        try {
+            const newNetworks = await streamNetworkDesignerPrompt(
```
dsargent, line 514: This method is poorly named. It describes the implementation, not what it's _for_. Something like, `applyNetworkDesignerChanges`. The "prompt" is just an input -- that's part of the _signature_. "Stream" is an implementation detail. 

It would be like having a function that uses a standard library to sort an array with some custom frills, and calling it `passArrayToStandardLibraryFunction` rather than, `sort`.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
dsargent: Or just, `saveUpdates` or something like that?

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
dsargent: Definitely +1 on the confirmation.

Also, users would expect the Escape key to trigger this too.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -348,84 +405,184 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
-                instructionsText,
-                descriptionText
+            // Find the temp network entry for the currently selected network.
+            const currentTempNetwork = networkId
+                ? tempNetworks.find((n) => n.agentInfo.agent_name === networkId)
+                : undefined
+
+            // Produce a new array with the saved agent's fields updated; all other entries pass through unchanged.
+            const currentDefinitions = currentTempNetwork?.agentNetworkDefinition ?? []
+            const updated = currentDefinitions.map((entry) =>
```
dsargent, line 564: Again, `updated` -- not a great variable name. Guessing vibed.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
dsargent: Yes, I'm all for the timeouts. Please make them module-level consts so we can see & adjust them easily.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -348,84 +405,184 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
+            const saveController = new AbortController()
+            saveAbortControllerRef.current = saveController
+            const saveTimeoutId = setTimeout(
+                () => saveController.abort(new DOMException("Save timed out", "TimeoutError")),
+                60_000
+            )
+            try {
+                await onSaveAgent(agentName, updated, currentTempNetwork?.agentNetworkName, saveController.signal)
+            } catch (e) {
+                console.error(`Error saving agent ${agentName}. See onSaveAgent implementation for details.`, e)
```
dsargent: Should we not notify the user here? Toast?

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
dsargent: Ha, just realized, when we have timeouts we can actually tell the user how long they can expect to wait. Because if it takes longer than that, we'll abort.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
dsargent: Thanks for adding the title. I feel it could be more prominent, perhaps a border or some kind of box shadow? Not sure.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
dsargent: The edit button is a good addition, but I agree with Vince -- not the best place for it. Maybe next to the legend? Also, could/should it be consistent in appearance with the other "action buttons"?

Image

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
dsargent: > It's also not clear if multiple topological edits to different networks at the same time will work in the same browser tab.

But that's not possible with this PR, due to the overlay, right?

Though, Babak has mentioned over and over that as a user, he would expect to be able to interact with multiple networks in parallel. As in, kick off some slow operation with a network, then click a different one and continue working while that cooks. I suggested he try multiple tabs. Related ticket: #341

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -743,40 +904,242 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
                 <AgentNodePopup
                     agentName={selectedAgent.agentName}
-                    initialInstructions={popupEntry?.instructions ?? ""}
-                    initialDescription={popupEntry?.description ?? ""}
+                    initialInstructions={selectedAgent.initialInstructions}
+                    initialDescription={selectedAgent.initialDescription}
                     isOpen={isPopupOpen}
-                    isSaving={isSaving}
+                    isSaving={isSavingAgent}
                     onClose={handlePopupClose}
```
dsargent, line 1161: We should resolve the `on` vs `handle` naming for event handlers. Here we're mixing both.

Or maybe our standard is the event itself is `onFoo` and the associated handler is `handleFoo`? Either way, we should be consistent.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -743,40 +904,242 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
+                                id={`${id}-dock-applying-spinner`}
+                                size={24}
+                            />
+                            <Box>
+                                <Typography
+                                    id={`${id}-dock-applying-title`}
+                                    variant="body1"
+                                    fontWeight="bold"
+                                >
+                                    Applying changes to network
```
dsargent, line 1037: Perhaps rather than just a `Box`, this should be a full-blown `Modal` with the "cancel" button in it. As a user, that's where I would expect to find "cancel".

Also, the UX is a bit confusing. The "Applying changes" box pops up front and center, as if it's a modal (meaning, no other interactions until this is done) but the rest of the UI is live and usable. That's a mixed message and not consistent with other UIs.

packages/ui-common/components/MultiAgentAccelerator/AgentNode.tsx
dsargent: Agreed.

packages/ui-common/components/MultiAgentAccelerator/AgentNode.tsx
```diff
@@ -205,13 +207,19 @@ export const AgentNode: FC<NodeProps<RFNode<AgentNodeProps>>> = (props: NodeProp
...
+    const [isHovered, setIsHovered] = useState(false)
+
     return (
         <>
             <AnimatedNode
                 id={agentId}
                 data-testid={agentId}
                 glowColor={glowColor}
                 isActive={isActiveAgent}
+                onMouseEnter={() => setIsHovered(true)}
```
dsargent, line 221: Usually tracking hover states manually like this is not required. (Vibed?)

Simpler using CSS states like ` "&:hover"` as is done elsewhere in this component.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
dsargent: As in our discussion in Slack, I'm not keen on this because it (1) is bespoke logic hardcoded for a particular network (which yes, we have elsewhere, but that doesn't mean we have to double down) and (2) it means AND behaves differently than all other networks regarding context and history.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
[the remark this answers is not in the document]
dsargent: Copilot doesn't realize these two are in lockstep. We can only get into editing mode for temporary networks, though if that changes some day (🙏🏻) this code will fail.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -939,6 +960,7 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
             {Tour}
             {getTourModal()}
             {getProgressPopper()}
+
```
dsargent, line 963: No need for a blank line here

packages/ui-common/components/MultiAgentAccelerator/TemporaryNetworks.ts
dsargent: Unless there's a good reason, please move things back to minimize diff noise before requesting reviews. It's kinder to reviewers.

jest.config.ts
dsargent: Those little bits are adding up. Please address #287 as the next ticket.

packages/ui-common/components/MultiAgentAccelerator/Sidebar/Sidebar.tsx
dsargent: Agreed, as a user this is exactly what I would expect to happen.

packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
```diff
@@ -623,6 +623,27 @@ describe("ChatCommon", () => {
...
+
+        // Call it and wait for all async state updates to settle
+        await act(async () => {
+            ref.current?.handleClearChat()
+        })
+
+        // After clearing, connectivity info (rendered into chatOutput via updateOutput) is gone
+        // We wait for sample queries to re-appear (they're in agentSampleQueries state, not chatOutput,
+        // so they stay after clear and confirm the component is still functional)
+        await screen.findByText("Sample query 1")
```
dsargent, line 653: This is a rather weak assertion. It would still pass even the chat wasn't cleared, as long as the sample query displays. Should assert that the actual chat that was there before...no longer is.

Also, rather than hardcoding the query text, better to use `MOCK_CONNECTIVITY_INFO.metadata.sample_queries[0]` or whatever you want to assert on. That's why we have the "data-driven tests" thing.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
dsargent: I've started removing all tests for MUI theme stuff _unless_ the component is hand-crafting something related to the theme. For example, if the component checks `if (darkMode)`, that gets a test. If the component just chills and lets MUI handle things, theme tests are superfluous. We're not here to test MUI.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
```diff
@@ -173,6 +174,65 @@ describe("AgentFlow", () => {
...
+    })
+
+    it("Should show the network title in dark mode", async () => {
+        renderAgentFlowComponent({networkDisplayName: "Dark Network"}, "dark")
+        expect(await screen.findByText("Dark Network")).toBeInTheDocument()
+    })
+
+    it("Should not show the title bar when networkDisplayName is not provided", async () => {
+        const {container} = renderAgentFlowComponent()
+        expect(container.querySelector("#test-flow-id-network-title-bar")).not.toBeInTheDocument()
```
dsargent, line 189: Better to use `screen.queryByText`. DOM ID queries are last resort.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
```diff
@@ -173,6 +174,65 @@ describe("AgentFlow", () => {
         return {mockSetThoughtBubbleEdges, getThoughtBubbleEdgesMap: () => map}
     }
 
+    it("Should show the network title when networkDisplayName is provided", async () => {
+        renderAgentFlowComponent({networkDisplayName: "My Network"})
```
dsargent, line 178: In general, please use single point of truth for test data as it makes maintenance easier.

```typescript
const networkName = "My Network"
...
```

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
dsargent, line 1: General comment, this component is turning into a monolith and getting hard to modify and reason about. We collectively have shoveled a lot of different concerns in here, and they probaly should be teased out:

1) Displaying the flow itself
2) Thought bubble logic
3) Now, editing logic and AND interaction

Sometimes the _easiest_ place to add new code is not the _right_ place.

I don't know where the natural fracture lines are, but please ponder how we could componentize this in a cleaner way (cf. #273)

packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
```diff
@@ -623,6 +623,27 @@ describe("ChatCommon", () => {
...
 
+    it("Should clear the chat when handleClearChat is called via ref", async () => {
+        const ref = createRef<ChatCommonHandle>()
+        renderChatCommonComponent({ref})
+
+        // Wait for the component to initialize so the ref is set up
+        await screen.findByRole("button", {name: "Send"})
+
+        // Verify handleClearChat is exposed via the imperative ref
+        expect(typeof ref.current?.handleClearChat).toBe("function")
```
dsargent, line 643: This is a strange assertion. What else would it be? Is this vibed paranoia by LLM?

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
```diff
@@ -173,6 +174,65 @@ describe("AgentFlow", () => {
...
+            isEditMode: false,
+            isAwaitingLlm: false,
+            onEnterEditMode,
+        })
+        const editBtn = await screen.findByRole("button", {name: "Edit"})
+        await user.click(editBtn)
+        expect(onEnterEditMode).toHaveBeenCalledTimes(1)
+    })
+
+    it("Should not show the Edit button for a non-temporary network", async () => {
```
dsargent, line 206: Or to avoid the double-negatives soup, "It should hide the Edit button for permanent networks".

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
```diff
@@ -1688,4 +1748,480 @@ describe("AgentFlow", () => {
...
+        })
+
+        it("does not show the dock when isEditMode is false", () => {
+            const {container} = renderAgentFlowComponent({
+                isEditMode: false,
+                isSelectedNetworkTemporary: true,
+                networkId: DOCK_NETWORK_ID,
+            })
+
+            expect(container.querySelector("#test-flow-id-topology-editor-dock")).not.toBeInTheDocument()
```
dsargent, line 1805: This test would probably pass anyway since you don't have any "wait" calls to allow the rendering to complete. If you're curious, try it: deliberately modify the code under test to always show the dock. I bet this test will still pass.

Negative tests are very very prone to false negatives (unsurprisingly). Usually you want a pattern like,

1) Render <thing>
2) Assert something positive (with `wait` so we know rendering and microtasks have completed)
3) Only _then_ assert the negatives -- `not.toBeInTheDocument()` etc.

I'm so paranoid about this that when writing negative tests, I first modify the code under test to make it fail so that `not.toBeInTheDocument` blows up, and only then fix it, basically a micro-level test-driven development.

dsargent: Also, again `querySelector` is a last resort compared to `queryByRole` (best) or `queryByText` (good).

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
```diff
@@ -1688,4 +1748,480 @@ describe("AgentFlow", () => {
...
+            await user.click(screen.getByRole("button", {name: /apply/iu}))
+
+            await waitFor(() => {
+                expect(enqueueSnackbar).toHaveBeenCalledWith(
+                    expect.anything(),
+                    expect.objectContaining({variant: "error"})
+                )
+            })
+            // Button should re-enable after error
+            expect(screen.getByRole("button", {name: /apply/iu})).toBeEnabled()
```
dsargent, line 2006: LLMs love the pointless `/i`.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
```diff
@@ -1688,4 +1748,480 @@ describe("AgentFlow", () => {
...
+                isEditMode: true,
+                isSelectedNetworkTemporary: true,
+                networkId: DOCK_NETWORK_ID,
+                neuroSanURL: "http://localhost:8080",
+                currentUser: "test-user",
+            })
+
+            // Use fireEvent (synchronous) so fake timers work from the start
+            const promptField = screen.getByPlaceholderText(/describe a change/iu)
+            fireEvent.change(promptField, {target: {value: "add a node"}})
```
dsargent, line 2031: Why not `user.foo()` events? If it's because of fake timers, you can create a user event that works with those too:

```typescript
        const localUser = userEvent.setup({advanceTimers: jest.advanceTimersByTime.bind(jest)})
```

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
```diff
@@ -1688,4 +1748,480 @@ describe("AgentFlow", () => {
...
+
+            // Advance past the 120-second dock apply timeout
+            await act(async () => {
+                jest.advanceTimersByTime(121_000)
+            })
+
+            jest.useRealTimers()
+
+            await waitFor(() => {
+                const timedOutCall = (enqueueSnackbar as jest.Mock).mock.calls.find((c: unknown[]) =>
```
dsargent, line 2042: perhaps cleaner and more idiomatic (IMO):

```typescript
expect(
    (enqueueSnackbar as jest.Mock).mock.calls.some((c: unknown[]) => JSON.stringify(c).includes("timed out"))
).toBeTruthy()
```

dsargent (dismissed):
This is a lot of work, especially the tests. Thanks for completing this.

I tried out the editing UI locally and it's an improvement over what we have. More intuitive and obvious.

I had a lot of comments on style and design, and I'm still uncomfortable with the "clear context on selecting AND" design.

Approving for progress. Please take note of my comments.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
dsargent: Even the separation of "us, the UI" vs "AND" is artificial and not something the user knows or cares about. That's just an architectural distinction [within our team](https://en.wikipedia.org/wiki/Conway's_law). To the user, it's all just "the app" -- MAUI, Node.js, Neuro-san, AND. They don't know or care about our architecture so the errors have to make sense to _them_.

Of course I'm talking general end users here, not highly technical ones like our own dev team.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
dsargent: This could (probably should) be an independent component.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
```diff
@@ -1716,12 +1717,13 @@ describe("AgentFlow", () => {
...
             await user.click(screen.getByRole("button", {name: "Save"}))
 
-            // Popup closes immediately on save; onSaveAgent error is handled silently in the background
+            // Popup closes immediately on save; onSaveAgent error is surfaced via notification
             await waitFor(() => {
                 expect(screen.queryByRole("button", {name: /applying changes/iu})).not.toBeInTheDocument()
             })
 
             consoleErrorSpy.mockRestore()
+            consoleDebugSpy.mockRestore()
```
dsargent, line 1726: Not needed due to `withStrictMocks`.

packages/ui-common/components/Common/notification.tsx
```diff
@@ -51,24 +51,29 @@ export const sendNotification = (
...
-        case NotificationType.success:
-            duration = SUCCESS_NOTIFICATION_DURATION_MS
-            break
-        case NotificationType.warning:
-        case NotificationType.error:
-        default:
-            duration = ERROR_WARNING_NOTIFICATION_DURATION_MS
-            break
+    let duration: number | null
+    if (durationOverride !== undefined) {
```
dsargent, line 63: It's typed as `number | null` but here checked against `undefined`?! That may work...but it's clunky. Everywhere else we just do "truthiness" checks -- `if (foo) ...`

dsargent (approved):

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
swensel: We lose `1` in terms of branches, if this test is removed.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
swensel: There's been some more updates made to this already.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
dsargent: That would be this part:

> _unless_ the component is hand-crafting something related to the theme

Issue #353, opened 2026-05-21
Agent intro double renders in chat window (Strict mode violation)


dsargent:
Now that #351 is merged, we are starting to see offenders.

The first one is as described in the title. React.js strict mode is exposing the issue where the rendering/effect is happening twice. See screenshot.

<img width="598" height="313" alt="Image" src="https://github.com/user-attachments/assets/300e93b3-53ca-4612-a480-ca8886a5598d" />

Issue #277, opened 2026-05-01
Regular text output from agents being mis-detected as JSON


dsargent:
Due to the history with the Neuro-san API, which in some cases would "smuggle" JSON in the `text` field, we have "JSON detection" logic in `ChatCommon` that is a bit over-eager.

<img width="568" height="430" alt="Image" src="https://github.com/user-attachments/assets/e2bdf644-194b-43a2-8917-9d86a0f9f65c" />

Note the square brackets around the output; it's being formatted/repaired as JSON, even though it's just text.

Basically it tries to parse the text as JSON, and if that succeeds, assumes it _is_ JSON. But a lot of things parse as JSON that aren't intended to be JSON.

For this ticket, investigate that logic in `ChatCommon`, determine the right thing to do so we don't get false positives on non-JSON content. This is not trivial; we have to come up with an "always-never" solution: we always detect JSON as JSON, and _never_ detect plain strings as JSON.

It's possible that we shouldn't even be looking for JSON in the `text` stream at this point, since Neuro-san now has the `structure` field for that.

swensel:
It's also pretty ugly in dark mode when this happens. Worth revising how the UI handles this. We should think clean, modern "ChatGPT vibe" for the UX.

PR #359, opened 2026-05-26
#353 Refactor, overhaul, ChatCommon

dsargent:
[Copilot overview](https://github.com/cognizant-ai-lab/neuro-san-ui/pull/359#pullrequestreview-4376093617).

Technical PR only. Major refactoring of the `ChatCommon` component. Clean up, componentization, React.js hygiene.

Motivation: after React.js Strict Mode was enabled (#351), we started seeing double rendering issues in the chat component. This was a "smell", a sign that things weren't being done right and not following the "rules of React". An offhand comment by Copilot made me realize `ChatCommon` was doing unhealthy things, like maintaing  the conversation as a mutable list of React.js nodes, modifying that list, and other hacky things like retroactively filtering the output to show/hide items depending on the setting of "show thinking". It was also co-mingling data (the list of messages in the conversation) with presentation (the rendered nodes). Copilot suggested something similar to the approach I ended up taking, with a separate data model, orthogonal to the presentation logic.

This PR addresses those items, and it:

1) Makes rendering almost 100% pure and immutable, in line with React.js guidelines.
1) Creates a separate data-only structure to hold the conversation.
1) Renders that data structure as part of the render function, separately from the structure itself
1) Breaks separate items out it into their own components:
    1)  Conversation-related items go in the new `Conversation` component (not to be confused with thought bubble "conversations"). Conversations are now modelled as a set of a new `ConversationTurn` entity.
    1) Agent intro/greeting gets its own component `AgentIntro`
    1) Agent metadata (network topology, sample queries) get their own component, `AgentMetadata`
1) Logic related to rendering the conversation is moved to the new `Conversation` component, which makes `ChatCommon` less monolithic and more focused
1) All of the (impure) calls to `updateOutput` are placed by adding items to the `turns` list, which is pure data; it only is consumed later at render time and only then converted into displayable items in a pure, React.js-idiomatic way.
1) Things that were previously "imperative", relying on effects, state synchronization, setting flags to remember where we were, are now done via declarative components in the "React way". This allowed removing effects and refs, simplified dependency arrays, and got rid of the multiple render issues, because the component is now much closer to the React.js "pure functional" ideal¹. For example, the agent intro and network info now naturally show up where needed, exactly once.
1) "Final answer" handling is also vastly simplified, as we no longer have the hacky solution of holding a "ref" to the final answer item in a mutable array of rendered nodes.

Known issues

* Tested with Neuro AI and mostly works fine. Just some strange looking headers when using Opp Finder, but still works and still usable. Not something I want to spend time fixing right now. Getting this overhaul deployed is more important.

¹The one item I wasn't able to fix 100% was the use of `Math.random()` to generate a random greeting. I found myself caught between two different linter rules whichever way I did it and even Claude ended up shrugging and saying "just disable the rule this once". I finally managed to come up with something using on-mount state. But, React.js rendering just doesn't like randomness.


apps/main/pages/_app.tsx
```diff
@@ -15,6 +15,7 @@ limitations under the License.
 */
 
 import "../styles/globals.css"
+import "@xyflow/react/dist/style.css"
```
dsargent, line 18: Not _directly_ related to this PR, but...I had to move this stylesheet back here. With the stylesheet imported in `AgentFlow`, when I tried to consume this library from Neuro AI, Node.js was trying to import the stylesheet and barfing on it. Known issue apparently. Copilot said, "make it the consumer's responsibility to import the stylesheet". 🤷🏻

apps/main/package.json
```diff
@@ -18,6 +18,7 @@
         "@mui/material": "7.3.1",
         "@mui/system": "catalog:default",
         "@mui/x-tree-view": "8.22.0",
+        "@xyflow/react": "catalog:default",
```
dsargent, line 21: The cost of having the stylesheet in the main module. At least we can use the catalog here to keep versions consistent.

packages/ui-common/components/AgentChat/ChatCommon/AgentConnectivity.tsx
```diff
@@ -3,6 +3,20 @@ import {FC} from "react"
 import {ConnectivityInfo} from "../../../generated/neuro-san/NeuroSanClient"
 import {MUIAccordion} from "../../Common/MUIAccordion"
 
+interface AgentConnectivityProps {
```
dsargent, line 6: Moved this earlier to try to be consistent.

packages/ui-common/components/AgentChat/ChatCommon/AgentIntro.tsx
dsargent, line 1: One of the components broken out from `ChatCommon`. Handles this:

Image

packages/ui-common/components/AgentChat/ChatCommon/AgentIntro.tsx
```diff
@@ -0,0 +1,37 @@
...
+    readonly agentDisplayName: string
+    readonly customAgentGreetings?: Partial<Record<CombinedAgentType, string>>
+    readonly targetAgent: string
+}
+
+/**
+ * Component to display agent intro, including agent name, image, and greeting.
+ */
+export const AgentIntro: FC<AgentIntroProps> = ({agentDisplayName, customAgentGreetings, targetAgent}) => {
+    // eslint-disable-next-line react/hook-use-state -- we want to set this an "on mount" value only and never update
```
dsargent, line 18: A minor wart, a sand grain of impurity. React.js (and specifically, the `react-hook` linter rules) do _not_ like randomness in the rendering path, as it's considered "impure": for the same inputs, you can get different outputs, which is a sin in React.js.

After a lot of time with Copilot, this was the best compromise I could come up with. A no-setter, initialize-on-mount state variable, which gets set up before the rendering so React.js is happy. I did have to silence the "hey, why no setter?!" linter rule, but in my opinion that is a much lesser sin than having to silence the "impure render" rule.

packages/ui-common/components/AgentChat/ChatCommon/AgentMetadata.tsx
dsargent, line 1: An umbrella component that handles showing the agent intro, network details, and sample queries. These all come from the same API call so it makes sense to group them into a single parent component.

packages/ui-common/components/AgentChat/ChatCommon/AgentMetadata.tsx
```diff
@@ -0,0 +1,85 @@
...
+    handleSend,
+    id,
+    neuroSanURL,
+    targetAgent,
+}) => {
+    const [sampleQueries, setSampleQueries] = useState<string[]>([])
+    const [connectivityInfo, setConnectivityInfo] = useState<readonly ConnectivityInfo[]>([])
+    const [description, setDescription] = useState<string>("")
+
+    useEffect(() => {
```
dsargent, line 34: We fetch what we need here

packages/ui-common/components/AgentChat/ChatCommon/AgentMetadata.tsx
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        if (targetAgent && !isLegacyAgentType(targetAgent)) {
+            void fetchAgentDetails()
+        }
+    }, [currentUser, neuroSanURL, targetAgent])
+
+    return (
+        <>
+            {description && connectivityInfo?.length > 0 && (
```
dsargent, line 68: Conditionally render the intro and sample queries.

packages/ui-common/components/AgentChat/ChatCommon/Const.ts
dsargent, line 1: Created and moved to here for reuse.

packages/ui-common/components/AgentChat/ChatCommon/Conversation.tsx
dsargent, line 1: Another abstraction. Rather than having the "mutable list of react nodes" in `ChatCommon`, which was anti-React.js-like, buggy, and required a lot of maintenance, we model a "conversation" as a series of "turns": human message, agent message, agent header etc. The turns are just abstract data. The rendering happens in a pure manner, in the rendering path.

Logic moved here and centralized from various places in `ChatCommon`.

packages/ui-common/components/AgentChat/ChatCommon/Conversation.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+/**
+ * Renders a "turn" from the conversation
+ * @param darkMode Whether the current theme is in dark mode. Used to determine syntax highlighter theme.
+ * @param shadowColor The color to use for shadows on the final answer accordion.
+ * @param shouldWrapOutput Whether to wrap long lines in the output. Passed down to the syntax highlighter component.
+ * @param turn The turn to render. @see ConversationTurn type for more info
+ * @returns A React component representing the "turn"
+ */
+const renderTurn = (
```
dsargent, line 35: This converts "turns" into display stuff.

packages/ui-common/components/AgentChat/ChatCommon/Conversation.tsx
```diff
@@ -0,0 +1,200 @@
...
+    // extract the parts of the line
+    let repairedJson: string
+
+    try {
+        // Attempt to parse as JSON
+
+        // First, repair it. Also replace "escaped newlines" with actual newlines for better display.
+        repairedJson = jsonrepair(turn.text)
+
+        // Now try to parse it. We don't care about the result, only if it throws on parsing.
```
dsargent, line 50: We agreed we can get rid of this, but I didn't want to for this PR. #277

packages/ui-common/components/AgentChat/ChatCommon/Conversation.tsx
```diff
@@ -0,0 +1,200 @@
...
+    const theme = useTheme()
+    const darkMode = theme.palette.mode === "dark"
+    const shadowColor = darkMode ? theme.palette.common.white : theme.palette.common.black
+
+    /**
+     * Render the list of conversation turns.
+     */
+    const nodesList: ReactNode[] = useMemo(
+        () =>
+            turns.flatMap((turn): ReactNode[] => {
```
dsargent, line 123: Handle the various types of "turns". 

Why `flatMap`? Allows one turn to render more than one output item, and perhaps more importantly, filters out the nulls. On reflection could maybe do this simpler without the array semantics and with a separate `filter` expression; not sure if that's better.

packages/ui-common/components/AgentChat/ChatCommon/Conversation.tsx
```diff
@@ -0,0 +1,200 @@
...
+                        return [
+                            <MUIAlert
+                                id={`error-${turn.id}-alert`}
+                                key={turn.id}
+                                severity="error"
+                            >
+                                {turn.text}
+                            </MUIAlert>,
+                        ]
+                    /* istanbul ignore next -- impossible to trigger this without subverting tsc */
```
dsargent, line 180: I really don't like disabling coverage, but this `default` is, by it's nature, impossible to hit. You'd have to subvert the `tsc` compiler in awful ways.

But the nice thing here is, if a new turn type is added, and the dev forgets to handle it in this `switch`, `tsc` will scream loudly. There is a linter rule for that (all cases handled) but we don't have it enabled.

packages/ui-common/components/AgentChat/ChatCommon/Conversation.tsx
```diff
@@ -0,0 +1,200 @@
...
+                        const _exhaustive: never = turn.role
+                        throw new Error(`Unhandled message role: ${_exhaustive}`)
+                    }
+                }
+            }),
+        [currentUser, darkMode, finalAnswerRef, shadowColor, shouldWrapOutput, showThinking, turns, userImage]
+    )
+
+    return (
+        <FormattedMarkdown
```
dsargent, line 193: I have a suspicion we are double-wrapping messages in `ReactMarkdown`, which may explain some of the ugly output and over-escaped JSON we see. Needs investigating.

The two places it's wrapped would be here and in `renderTurn`.

But this is a refactoring, so if we are double-wrapping, we _always_ have been.

packages/ui-common/components/AgentChat/ChatCommon/ConversationTurn.ts
dsargent, line 1: The new data model for modelling conversations.

packages/ui-common/components/AgentChat/ChatCommon/Greetings.ts
```diff
@@ -15,7 +15,7 @@ limitations under the License.
 */
 
 // Random greetings agents can use
-export const AGENT_GREETINGS: string[] = [
+export const GENERIC_AGENT_GREETINGS: string[] = [
```
dsargent, line 18: Renamed for clarity -- since we have generic greetings (these) and custom greetings (currently only AND has one).

packages/ui-common/components/MultiAgentAccelerator/AgentFlow.tsx
```diff
@@ -13,7 +13,6 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-import "@xyflow/react/dist/style.css"
```
dsargent, line 16: Removed from here; see explanation in `_app.tsx`.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -731,7 +731,7 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
                     }}
                 >
                     <ChatCommon
-                        agentGreetings={{
+                        customAgentGreetings={{
```
dsargent, line 734: Renamed for clarity.

packages/ui-common/package.json
```diff
@@ -12,7 +12,7 @@
     "dependencies": {
         "@dagrejs/dagre": "1.0.4",
         "@langchain/core": "catalog:default",
-        "@xyflow/react": "12.10.1",
+        "@xyflow/react": "catalog:default",
```
dsargent, line 15: Version moved to catalog.

jest.config.ts
```diff
@@ -74,10 +74,10 @@ const config: Config.InitialOptions = {
 
     coverageThreshold: {
         global: {
-            statements: -120,
-            branches: -161,
+            statements: -119,
```
dsargent: 3 improved, branches down a notch. I'll take it.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
dsargent, line 1: Obviously this is ground zero for the refactor. See PR description for overview of the changes.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -271,26 +253,19 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
     // For tracking if we're auto-scrolling. A button allows the user to enable or disable auto-scrolling.
     const [autoScrollEnabled, setAutoScrollEnabled] = useState<boolean>(true)
 
-    // ref for same
```
dsargent, line 274: Not needed, we do things more cleanly now.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -271,26 +253,19 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
-    const autoScrollEnabledRef = useRef<boolean>(autoScrollEnabled)
-
     // Whether to wrap output text
     const [shouldWrapOutput, setShouldWrapOutput] = useState<boolean>(true)
 
     // Keeps a copy of the last AI message so we can highlight it as "final answer"
     const lastAIMessage = useRef<string>("")
 
-    // Ref for the final answer key, so we can highlight the accordion
```
dsargent, line 283: Not needed, we do things more cleanly now.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -271,26 +253,19 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
-    const finalAnswerKey = useRef<string>("")
-
     // Persistent agent chat history store, which is where we store both kinds of chat histories
     // (see store implementation for details)
     const storedChatHistory = useAgentChatHistoryStore((state) => state?.history?.[targetAgent])
     const agentChatHistory = useMemo(
         () => storedChatHistory ?? {chatHistory: [], chatContext: null, slyData: {}},
         [storedChatHistory]
     )
-    const [agentSampleQueries, setAgentSampleQueries] = useState<string[]>([])
```
dsargent, line 293: Moved to `SampleQueries` component.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -336,10 +311,6 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
     // Keeps track of whether the agent completed its task
     const succeeded = useRef<boolean>(false)
 
-    const darkMode = theme.palette.mode === "dark"
```
dsargent, line 339: Moved to `Conversation` component.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -350,128 +321,38 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
-    }, [autoScrollEnabled])
-
     useEffect(() => {
         // Delay for a second before focusing on the input area; gets around ChatBot stealing focus.
         setTimeout(() => chatInputRef?.current?.focus(), 1000)
     }, [])
 
     // Auto scroll chat output window when new content is added
     useEffect(() => {
+        const container = chatOutputRef.current
```
dsargent, line 331: Cleaned this up a little, but basically works the same. Direct DOM manipulation stuff like "scrolling into view" (ironically the `scrollIntoView` API didn't work here, can't remember why) are still handled outside of the rendering path.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -350,128 +321,38 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
 
-        if (autoScrollEnabledRef.current && chatOutputRef?.current) {
-            chatOutputRef.current.scrollTop = chatOutputRef.current.scrollHeight
+        // Live-streaming auto-scroll
+        if (autoScrollEnabled) {
+            container.scrollTop = container.scrollHeight
         }
-    }, [chatOutput, isAwaitingLlm])
-
-    /**
```
dsargent, line 376: Moved to `Conversation` component and renamed to `renderTurn`, where it is now pure (no side effects, unlike this one.)

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -350,128 +321,38 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
-            const next = [...current, node]
+    const addTurn = useCallback((turn: ConversationTurn) => {
+        setTurns((current) => {
+            const next = [...current, turn]
             return next.length > MAX_CHAT_OUTPUT_ITEMS ? next.slice(-MAX_CHAT_OUTPUT_ITEMS) : next
         })
     }, [])
 
+    // We use this to update the same "turn" as chunks come in from legacy agents
+    const legacyTurnIdRef = useRef<string | null>(null)
```
dsargent, line 354: The cost of supporting legacy agents. See usage below.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -480,10 +361,25 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
 
             // For legacy agents, we either get plain text or Markdown. Just output it as-is.
             if (isLegacyAgentType(targetAgent)) {
-                // Display output as-is
-                updateOutput(chunk)
                 currentResponse.current += chunk
 
+                if (!legacyTurnIdRef.current) {
```
dsargent, line 366: A bit of ugliness. This is mutability but no easy way to avoid it. The response from legacy agents arrive in chunks (basically token by token) so we accumulate the entire response within a single `Turn`. Note this is faster than what we had before, since now React knows it only has to re-render this one turn. Previously we were mutating the _entire output array_. Now it should be _much faster_. Working _with_ React, not fighting it.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -500,6 +396,11 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
                 return
             }
 
+            // Keep track of AI messages. The last one is (by definition) the "final answer" from the agents.
+            if (chatMessage.type === ChatMessageType.AI && chatMessage.text) {
```
dsargent, line 400: Previously `updateOutput` did this as a side effect, but no more.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -517,14 +418,12 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
                 if (errorMessage) {
-                    updateOutput(
-                        <MUIAlert
-                            id="retry-message-alert"
-                            severity="warning"
-                        >
-                            {errorMessage}
-                        </MUIAlert>
-                    )
+                    addTurn({
```
dsargent, line 421: Throughout this file, you will see `updateOutput` replaced with `addTurn`. `updateOutput` was conflating data with presentation, and mutating the displayed info directly; `addTurn` simply appends to our data which gets rendered later on. Much cleaner separation of data and presentation, safer, easier to reason about. And doesn't run afoul of "React strict mode"; the double render issues are _gone_.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -535,32 +434,18 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+                    text: chatMessage.text,
+                })
                 currentResponse.current += chatMessage.text
             }
         },
-        [onChunkReceived, processLogLine, updateSlyData, targetAgent, updateChatContext, updateOutput]
+        [onChunkReceived, targetAgent, addTurn, updateSlyData, updateChatContext]
     )
 
-    const introduceAgent = useCallback(() => {
```
dsargent, line 545: Moved to `AgentIntro`.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -571,9 +456,9 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
         lastAIMessage.current = ""
         finalAnswerRef.current = null
 
-        // Get agent name, either from the enum (Neuro-san) or from the targetAgent string directly (legacy)
         setPreviousResponse?.(targetAgent, currentResponse.current)
         currentResponse.current = ""
+        legacyTurnIdRef.current = null
```
dsargent, line 461: We need to reset this between interactions too.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -244,16 +227,15 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
     // MUI theme
     const theme = useTheme()
-    const shadowColor = theme.palette.mode === "dark" ? theme.palette.common.white : theme.palette.common.black
 
     // User LLM chat input
     const [chatInput, setChatInput] = useState<string>("")
 
     // Previous user query (for "regenerate" feature)
-    const previousUserQuery = useRef<string>("")
+    const [previousUserQuery, setPreviousUserQuery] = useState<string>("")
```
dsargent, line 235: I made this a `ref` initially in a naive attempt at optimization; refs don't cause re-renders, and we don't want to re-render just because the previous user query changed, right?

Wrong. I was wrong. Copilot convinced me (eventually). We absolutely _do_ want to re-render when it changes, because we are passing it as a prop to `ControlButtons` for the "resend" functionality. If we don't make it state, we won't re-render when it changes, `ControlButtons` won't detect the change, and there's a chance that "resend" could resent the _wrong_ user query. It was working previously "by accident" because other things forced a re-render; now it's working by default, and by design.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -671,30 +554,30 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
             // that we generated behind their back. Ultimately, we shouldn't need to generate a fake query on behalf
             // of the user, but currently we do for orchestration.
-            updateOutput(
-                <UserQueryDisplay
-                    userQuery={query}
-                    title={currentUser}
-                    userImage={userImage}
-                />
-            )
+            addTurn({
```
dsargent, line 565: Pretty much one for one replace impure `updateOutput` with adding turns.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -703,51 +586,34 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+                        id: uuid(),
+                        role: MessageRole.Error,
+                        text: `Gave up after ${MAX_AGENT_RETRIES} attempts.`,
+                    })
                 }
 
                 // Display prominent "Final Answer" message if we have one
                 if (lastAIMessage.current) {
-                    // Legacy agents text is a bit messy and doesn't add a blank line, so we add it here
-                    if (isLegacyAgentType(targetAgent)) {
```
dsargent, line 738: We don't need this hacky rendering stuff here any more.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -757,9 +623,6 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
                     // as the bot answer in that case.
                     updateChatHistory(targetAgent, [new AIMessage({content: currentResponse.current, id: uuid()})])
                 }
-
```
dsargent, line 760: Nor here

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -768,83 +631,33 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
         ]
     )
 
-    useEffect(() => {
-        if (targetAgent) {
-            introduceAgent()
-        }
-    }, [targetAgent, introduceAgent])
-
-    useEffect(() => {
```
dsargent, line 794: This gets moved to `AgentMetadata`.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -865,44 +678,18 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+        setTurns([])
         resetHistory(targetAgent)
-        previousUserQuery.current = ""
+        setPreviousUserQuery("")
         currentResponse.current = ""
         lastAIMessage.current = ""
-        introduceAgent()
-    }, [introduceAgent, resetHistory, targetAgent])
-
-    /**
```
dsargent, line 882: This is (probably) slow and no longer needed. At render time, we do the right thing, based on what we have in the `turns` array. Much simpler and cleaner.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -1015,6 +802,15 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
         </>
     )
 
+    const agentIntro = (
```
dsargent, line 805: We use this in a couple of places (see below) so factor it out.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -1061,17 +857,29 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
-                    id={`${id}-formatted-markdown`}
-                    nodesList={nodesList}
-                    style={darkMode ? atelierDuneDark : a11yLight}
-                    wrapLongLines={shouldWrapOutput}
-                />
-                <SampleQueries
-                    disabled={isAwaitingLlm}
-                    handleSend={handleSend}
-                    sampleQueries={agentSampleQueries}
+                {/*For Neuro-san agents, intro goes above the metadata and conversation*/}
```
dsargent, line 860: A wart. The opp finder agents want to have a continuous conversation; as you switch agents, the chat just keeps going. So they want the intro stuff after the conversation. In MAUI when you switch agents, you want the intro block first, then the chat.

This is a hacky way to deal with that. It still doesn't work 100% in Neuro AI -- you get intro, chat, spinner while it's busy, and then another intro in preparation for the _next_ interaction. It doesn't look great but it works. And I'm more concerned about the MAUI use case right now.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -1061,17 +857,29 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+                {!isLegacyAgentType(targetAgent) && agentIntro}
+                {!isLegacyAgentType(targetAgent) && (
+                    <AgentMetadata
```
dsargent, line 863: This item now encapsulates the agent intro (greeting + network info) and is added here as a render item. Previously we were injecting the greeting into the chat stream. This is cleaner and avoids (1) the double rendering issue and (2) having to keep track of when we inject it and ugly effects.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -1061,17 +857,29 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+                        disableQueries={isAwaitingLlm}
+                        handleSend={handleSend}
+                        currentUser={currentUser}
+                        id={`${id}-agent-metadata-display`}
+                        neuroSanURL={neuroSanURL}
+                        targetAgent={targetAgent}
+                    />
+                )}
+                <Conversation
```
dsargent, line 872: This replaces the mutable nodes array. Now `Conversation` is its own first-class component with its own props and rendering cycles. "Correct by default".

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -1095,12 +903,12 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
             </Box>
 
             <ControlButtons
-                handleClearChat={handleClearChat}
                 enableClearChatButton={enableClearChatButton}
-                isAwaitingLlm={isAwaitingLlm}
+                handleClearChat={handleClearChat}
```
dsargent, line 907: Alphabetic sorting of props, and now we send `previousUserQuery` as a _state_ variable rather than ref, so that re-renders are triggered when they should be and only when they should be.

packages/ui-common/__tests__/components/AgentChat/ChatCommon/AgentMetadata.test.tsx
dsargent, line 1: Tests for new component. Mostly just to get coverage up tbh.

__tests__/common/NetworksListMock.ts
```diff
@@ -72,3 +72,24 @@ export const TEMPORARY_NETWORK: TemporaryNetwork = {
     agentNetworkName: TEMPORARY_NETWORK_NAME,
     networkHocon: JSON.stringify(TEMP_NETWORK_AGENT_INFO, null, 2),
 }
+
+export const TEST_TOOL_SPOTIFY = "spotify_tool"
```
dsargent, line 76: Moved some more common test data to here so it can be shared.

packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
```diff
@@ -73,22 +70,6 @@ const getResponseMessage = (type: ChatMessageType, text: string): ChatMessage =>
     },
 })
 
-const MOCK_CONNECTIVITY_INFO = {
```
dsargent, line 76: Moved to common module for sharing.

packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
```diff
@@ -197,13 +182,14 @@ describe("ChatCommon", () => {
...
 
-        // Check that connectivity info is rendered correctly
-        await user.click(await screen.findByText("Network Details")) // Click to expand
-        const connectivityList = screen.getByRole("list", {name: CONNECTIVITY_LIST_HEADER})
+        await screen.findByText(TEST_AGENT_MATH_GUY_DISPLAY)
 
-        // Should be no connectivity info rendered
-        expect(within(connectivityList).queryByText(TEST_AGENT_MATH_GUY_DISPLAY)).toBeNull()
-        expect(within(connectivityList).queryByText(TEST_AGENT_MUSIC_NERD)).toBeNull()
+        // Connectivity info block should not be present
```
dsargent, line 187: Functionality changed here slightly; we don't even display items we don't have data for any more.

packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
```diff
@@ -417,7 +406,7 @@ describe("ChatCommon", () => {
         screen.getByText("Final Answer")
     })
 
-    it("Should correctly handle errors thrown while fetching", async () => {
+    it("Should handle error thrown while fetching", async () => {
```
dsargent, line 409: This test was randomly timing out in CI. I split it into two to see if that helps.

packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
```diff
@@ -578,6 +573,20 @@ describe("ChatCommon", () => {
         await screen.findByText(TEST_AGENT_MUSIC_NERD_DISPLAY)
     })
 
+    it("Should use custom agent greetings when supplied", async () => {
```
dsargent, line 576: New test for custom greetings.

packages/ui-common/__tests__/components/AgentChat/ChatCommon/Conversation.test.tsx
dsargent, line 1: Test for new component.

packages/ui-common/__tests__/components/AgentChat/ChatCommon/FormattedMarkdown.test.tsx
dsargent, line 1: Added some cases to these tests to get coverage up.

.yarnrc.yml
```diff
@@ -19,6 +19,7 @@ catalogs:
         "@types/node": "22.17.1"
         "@types/react": "19.2.14"
         "@types/react-dom": "19.2.3"
+        "@xyflow/react": "12.10.1"
```
dsargent, line 22: Centralize version here.

packages/ui-common/components/AgentChat/ChatCommon/AgentIntro.tsx
swensel: I suppose it is fine for now. I don't have a better idea for the time being.

dsargent: The random greetings are a bit cheesy so may consider dropping them entirely.

swensel (dismissed):
LGTM. It's a great improvement over what we had.

swensel (approved):

Issue #361, opened 2026-05-27
Automate jest coverage adjustment


dsargent:
Create automation to update the values for various coverage categories in `jest.config.ts` based on current test numbers.

PR #362, opened 2026-05-27
#361 Automate jest coverage threshold updates

dsargent:
Helps with getting PRs to green by temporarily lowering coverage, or "claiming" progress when adding to tests to increase it.

Not much to see:

```console
❯ yarn update-coverage
Coverage updated successfully in jest.config.ts
```

Then the values get updated in `jest.config.ts` to current test numbers.

Similar in spirit to [jest-it-up](https://github.com/rbardini/jest-it-up) but (1) works for our case -- we use absolute coverage numbers, whereas jest-it-up only supports percentages and (2) adjusts up or down, without judgement, so you can either temporarily lower the heat while working on a PR or claim progress when you improve the coverage.

Possible enhancements: 
* Option to use existing coverage report to avoid running tests, which takes a while.
* ~Less destructive: don't blow away existing coverage report~ Done
* Integrate with PRs/GitHub Action? "Hey, you have increased test coverage. You can up the values in the config."

### DIFF
--- a/build_scripts/UpdateCoverage.sh
+++ b/build_scripts/UpdateCoverage.sh
@@ -50,51 +50,48 @@ main() {
         exit 1
     fi
 
-    # remove old coverage summary
-    rm -f coverage/coverage-summary.json
+    # Create a temporary directory for coverage output and ensure it gets cleaned up on exit
+    tmp_dir="$(mktemp -d)"
+    trap 'rm -rf "${tmp_dir}"' EXIT
+
+    # Where we want to put output files
+    coverage_summary="${tmp_dir}/coverage-summary.json"
+    new_coverage_values="${tmp_dir}/new_coverage_values.txt"
 
     # Run the tests with coverage
-    if ! jest --config ./jest_quiet.config.ts --coverage --coverageReporters=json-summary --coverageThreshold '{}'; then
+    if ! jest --config ./jest_quiet.config.ts --collectCoverage --coverageReporters=json-summary \
+        --coverageThreshold '{}' --coverageDirectory="${tmp_dir}"; then
         # If jest failed, don't update
         echo "Tests failed. Coverage not updated." >&2
         exit 1
     fi
 
     # If coverage summary file doesn't exist, exit with error
-    if [ ! -f coverage/coverage-summary.json ]; then
-        echo "Error: Coverage file \"coverage/coverage-summary.json\" not found. Unable to proceed." >&2
+    if [ ! -f "${coverage_summary}" ]; then
+        echo "Error: Coverage file \"${coverage_summary}\" not found. Unable to proceed." >&2
         exit 1
     fi
 
     # Use jq to extract the uncovered counts and format them as a new global block for jest.config.ts.
     jq -r '
-  "        global: {",
-  "            statements: -\(.total.statements.total - .total.statements.covered),",
-  "            branches: -\(.total.branches.total - .total.branches.covered),",
-  "            functions: -\(.total.functions.total - .total.functions.covered),",
-  "            lines: -\(.total.lines.total - .total.lines.covered),",
-  "        },"
-' coverage/coverage-summary.json >/tmp/new_global_block.txt
+    "        global: {",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -22,7 +22,8 @@
         "lint-rules": "./build_scripts/eslint_list_rules.sh",
         "lint-summary": "eslint --format summary",
         "start": "yarn workspace @cognizant-ai-lab/neuro-san-web start",
-        "test": "jest --config ./jest.config.ts"
+        "test": "jest --config ./jest.config.ts",
+        "update-coverage": "./build_scripts/UpdateCoverage.sh"
```
dsargent, line 26: New yarn script to launch it.

build_scripts/UpdateCoverage.sh
dsargent, line 1: Bash script that does the work.

build_scripts/UpdateCoverage.sh
```diff
@@ -0,0 +1,100 @@
...
+' coverage/coverage-summary.json >/tmp/new_global_block.txt
+
+    # Update coverage numbers in jest.config.ts.
+    if sed --version >/dev/null 2>&1; then
+        sed_inplace=(-i) # GNU sed
+    else
+        sed_inplace=(-i '') # BSD sed (macOS)
+    fi
+
+    if sed "${sed_inplace[@]}" '/^[[:space:]]*global:[[:space:]]*{/,/^[[:space:]]*},[[:space:]]*$/{
```
dsargent, line 86: This is a bit ugly. Since we're dealing with TypeScript (`jest.config.ts`) we have to regex our way into injecting the new coverage values. 

Couple of alternative (more complex) approaches that I rejected:

1) Externalize the coverage numbers in a JSON file, consume that from the `jest.config.ts` TypeScript code. This script would then just update that pure JSON file, which would be a lot easier. Downside: change to how we do things. We're used to seeing the numbers in `jest.config.ts`. But maybe not a bad idea to externalize them anyway in a pure JSON "data file"? 🤔 To be considered for future.

2) Use some kind of Typescript lexical parser to do the surgery. _Way_ overkill for this, IMO.

So, ugly regex it is. The world's most popular write-only language.

build_scripts/UpdateCoverage.sh
```diff
@@ -0,0 +1,100 @@
...
+    if ! command -v jq &>/dev/null; then
+        echo "Error: jq is not installed. Please install it and try again." >&2
+        exit 1
+    fi
+
+    # remove old coverage summary
+    rm -f coverage/coverage-summary.json
+
+    # Run the tests with coverage
+    if ! jest --config ./jest_quiet.config.ts --coverage --coverageReporters=json-summary --coverageThreshold '{}'; then
```
dsargent: Override `coverageThreshold` here -- we don't want the tests to fail due to that because that's exactly what we're planning to update.

build_scripts/UpdateCoverage.sh
```diff
@@ -0,0 +1,100 @@
...
+    fi
+
+    # If coverage summary file doesn't exist, exit with error
+    if [ ! -f coverage/coverage-summary.json ]; then
+        echo "Error: Coverage file \"coverage/coverage-summary.json\" not found. Unable to proceed." >&2
+        exit 1
+    fi
+
+    # Use jq to extract the uncovered counts and format them as a new global block for jest.config.ts.
+    jq -r '
```
dsargent, line 76: TIL `jq` can do arithmetic.

build_scripts/UpdateCoverage.sh
```diff
@@ -0,0 +1,97 @@
...
+        exit 1
+    fi
+
+    # if jq not present, exit
+    if ! command -v jq &>/dev/null; then
+        echo "Error: jq is not installed. Please install it and try again." >&2
+        exit 1
+    fi
+
+    # Create a temporary directory for coverage output and ensure it gets cleaned up on exit
```
dsargent, line 53: Now with tmpdir goodness. This is a better way of handling things.

swensel (approved):
LGTM. Thanks for putting this together @dsargent

PR #273, opened 2026-04-24
Refactor ChatCommon

dsargent:
[Copilot summary](https://github.com/cognizant-ai-lab/neuro-san-ui/pull/273#pullrequestreview-4176018311).

Mostly just a clean-up to put things into subfolders and modularize `ChatCommon`, factoring out independent parts into their own component, such as `SampleQueries` and `AgentConnectivity`. Adapted test folder structures accordingly to mirror the modules under test.

Some enhanced test coverage of `ChatCommon`, for the first time bringing us above 90% on branch coverage overall 🎉

A lot of the changes in this PR is just shuffling things around and adjusting paths. The non-trivial stuff hits `ChatCommon.tsx` the hardest.

Summary of structure:

Before:

image
<br />
After:
<br />
<br />
image
<br />
Or, neatly collapsed:
<br />
<br />
image

PR #371, opened 2026-05-30
Enable linter rules and fix violations

dsargent:
The origin of this was in a previous PR #273 I noticed we had no rule for catching when not all cases are handled in a `switch`. So I enabled that rule here, and a bunch more, and fixed the resulting violations.

I picked rules that seemed useful, with a few easy-to-fix violations in the existing codebase.

Broad but shallow PR.

### DIFF
--- a/apps/main/next.config.ts
+++ b/apps/main/next.config.ts
@@ -3,7 +3,7 @@
 // enables IDEs to type check this config file.
 // See: https://nextjs.org/docs/basic-features/typescript
 
-import path from "path"
+import path from "node:path"
 
 /* eslint-disable no-shadow */
 const __dirname = import.meta.dirname
--- a/apps/main/pages/api/userInfo/index.ts
+++ b/apps/main/pages/api/userInfo/index.ts
@@ -21,19 +21,6 @@ import {OidcProvider, UserInfoResponse} from "../../../../../packages/ui-common/
 
 const EXPECTED_NUMBER_OF_JWT_HEADERS = 3
 
-export default async function handler(req: NextApiRequest, res: NextApiResponse<UserInfoResponse>) {
-    // Fetch user info from ALB or your backend service
-    const userInfo = fetchUserInfoFromALB(req)
-    if (userInfo.oidcHeaderFound && !userInfo.oidcHeaderValid) {
-        // We found the header, but couldn't parse it
-        res.status(httpStatus.UNAUTHORIZED).json({oidcHeaderFound: true, oidcHeaderValid: false})
-        return
-    }
-
-    // We either didn't find the header at all, or we found it and parsed it
-    res.status(httpStatus.OK).json(userInfo)
-}
-
 // Expected header from ALB
 const AWS_OIDC_HEADER = "x-amzn-oidc-data"
 
@@ -88,4 +75,17 @@ const fetchUserInfoFromALB = (req: NextApiRequest): UserInfoResponse => {
         oidcHeaderValid: true,
         oidcProvider,
     }
+}
+
+export default async function handler(req: NextApiRequest, res: NextApiResponse<UserInfoResponse>) {
+    // Fetch user info from ALB or your backend service
+    const userInfo = fetchUserInfoFromALB(req)
+    if (userInfo.oidcHeaderFound && !userInfo.oidcHeaderValid) {
+        // We found the header, but couldn't parse it
+        res.status(httpStatus.UNAUTHORIZED).json({oidcHeaderFound: true, oidcHeaderValid: false})
+        return
+    }
+
+    // We either didn't find the header at all, or we found it and parsed it
+    res.status(httpStatus.OK).json(userInfo)
 }
--- a/packages/dev-common/Configs/eslint.config.mjs
+++ b/packages/dev-common/Configs/eslint.config.mjs
@@ -39,10 +39,6 @@ export default defineConfig([
     eslintPluginUnicorn.configs.all,
 
     typescriptEslint.configs.all,
-    {
-        files: ["**/*.{js,mjs,cjs}"],
-        ...typescriptEslint.configs.disableTypeChecked,
-    },
 
     // Have to configure import plugin in the "legacy" way due to:
     // https://github.com/import-js/eslint-plugin-import/issues/3227
@@ -402,7 +398,7 @@ export default defineConfig([
             "@typescript-eslint/prefer-return-this-type": "off",
             "@typescript-eslint/prefer-string-starts-ends-with": "off",
             "@typescript-eslint/require-array-sort-compare": "off",
-            "@typescript-eslint/switch-exhaustiveness-check": "off",
+            "@typescript-eslint/switch-exhaustiveness-check": "error",
             "@typescript-eslint/unbound-method": "off",
             "@typescript-eslint/no-for-in-array": "off",
             "@typescript-eslint/object-curly-spacing": "off",
@@ -593,6 +589,11 @@ export default defineConfig([
             "import/no-anonymous-default-export": "error",
             "@typescript-eslint/no-empty-interface": "error",
         },
+    },
+    // Disable typed linting rules for non-TypeScript files since they're not applicable
+    {
+        files: ["**/*.{js,mjs,cjs}"],
+        ...typescriptEslint.configs.disableTypeChecked,
     },
     {
         files: ["**/eslint.config.mjs"],
--- a/packages/dev-common/Configs/eslint.config.mjs
+++ b/packages/dev-common/Configs/eslint.config.mjs
@@ -396,7 +396,6 @@ export default defineConfig([
             "@typescript-eslint/prefer-reduce-type-parameter": "off",
             "@typescript-eslint/prefer-return-this-type": "off",
             "@typescript-eslint/require-array-sort-compare": "off",
-            "@typescript-eslint/switch-exhaustiveness-check": "error",
             "@typescript-eslint/unbound-method": "off",
             "@typescript-eslint/no-for-in-array": "off",
             "@typescript-eslint/object-curly-spacing": "off",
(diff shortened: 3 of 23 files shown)


apps/main/pages/api/userInfo/index.ts
```diff
@@ -21,19 +21,6 @@ import {OidcProvider, UserInfoResponse} from "../../../../../packages/ui-common/
 
 const EXPECTED_NUMBER_OF_JWT_HEADERS = 3
 
-export default async function handler(req: NextApiRequest, res: NextApiResponse<UserInfoResponse>) {
```
dsargent, line 24: Moved lower to fix "using item before definition" linter error

apps/main/next.config.ts
```diff
@@ -3,7 +3,7 @@
 // enables IDEs to type check this config file.
 // See: https://nextjs.org/docs/basic-features/typescript
 
-import path from "path"
+import path from "node:path"
```
dsargent, line 6: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-node-protocol.md

packages/dev-common/Configs/eslint.config.mjs
```diff
@@ -39,10 +39,6 @@ export default defineConfig([
     eslintPluginUnicorn.configs.all,
 
     typescriptEslint.configs.all,
-    {
```
dsargent, line 42: After discussion with Copilot, we agreed to move this to the bottom and it's by way of an "override". Since flat config is layered with "last one wins" semantics, you want the items to go from more general to less general. This `file` block is a specific escape hatch to avoid targeting JS-type files with typed linter rules (which makes no sense) so it should be later in the config.

apps/main/pages/api/userInfo/index.ts
```diff
@@ -89,3 +76,16 @@ const fetchUserInfoFromALB = (req: NextApiRequest): UserInfoResponse => {
         oidcProvider,
     }
 }
+
+export default async function handler(req: NextApiRequest, res: NextApiResponse<UserInfoResponse>) {
```
dsargent, line 80: Moved to here

packages/dev-common/Configs/eslint.config.mjs
```diff
@@ -396,13 +392,10 @@ export default defineConfig([
             "@typescript-eslint/non-nullable-type-assertion-style": "off",
             "@typescript-eslint/prefer-includes": "off",
             "@typescript-eslint/no-unsafe-type-assertion": "off",
-            "@typescript-eslint/no-use-before-define": "off",
```
dsargent, line 399: Remember we enable all rules by default, so removing the "off" disable here means, enabling that rule.

So the ones you see deleted here are newly enabled by this PR. I picked ones that seemed reasonable and didn't have a ton of violations to fix. Some are a bit opinionated so feel free to discuss.

packages/dev-common/Configs/eslint.config.mjs
```diff
@@ -396,13 +392,10 @@ export default defineConfig([
...
             "@typescript-eslint/prefer-readonly": "off",
             "@typescript-eslint/prefer-reduce-type-parameter": "off",
             "@typescript-eslint/prefer-return-this-type": "off",
-            "@typescript-eslint/prefer-string-starts-ends-with": "off",
             "@typescript-eslint/require-array-sort-compare": "off",
-            "@typescript-eslint/switch-exhaustiveness-check": "off",
```
dsargent, line 405: This was the original one I wanted, to protect us from not handling cases in `switch`.

packages/dev-common/Configs/eslint.config.mjs
```diff
@@ -594,6 +582,11 @@ export default defineConfig([
             "@typescript-eslint/no-empty-interface": "error",
         },
     },
+    // Disable typed linting rules for non-TypeScript files since they're not applicable
```
dsargent, line 585: Moved to here. See comment at top.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow.test.tsx
```diff
@@ -152,7 +152,7 @@ describe("AgentFlow", () => {
         expect(nodes).toHaveLength(3)
 
         const agentNames = NETWORK.map((agent) => agent.origin)
-        const nodesArray = Array.from(nodes)
+        const nodesArray = [...nodes]
```
dsargent, line 155: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-spread.md

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentNode.test.tsx
```diff
@@ -89,7 +89,7 @@ describe("AgentNode", () => {
         expect(parentDiv).toBeInTheDocument()
 
         // Check the background color of the associated node (the circle div)
-        const nodeDiv = Array.from(parentDiv.children).find((el) => el.tagName === "DIV") as HTMLDivElement
+        const nodeDiv = [...parentDiv.children].find((el) => el.tagName === "DIV") as HTMLDivElement
```
dsargent, line 92: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-spread.md

packages/ui-common/__tests__/components/MultiAgentAccelerator/MultiAgentAccelerator.test.tsx
```diff
@@ -481,7 +481,7 @@ describe("Multi Agent Accelerator Page", () => {
         // Verify the conversations array contains the expected agent
         const conversationCall = conversationMock.mock.calls[conversationMock.mock.calls.length - 1][0]
         const hasAgent = conversationCall.some((conv: {agents: Set<string>}) =>
-            Array.from(conv.agents).includes(TEST_AGENT_MATH_GUY)
+            [...conv.agents].includes(TEST_AGENT_MATH_GUY)
```
dsargent, line 484: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-spread.md

packages/ui-common/__tests__/unit/Utils/File.test.ts
```diff
@@ -49,11 +49,11 @@ describe("getFileName", () => {
     })
 
     it("should return the file name when the path includes backslashes", () => {
-        expect(getFileName("C:\\Users\\JaneDoe\\Documents\\myfile.txt")).toBe("myfile.txt")
+        expect(getFileName(String.raw`C:\Users\JaneDoe\Documents\myfile.txt`)).toBe("myfile.txt")
```
dsargent, line 52: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-string-raw.md

packages/ui-common/components/AgentChat/ChatCommon/Conversation.tsx
```diff
@@ -50,7 +50,7 @@ const renderTurn = (
         // Now try to parse it. We don't care about the result, only if it throws on parsing.
         JSON.parse(repairedJson)
 
-        repairedJson = repairedJson.replace(/\\n/gu, "\n").replace(/\\"/gu, "'")
+        repairedJson = repairedJson.replaceAll(String.raw`\n`, "\n").replaceAll(String.raw`\"`, "'")
```
dsargent, line 53: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-string-replace-all.md

packages/ui-common/components/MultiAgentAccelerator/Sidebar/Sidebar.tsx
```diff
@@ -184,6 +184,8 @@ export const Sidebar: FC<SidebarProps> = ({
 
     const [expandedItems, setExpandedItems] = useState<string[]>([])
 
+    const [selectedItem, setSelectedItem] = useState<string | null>(null)
```
dsargent, line 187: https://typescript-eslint.io/rules/no-use-before-define/

Also cleaner to have it with the other state items. I still want to find a linter rule to enforce the order of items within a module. I've seen a couple but no clear "winner".

packages/ui-common/components/MultiAgentAccelerator/ThoughtBubbleOverlay.tsx
```diff
@@ -409,16 +409,15 @@ export const ThoughtBubbleOverlay: FC<ThoughtBubbleOverlayProps> = ({
         [activeAgentIds]
     )
 
-    // Get all bubbles to render (including exiting ones)
```
dsargent, line 412: This one was interesting. Initially I changed it from `Array.from` to a spread, per [prefer-spread](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-spread.md). But when I did that, suddenly the `react-hooks` rule kicked in and objected, rightly, "hey! you're re-spreading this on every render! It should either be memoized, or moved inside the effect!".

It seems the react-hooks rule had a blind spot when it was `Array.from` but when changed to a spread, the rule caught the issue.

Therefore, I moved this item inside the effect which is simpler and more efficient.

packages/ui-common/utils/Authentication.ts
```diff
@@ -67,6 +67,8 @@ const createAuth0LogoutUrl = (oidcProvider: OidcProvider, auth0Domain: string, a
             return `https://login.microsoftonline.com/${AD_TENANT_ID}/oauth2/v2.0/logout`
 
         case "Github":
+        case "NextAuth":
```
dsargent, line 70: The one violation of [switch-exhaustiveness-check](https://typescript-eslint.io/rules/switch-exhaustiveness-check/).

I piled these non-ALB cases into the next-auth handler because, frankly we don't really care much about those cases and I'm not sure if they've even worked for a while. Or rather, I remember testing a while back for the `GitHub` case (Neuro AI) but `NextAuth` would mean local dev set-up, and `None` is...who knows.

swensel (approved):
LGTM.

Issue #377, opened 2026-06-01
Word wrapping not working in JSON outputs


dsargent:
Reported by @GPaolo but long term known issue. Couldn't find an existing ticket.

<img width="1186" height="614" alt="Image" src="https://github.com/user-attachments/assets/7126a7e7-99f7-483b-9da3-96745ede4ba5" />

The text doesn't wrap whether the user selects "wrap output" or not.

Issue #367, opened 2026-05-28
Chat redesign


swensel:
# Description

Make chat window look more like ChatGPT.

Here are some ideas from Claude Design:

<img width="1390" height="978" alt="Image" src="https://github.com/user-attachments/assets/fa2849b4-30a6-4e1a-9c08-fc97aac26678" />
<img width="1413" height="966" alt="Image" src="https://github.com/user-attachments/assets/1c5982d4-7a71-4054-8837-90a9ee6cc053" />
<img width="1365" height="963" alt="Image" src="https://github.com/user-attachments/assets/fb217b88-c43f-4091-965a-363ffcc1c6b7" />

swensel:
Current chat window can look pretty ugly:

<img width="660" height="886" alt="Image" src="https://github.com/user-attachments/assets/3e885d6a-fe05-4dbd-b4f0-cd589c453f82" />

<img width="668" height="878" alt="Image" src="https://github.com/user-attachments/assets/ffdbe12f-e7e5-45d8-b489-3d5d6aacb633" />

Issue #347, opened 2026-05-20
Markdown issues for ChatWindow for MAUI


swensel:
# Description

Figure out why table is not showing in ChatWindow for MAUI.

<img width="640" height="710" alt="Image" src="https://github.com/user-attachments/assets/956a1256-922a-4cd4-af53-207f40d7a2f4" />

Perhaps run the ChatWindow through Claude Design as well to get some ideas. It looks like a little old school right now (could be separate).

swensel:
There was an issue with clickable links as well.

<img width="658" height="524" alt="Image" src="https://github.com/user-attachments/assets/700504a2-2a61-4791-ac19-6717e19164c2" />

Issue #337, opened 2026-05-18
Minor concern: Header text size in agent output is too big


deepsaia:
Minor concern: Sometimes, the header text size in agent output is disproportionately bigger.

<img width="569" height="675" alt="Image" src="https://github.com/user-attachments/assets/6b7e95ef-d32c-4e8e-9715-ba478cac2dcc" />

dsargent:
Thanks for the reminder. This is a problem we have in general with our markdown rendering. I agree, needs tweaking.

PR #382, opened 2026-06-02
#367 (and many more) Redesign chat UI

dsargent:
### This PR is a major overhaul of the MAUI Chat UI. 

[Copilot overview.](https://github.com/cognizant-ai-lab/neuro-san-ui/pull/382#pullrequestreview-4452078670)

#### Tickets
#196, #252, #277, #337, #347, #367, #377, #379, [#NSS-898](https://github.com/cognizant-ai-lab/neuro-san-studio/issues/898)

#### Demo

https://github.com/user-attachments/assets/34c4e8ea-96f4-4390-8b69-d897f9fa4fbb

<br />

Before | After
:-------------------------:|:-------------------------:
image |  image
image | image

- Addresses the tickets listed in the title. All of them.
- Fixes the "final answer" handling. Uses `AGENT_FRAMEWORK` as definitive final answer for Neuro-san agents, and handles both `text` and `structure` final answers. Note: adopts a "text if available, if not, display structure" policy, rather than displaying both if both happen to be present. This may change depending on...discussions.
- No longer assumes that `structure` field can only contain error blocks (that contract was changed on the backend). Now, handles `structure` fields containing anything and displays them if they are all that is available for the final answer (fixes "Music Nerd" network interactions).
- Brings the UI up to modern standards, resembling those of products like ChatGPT, Gemini etc.
- The "header bubbles" with user and system avatars are gone. 
- No more "guessing JSON". Everything is formatted with Markdown, and potentially syntax formatted if we detect programming code, but no more than that. **We now assume that the `text` field only contains text, and `structure` only has JSON.**
- The "options" buttons for things like auto-scroll and wrapping are moved behind a discreet icon, which shows a pop-up menu, gaining back real estate.
- Heavyweight accordions are gone in favor of lightweight collapsible sections.
- "Show thinking" now becomes an inline feature; after each response a discreet, collapsible "Show Thinking" section is displayed, that the user can expand at will. This is more in line with modern standards, and more discoverable than the previous menu button, which users tended to miss and never use.
- Removes the "double ReactMarkdown" wrapping that we were doing. Now we only wrap once.
- Fixes all kinds of markdown issues in the output: URLs are now clickable links, lists, headers etc. render properly.
- Fixes a lot of ugly rendering (huge headers) that were due to meddling from `globals.css`.
- Reclaims some of the excess padding and margin so that more text in the chat can be displayed.
- Random greetings are gone; now it's just a generic "How can I help?" with a custom greeting for AND.
- No more "connectivity" display; clutters the chat, takes up real estate, and is redundant: the same info is already depicted visually in the flow graph.
- Fixes, cleans up and consolidates "final answer" and "thinking" handling for legacy agents (Neuro AI opp finder, DMS Chat) to bring their UX on par with the Neuro-san agents.

Tested locally with Neuro AI. The two remaining places that use `ChatCommon` over there are Opp Finder and DMS Chat. Both work with these changes, though Opp Finder is a little strange because the agent info remains at the top of the screen as it never clears the screen, even when changing agents. The solution to that is to migrate Neuro AI to use standard Neuro-san agents.


packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -544,6 +549,109 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+
+            // Add explicit final answer as a new terminal turn
+            updated.push({
+                id: uuid(),
+                role: MessageRole.FinalAnswer,
+                text: finalAnswer,
+            })
+
+            return updated
+        })
```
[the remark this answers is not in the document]
dsargent: Can't happen -- we check earlier and short circuit.

packages/ui-common/components/AgentChat/ChatCommon/Thinking.tsx
```diff
@@ -0,0 +1,77 @@
...
+])
+
+const DEFAULT_AGENT_NAME = "Agent"
+
+const formatTurn = (turn: ConversationTurn) => {
+    const headerLine = `**${turn.agentName ?? DEFAULT_AGENT_NAME}**: ${turn.text}`
+    const structureLine = turn.structure != null ? `\n\`${JSON.stringify(turn.structure, null, 2)}\`` : ""
+
+    return `${headerLine}${structureLine}`
+}
```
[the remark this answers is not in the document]
dsargent: We don't want syntax highlighting in "thinking".

apps/main/public/agent.svg
dsargent, line 1: No longer needed since we don't do the "intro block with avatar" thing any more.

apps/main/styles/globals.css
```diff
@@ -181,93 +181,6 @@ h5 {
     font-weight: 400;
 }
 
-.h1,
```
dsargent, line 184: This was what was causing the huge headers. Bit controversial removing these maybe, but I think it's time we take our own destiny into our hands. I see no reason at all for these huge headers, other than, the corporate stylesheet is designed mostly for Mobile and perhaps for static websites with huge press releases, neither of which is what we do.

apps/main/styles/globals.css
```diff
@@ -181,93 +181,6 @@ h5 {
...
-        font-size: 1.35rem;
-    }
-}
-
-.h6,
-h6 {
-    font-size: 1rem;
-}
-
-p {
```
dsargent, line 266: This is the kind of stuff we should be using MUI for, not action-at-a-distance global styles.

packages/ui-common/components/AgentChat/ChatCommon/AgentConnectivity.tsx
dsargent, line 1: Deleted. We no longer show this: it is redundant with the info in the flow graph.

packages/ui-common/components/AgentChat/ChatCommon/AgentIntro.tsx
dsargent, line 1: We no longer show this in the redesigned chat, so deleted.

packages/ui-common/components/AgentChat/ChatCommon/AgentMetadata.tsx
dsargent, line 1: No longer required or used in the redesigned chat.

packages/ui-common/components/AgentChat/ChatCommon/ChatHistory.tsx
```diff
@@ -1,72 +1,66 @@
...
-    readonly currentUser: string
-    readonly id: string
     readonly messages: BaseMessage[]
-    readonly targetAgent: string
-    readonly userImage: string
+    readonly id: string
 }
 // #endregion: Types
 
+/**
```
dsargent, line 17: Moved to here as it is closely related to this component.

packages/ui-common/components/AgentChat/ChatCommon/ChatHistory.tsx
```diff
@@ -1,72 +1,66 @@
...
-                    } else {
-                        console.warn(`Unrecognized message type in chat history: ${message.type}`)
-                        return null
-                    }
-                }),
-            },
-        ]}
-    />
-)
+export const ChatHistory: FC<ChatHistoryProps> = ({id, messages}) => {
```
dsargent, line 37: This component is heavily simplified due to the redesign. I also realized "chat history is basically just a conversation that happened in the past" so I retooled it to re-use the `Conversation` component, which allowed removing a lot of code and makes things consistent and neat.

packages/ui-common/components/AgentChat/ChatCommon/ChatHistory.tsx
```diff
@@ -1,72 +1,66 @@
...
+            // Pass "true" here to include final answers, which get persisted as agent
+            // messages in chat history.
+            includeAgentMessages={true}
+            shouldWrapOutput={true}
+            turns={turns}
+        />
+    )
+
+    return (
+        <AccordionLite
```
dsargent, line 55: New "lightweight Accordion". See elsewhere in this PR for info.

packages/ui-common/components/AgentChat/ChatCommon/Const.ts
dsargent, line 1: No longer used

packages/ui-common/components/AgentChat/ChatCommon/Conversation.tsx
dsargent, line 1: This is highly simplified and streamlined as part of the redesign.

packages/ui-common/components/AgentChat/ChatCommon/Conversation.tsx
```diff
@@ -1,119 +1,62 @@
...
  */
-const renderTurn = (
-    darkMode: boolean,
-    shadowColor: string,
-    shouldWrapOutput: boolean,
-    turn: ConversationTurn
-): ReactNode => {
-    // extract the parts of the line
-    let repairedJson: string
+const UserTurnBubble = styled(Box)(({theme}) => ({
```
dsargent, line 22: Created a styled component for this.

packages/ui-common/components/AgentChat/ChatCommon/Conversation.tsx
```diff
@@ -1,119 +1,62 @@
...
-        repairedJson = repairedJson.replaceAll(String.raw`\n`, "\n").replaceAll(String.raw`\"`, "'")
-    } catch {
-        // Not valid JSON
-        repairedJson = null
+/**
+ * Format the content of a conversation turn for display. Pretty-print JSON.
+ * @param turn The conversation turn to format.
+ * @returns The formatted content of the turn with JSON fences as appropriate.
+ */
+const formatTurnContent = (turn: ConversationTurn): string => {
```
dsargent, line 41: This replaces the old `renderTurn`, which traces its history back to the first days of Neuro-san UIs. Got rid of the "is this JSON? can I somehow interepret it as JSON?" logic since, nowadays text is text and structure is JSON, so we don't have to do the sniffing.

Also allows us to drop a dependency, `jsonrepair`.

packages/ui-common/components/AgentChat/ChatCommon/Conversation.tsx
```diff
@@ -1,119 +1,62 @@
...
-                fontSize: "large",
-                marginBottom: "1rem",
-                boxShadow: isFinalAnswer
-                    ? `0 6px 16px 0 ${alpha(shadowColor, 0.08)}, 0 3px 6px -4px ${alpha(shadowColor, 0.12)}, 
-                                    0 9px 28px 8px ${alpha(shadowColor, 0.05)}`
-                    : "none",
-            }}
-        />
-    )
+    if (turn.structure != null) {
```
dsargent, line 46: This is so that when agents only return a `structure` (currently, only the "music nerd" family), we display it nicely with syntax formatting.

packages/ui-common/components/AgentChat/ChatCommon/Conversation.tsx
```diff
@@ -1,119 +1,62 @@
...
-                                    id="syntax-highlighter"
-                                    language="json"
-                                    style={darkMode ? atelierDuneDark : a11yLight}
-                                    showLineNumbers={false}
-                                    wrapLongLines={shouldWrapOutput}
-                                >
-                                    {repairedJson}
-                                </SyntaxHighlighter>
-                            ) : (
-                                <ReactMarkdown>{turn.text}</ReactMarkdown>
```
dsargent, line 84: Previously we were wrapping individual nodes in `ReactMarkdown`, _then_ wrapping the whole output _again_ in `ReactMarkdown`. I don't remember why (probably historical reasons/fluffy, poorly-defined-APIs/) but, not healthy, so we no longer do that.

I also wonder if it's responsible for some of the ugly responses we were seeing, with half-escaped JSON and other aberrations.

packages/ui-common/components/AgentChat/ChatCommon/Conversation.tsx
```diff
@@ -124,39 +67,17 @@ export const Conversation: FC<ConversationProps> = ({
                 switch (turn.role) {
                     case MessageRole.User:
                         return [
-                            <UserQueryDisplay
+                            <UserTurnBubble
```
dsargent, line 70: The rendering gets a lot simpler now. Only a handful of cases to worry about.

And of course it's _pure_, no mutation, no side-effects. `turns` in, JSX out.

packages/ui-common/components/AgentChat/ChatCommon/Conversation.tsx
```diff
@@ -186,7 +107,7 @@ export const Conversation: FC<ConversationProps> = ({
                     }
                 }
             }),
-        [currentUser, darkMode, finalAnswerRef, shadowColor, shouldWrapOutput, showThinking, turns, userImage]
+        [includeAgentMessages, turns]
```
dsargent, line 110: The change in deps array tells you everything you need to know.

packages/ui-common/components/AgentChat/ChatCommon/ConversationTurn.ts
```diff
@@ -1,21 +1,25 @@
...
     "Warning" = "Warning",
 }
 
+/**
+ * Represents a single turn in a conversation, which may include a message from the user or agent, or an error
+ */
 export interface ConversationTurn {
-    readonly agentName?: string
     readonly agentDisplayName?: string
-    readonly alwaysShow?: boolean
```
dsargent, line 16: Glad to get rid of this prop from here. It always felt out of place -- policy about when to show what should not reside within each turn. Now, that policy is fully externalized. Turns don't know, or care, under what circumstances they will be shown, a much cleaner design.

packages/ui-common/components/AgentChat/ChatCommon/ConversationTurn.ts
```diff
@@ -1,21 +1,25 @@
...
+    readonly agentName?: string
     readonly id: string
     readonly messageType?: ChatMessageType
     readonly role: MessageRole
-    readonly text: string
+    readonly structure?: Record<string, unknown>
```
dsargent, line 23: Both `text` and `structure` are optional. Still a bit fluffy and vague, but that is the Neuro-san contract we're working with. Agents can return one of the other, or neither.

packages/ui-common/components/AgentChat/ChatCommon/FormattedMarkdown.tsx
```diff
@@ -70,16 +72,18 @@ export const FormattedMarkdown = ({
         <ReactMarkdown
             key={`${hashString(stringToFormat)}-${index}`}
             rehypePlugins={[rehypeRaw, rehypeSlug]}
+            remarkPlugins={[remarkGfm]}
```
dsargent, line 75: This fixes the issues with markdown that Olivier noticed -- #347. With this plugin (incidentally already used in Neuro AI), all markdown renders properly, including lists, links etc.

The specific issue with links is that the agents were returning them as bare literals:

`https://example.com`

rather than proper markdown links

```markdown
[Some link](https://example.com)
```

So nothing was "linkifying" them on our side. `remarkGfm` does that, and a lot more.

packages/ui-common/components/AgentChat/ChatCommon/FormattedMarkdown.tsx
```diff
@@ -70,16 +72,18 @@ export const FormattedMarkdown = ({
...
                     const {children, className, ...rest} = codeProps
                     const match = /language-(?<language>\w+)/u.exec(className || "")
                     return match ? (
                         <SyntaxHighlighter
-                            id={`syntax-highlighter-${match.groups["language"]}`}
                             PreTag="div"
+                            id={`syntax-highlighter-${match.groups["language"]}`}
                             language={match.groups["language"]}
                             style={style}
+                            wrapLongLines={wrapLongLines}
```
dsargent, line 86: Another bug -- we weren't passing `wrapLongLines` to the highlighter here. People have complained about that over time -- "why doesn't the syntax highlighting code obey the line wrap preferene?!" Now it does. See: #377

packages/ui-common/components/AgentChat/ChatCommon/FormattedMarkdown.tsx
```diff
@@ -149,5 +153,5 @@ export const FormattedMarkdown = ({
         formattedOutput.push(concatenatedText)
     }
 
-    return <div id={id}>{formattedOutput}</div>
+    return <Box id={id}>{formattedOutput}</Box>
```
dsargent, line 156: Changed this to `Box` to conform to MUI styling.

packages/ui-common/components/AgentChat/ChatCommon/Greetings.ts
dsargent, line 1: No more cheesy greetings. Happy/sad face.

packages/ui-common/components/AgentChat/ChatCommon/SampleQueries.tsx
```diff
@@ -19,6 +19,9 @@ interface SampleQueriesProps {
 
 /**
  * Render sample queries as clickable chips. Agents may or may not have sample queries defined.
+ * @param disabled Whether the chips should be disabled or not. Chips should be disabled when the agent is
```
dsargent, line 22: Doc was missing

packages/ui-common/components/AgentChat/ChatCommon/SampleQueries.tsx
```diff
@@ -27,7 +30,7 @@ export const SampleQueries: FC<SampleQueriesProps> = ({disabled, handleSend, sam
     sampleQueries?.length > 0 ? (
         <Box
             id="sample-queries-box"
-            sx={{marginTop: "2rem", marginBottom: "1rem"}}
+            sx={{marginTop: "1rem"}}
```
dsargent, line 33: Slightly tweak look for redesign, and use more subdued color for sample query chips.

packages/ui-common/components/AgentChat/ChatCommon/Thinking.tsx
dsargent, line 1: Broke this out into its own component for modularity and ease of reasoning.

packages/ui-common/components/AgentChat/ChatCommon/Thinking.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+interface ThinkingProps {
+    readonly id: string
+    readonly turns: ConversationTurn[]
+}
+
+/**
+ * Set of message types that should be included in the "thinking" section.
+ */
+const THINKING_MESSAGE_TYPES = new Set<ChatMessageType>([
```
dsargent, line 18: We may want to revisit this list at some point.

packages/ui-common/components/AgentChat/ChatCommon/Thinking.tsx
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Component to render the "thinking" section of the chat, which includes all messages from the agent that are
+ * of a type included in THINKING_MESSAGE_TYPES.
+ */
+export const Thinking: FC<ThinkingProps> = ({id, turns}) => {
+    const thinkingText: string = useMemo(() => {
+        // Find start of latest interaction (most recent user prompt)
+        const lastUserTurnIndex = turns.map((t) => t.role).lastIndexOf(MessageRole.User)
+
+        // Only include thinking emitted after latest user prompt
```
dsargent, line 43: See comment. Since the "turns" can go on for multiple human-Neuro-san interactions, we only want to cut off the slice for the most recent interaction to show the relevant "thinking" steps.

packages/ui-common/components/AgentChat/ChatCommon/Thinking.tsx
```diff
@@ -0,0 +1,78 @@
...
+            .slice(lastUserTurnIndex + 1)
+            .filter((turn) => THINKING_MESSAGE_TYPES.has(turn.messageType))
+            .filter((turn) => turn.text?.trim().length > 0 || !isEmpty(turn.structure))
+            .map((turn) => formatTurn(turn))
+            .join("\n\n")
+    }, [turns])
+
+    return (
+        thinkingText.length > 0 && (
+            <AccordionLite
```
dsargent, line 55: Again uses the "lightweight accordion" to allow expanding thinking.

packages/ui-common/components/AgentChat/ChatCommon/UserQueryDisplay.tsx
dsargent, line 1: No longer required. We show the user's query in a styled bubble though.

packages/ui-common/components/AgentChat/Common/Types.ts
```diff
@@ -26,13 +26,21 @@ export const isLegacyAgentType = (agent: string) => {
     return Object.keys(LegacyAgentType).includes(agent)
 }
 
+// Agents which are known to provide a predictable "Final Answer" response
+export const givesFinalAnswer = (agent: CombinedAgentType) => {
```
dsargent, line 30: Hack alert: had to do this for legacy support. The opp finder agents do not return "final answer". 

If/once we migrate those older agents to Neuro-san, we can remove the legacy support code. If we want to spin up newer agents in Node.js for prototyping, FedEx, or rapid development, we can make them follow the Neuro-san protocol (Chat messages etc.) so we won't need exception logic.

packages/ui-common/components/AgentChat/Common/Types.ts
```diff
@@ -26,13 +26,21 @@ export const isLegacyAgentType = (agent: string) => {
...
 export type CombinedAgentType = LegacyAgentType | string
 
 /**
  * Models the error we receive from neuro-san agents.
  */
-export interface AgentErrorProps {
-    error: string
-    traceback?: string
-    tool?: string
+export type AgentErrorProps = {
```
dsargent, line 42: I feel this is better as a `type`, and also add `readonly` for immutability.

packages/ui-common/components/AgentChat/Common/Utils.ts
```diff
@@ -54,17 +55,23 @@ export const chatMessageFromChunk = (chunk: string): ChatMessage | null => {
...
-    if (chatMessageJson && "error" in chatMessageJson) {
-        const agentError: AgentErrorProps = chatMessageJson as AgentErrorProps
-        return (
-            `Error occurred. Error: "${agentError.error}", ` +
-            `traceback: "${agentError?.traceback}", ` +
-            `tool: "${agentError?.tool}"`
-        )
-    } else {
+/**
+ * Type guard to check if a value has the shape of an AgentErrorProps, which indicates it's an error message from
```
dsargent, line 59: Add doc, and harden this. Also make it type-safe so we can get rid of the casting.

packages/ui-common/components/Common/AccordionLite.tsx
dsargent, line 1: I created this to use instead of `MUIAccordion`, which tbh looks a bit outdated. Now `MUIAccordion` is only used in a couple of places within Neuro AI, and we can probably deprecate it entirely.

packages/ui-common/components/MultiAgentAccelerator/AgentNode.tsx
```diff
@@ -210,127 +211,134 @@ export const AgentNode: FC<NodeProps<RFNode<AgentNodeProps>>> = (props: NodeProp
...
-    return (
-        <>
-            <AnimatedNode
-                id={agentId}
-                data-testid={agentId}
-                glowColor={glowColor}
-                isActive={isActiveAgent}
-                onMouseEnter={() => setIsHovered(true)}
-                onMouseLeave={() => setIsHovered(false)}
+        // Also split abbreviation->word boundary (e.g., HTTPServer -> HTTP|Server), but not H|T|T|P
```
dsargent, line 221: Not necessarily acronyms!

packages/ui-common/components/MultiAgentAccelerator/AgentNode.tsx
```diff
@@ -210,127 +211,134 @@ export const AgentNode: FC<NodeProps<RFNode<AgentNodeProps>>> = (props: NodeProp
...
 
     const wrappedAgentName = agentName
         // Allow wrap after underscore
         .replaceAll("_", `_${ZERO_WIDTH_SPACE}`)
         // Allow wrap before capitals in camel/PascalCase (e.g., CustomerSupport -> Customer|Support)
         .replaceAll(/(?<lower>[\da-z])(?<upper>[A-Z])/gu, `$<lower>${ZERO_WIDTH_SPACE}$<upper>`)
-        // Also split acronym->word boundary (e.g., HTTPServer -> HTTP|Server), but not H|T|T|P
-        .replaceAll(/(?<acronym>[A-Z])(?<word>[A-Z][a-z])/gu, `$<acronym>${ZERO_WIDTH_SPACE}$<word>`)
-
-    return (
```
dsargent, line 225: Extracted this to a function for modularity, so the diffs are noisy.

packages/ui-common/components/MultiAgentAccelerator/AgentNode.tsx
```diff
@@ -210,127 +211,134 @@ export const AgentNode: FC<NodeProps<RFNode<AgentNodeProps>>> = (props: NodeProp
...
-                isActive={isActiveAgent}
-                onMouseEnter={() => setIsHovered(true)}
-                onMouseLeave={() => setIsHovered(false)}
+        // Also split abbreviation->word boundary (e.g., HTTPServer -> HTTP|Server), but not H|T|T|P
+        .replaceAll(/(?<abbr>[A-Z])(?<word>[A-Z][a-z])/gu, `$<abbr>${ZERO_WIDTH_SPACE}$<word>`)
+
+    const height = NODE_HEIGHT * (isFrontman ? 1.25 : 1.0)
+    const width = height
+
+    const getEditButton = () => (
```
dsargent, line 227: Extracted this as a render function too

packages/ui-common/components/MultiAgentAccelerator/AgentNode.tsx
```diff
@@ -210,127 +211,134 @@ export const AgentNode: FC<NodeProps<RFNode<AgentNodeProps>>> = (props: NodeProp
...
-            <Tooltip
-                id={`${agentId}-tooltip`}
-                title={agentName}
-                placement="top"
-                disableInteractive
+            </IconButton>
+        </Tooltip>
+    )
+
+    const getNodeName = () => (
```
dsargent, line 254: This is getting complex, so extracted as a render function. Could even be a separate component some day, but felt like overkill for now, especially since I don't see us reusing it anywhere else.

packages/ui-common/components/MultiAgentAccelerator/AgentNode.tsx
```diff
@@ -210,127 +211,134 @@ export const AgentNode: FC<NodeProps<RFNode<AgentNodeProps>>> = (props: NodeProp
...
-                >
-                    {wrappedAgentName}
-                </Typography>
-            </Tooltip>
+                {wrappedAgentName}
+            </Typography>
+        </Tooltip>
+    )
+
+    const getNode = () => (
```
dsargent, line 290: Extracted as a function here

packages/ui-common/components/MultiAgentAccelerator/AgentNode.tsx
dsargent, line 1: A lot of noisy diffs here because I extracted render logic into functions. I'll attach a diff output showing only the diffs not related to those function extractions as it's clearer. 

[diffs.txt](https://github.com/user-attachments/files/28728067/diffs.txt)

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -325,6 +327,24 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
         })()
     }, [neuroSanURL, changeSelectedNetwork])
 
+    useEffect(() => {
```
dsargent, line 330: Lifted this effect, as now we pass the sample queries and other agent-related metadata into `ChatCommon`.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -352,6 +372,12 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
                         a?.origin.localeCompare(b?.origin)
                     )
                     setAgentsInNetwork(agentsInNetworkSorted)
+                    const sampleQueriesTmp = connectivity?.metadata?.["sample_queries"]
```
dsargent, line 375: Do some sanity checking on the results

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -574,15 +600,9 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
         // Reset Agent Network Designer preview
         setAgentsInNetworkDesigner([])
 
-        // Show info popup only once per session
```
dsargent, line 577: I got rid of this annoying popup. I think it's a bit...amateurish and it should be obvious from the UI that users can click the Stop button (or use Escape, which is a standard). If it's not obvious, we should make it more obvious, rather than the annoying popup.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -764,11 +784,13 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
                         id="agent-network-ui"
                         isAwaitingLlm={isAwaitingLlm}
                         key={selectedNetwork ?? "no-network"}
+                        networkDescription={networkDescription}
```
dsargent, line 787: We now pass these in, and only call the API once to get the info.

packages/ui-common/index.ts
```diff
@@ -30,7 +30,6 @@ export * from "./components/Authentication/Auth"
 export * from "./components/Common/Breadcrumbs"
 export * from "./components/Common/ConfirmationModal"
 export * from "./components/Common/Footer"
-export * from "./components/Common/LlmChatOptionsButton"
```
dsargent, line 33: No longer used; now we have the "settings" icon and the dropdown menu, which (1) looks cleaner (IMO) and (2) gives us more real estate for the Chat.

packages/ui-common/package.json
```diff
@@ -14,14 +14,14 @@
         "@langchain/core": "catalog:default",
         "@xyflow/react": "catalog:default",
         "http-status": "catalog:default",
-        "jsonrepair": "3.8.0",
```
dsargent, line 17: No longer needed. We no longer deal with "Schroedinger's JSON". A field is either JSON or it's not.

packages/ui-common/package.json
```diff
@@ -14,14 +14,14 @@
...
         "lodash-es": "catalog:default",
         "notistack": "catalog:default",
         "react-joyride": "catalog:default",
         "react-markdown": "catalog:default",
         "react-syntax-highlighter": "16.1.0",
         "rehype-raw": "catalog:default",
         "rehype-slug": "catalog:default",
+        "remark-gfm": "4.0.1",
```
dsargent, line 24: Added this to help with Markdown rendering.

jest.config.ts
```diff
@@ -22,18 +22,38 @@ import type {Config} from "@jest/types"
 
 // Packages in ESM format that need to be transformed by Jest because they are not distributed in CommonJS format
 const esmPackagesToTransform = [
+    "bail",
```
dsargent, line 25: These extra transpilation items are part of the price of (1) no longer mocking `ReactMarkdown` and (2) adding `remarkGfm`.

The real long-term solution: get off Jest (which is long in the tooth, and struggling to catch up with ESM etc.) and move to `vitest` once we move from Next.js to vite. `vitest` is supposedly many times faster than Jest, as in "blazingly fast", supports ESM natively, and just all around more modern and better DX.

For now, I'm okay with these extra transpilations as long as it doesn't slow down tests noticeably. Which, for me it didn't seem to.

jest.config.ts
```diff
@@ -74,8 +94,8 @@ const config: Config.InitialOptions = {
 
     coverageThreshold: {
         global: {
-            statements: -127,
-            branches: -175,
+            statements: -129,
```
dsargent, line 97: Slight dip in statements, but branches are improved.

I think the dip is because we no longer use `MUIAccordion` in this project so bits of it are now not getting hit by tests.

jest.setup.ts
```diff
@@ -104,40 +103,6 @@ Object.defineProperties(global.HTMLElement.prototype, {
 // eslint-disable-next-line unicorn/prefer-structured-clone
 global.structuredClone = (val: object) => JSON.parse(JSON.stringify(val))
 
-/* Have to mock these up due to https://github.com/remarkjs/react-markdown/issues/635
```
dsargent, line 107: We actually don't have to mock this, and IMO the tests are better allowed it to be "real". The downside -- those extra transpilations in `jest.config.ts`.

package.json
```diff
@@ -60,7 +59,6 @@
         "jest-silent-reporter": "0.6.0",
         "knip": "6.3.1",
         "prettier": "3.8.2",
-        "react": "catalog:default",
```
dsargent, line 63: We're no longer mocking a bunch of stuff in `jest.setup.ts` so that frees us from needing React.js at the root level of the monorepo!

__tests__/common/NetworksListMock.ts
```diff
@@ -73,8 +73,8 @@ export const TEMPORARY_NETWORK: TemporaryNetwork = {
     networkHocon: JSON.stringify(TEMP_NETWORK_AGENT_INFO, null, 2),
 }
 
-export const TEST_TOOL_SPOTIFY = "spotify_tool"
-export const TEST_TOOL_LAST_FM = "last_fm_tool"
+const TEST_TOOL_SPOTIFY = "spotify_tool"
```
dsargent, line 76: No longer used externally in tests. Can't even remember why, but they're not.

packages/ui-common/components/Common/LlmChatOptionsButton.tsx
dsargent, line 1: No longer used. Now we have the "settings" icon & popup menu.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
dsargent, line 1: As you might imagine, ground zero for the redesign. To make matters worse, I also modularized this, so things relating to legacy/Neuro-san agents now get their own functions. Cleaner, safer, easier to reason about, easier for _Copilot_ to reason about, but...the diffs are a nightmare. Apologies.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -169,10 +171,17 @@ export interface ChatCommonProps {
      * to re-supply data that lives outside the IndexedDB slyData store (e.g. localStorage).
      */
     readonly extraSlyData?: Record<string, unknown>
-}
 
-// Key for the chat history, which gets special treatment; always visible even if "show thinking" is off.
-const CHAT_HISTORY_KEY = "chat-history-accordion"
+    /**
+     * Optional description of the network to display in the UI.
```
dsargent, line 176: These are newly passed in from the parent.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -195,8 +204,8 @@ export type ChatCommonHandle = {
 const extractFinalAnswer = (response: string) =>
     /Final Answer: (?<finalAnswerText>.*)/su.exec(response)?.groups?.["finalAnswerText"]
 
-// Maximum number of items to keep in the chat output window
-const MAX_CHAT_OUTPUT_ITEMS = 50
+// Maximum number of turns to save
```
dsargent, line 207: Updated to reflect reality.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -257,9 +267,6 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
     // Whether to wrap output text
     const [shouldWrapOutput, setShouldWrapOutput] = useState<boolean>(true)
 
-    // Keeps a copy of the last AI message so we can highlight it as "final answer"
```
dsargent, line 260: No longer needed. We infer at the end now.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -274,11 +281,8 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
     const updateSlyData = useAgentChatHistoryStore((state) => state.updateSlyData)
     const resetHistory = useAgentChatHistoryStore((state) => state.resetHistory)
 
-    // Ref to the item we think is the Final Answer from the agent
-    const finalAnswerRef = useRef<HTMLDivElement>(null)
-
-    // Track state of "show thinking" toggle
```
dsargent, line 280: No longer relevant -- part of rendering.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -274,11 +281,8 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
     const updateSlyData = useAgentChatHistoryStore((state) => state.updateSlyData)
     const resetHistory = useAgentChatHistoryStore((state) => state.resetHistory)
 
-    // Ref to the item we think is the Final Answer from the agent
```
dsargent, line 277: No longer needed

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -274,11 +281,8 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
-    const finalAnswerRef = useRef<HTMLDivElement>(null)
-
-    // Track state of "show thinking" toggle
-    const [showThinking, setShowThinking] = useState<boolean>(false)
+    // Ref copy of current turns, so we can safely use it in callbacks without worrying about stale closures
+    const turnsRef = useRef<ConversationTurn[]>([])
```
dsargent, line 289: This is a bit of ugliness, the cost of React purity. See where it's used later for an explanation.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -332,63 +336,53 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
         const container = chatOutputRef.current
         if (!container) return
 
-        // Scroll the final answer into view
```
dsargent, line 335: No longer relevant. The final answer now _is_ the answer (just like ChatGPT) and everything else is stuffed into `Thinking`.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -332,63 +336,53 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
             const next = [...current, turn]
-            return next.length > MAX_CHAT_OUTPUT_ITEMS ? next.slice(-MAX_CHAT_OUTPUT_ITEMS) : next
+            return next.length > MAX_TURNS ? next.slice(-MAX_TURNS) : next
         })
     }, [])
 
     // We use this to update the same "turn" as chunks come in from legacy agents
     const legacyTurnIdRef = useRef<string | null>(null)
 
-    const handleChunk = useCallback(
```
dsargent, line 357: Refactored into separate functions for legacy vs Neuro-san.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -397,11 +391,6 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
                 return
             }
 
-            // Keep track of AI messages. The last one is (by definition) the "final answer" from the agents.
```
dsargent, line 400: We no longer need to do this. Once we're done, we simply examine the `turns` array and find the last message of the appropriate type.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -415,36 +404,58 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
-                        text: errorMessage,
-                        alwaysShow: true,
-                    })
-                    succeeded.current = false
-                }
-            } else if (chatMessage?.text?.trim() !== "") {
-                // Not an error, so output it if it has text. The backend sometimes sends messages with no text content,
-                // and we don't want to display those to the user.
-                // Agent name is the last tool in the origin array. If it's not there, use a default name.
+            const errorMessage = checkError(chatMessage.structure)
```
dsargent, line 411: Logic changed. Now we handle any combination of text & structures, and detect error blocks in structures as appropriate. @Noravee this will unblock your PR that's been "in review" forever.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -415,36 +404,58 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
                         ? cleanUpAgentName(chatMessage.origin[chatMessage.origin.length - 1].tool)
-                        : "Agent message"
+                        : "Agent"
                 addTurn({
+                    agentName,
                     id: uuid(),
+                    messageType: chatMessage.type,
                     role: MessageRole.Agent,
-                    agentName,
+                    structure: chatMessage.structure,
```
dsargent, line 434: Here we also save the `structure`.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -415,36 +404,58 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+                }
+            }
+        },
+        [addTurn, targetAgent, updateChatContext, updateSlyData]
+    )
+
+    /**
+     * Handle a chunk of response from the server. Called each time the server streams a chunk.
+     */
+    const handleChunk = useCallback(
```
dsargent, line 449: This is what became of the old `handleChunk` function: it now splits depending what kind of agent we're dealing with.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -415,36 +404,58 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+
+            if (isLegacyAgentType(targetAgent)) {
+                // For legacy agents, we either get plain text or Markdown. Just output it as-is.
+                handleLegacyAgentChunk(chunk)
+            } else {
+                handleNeuroSanAgentChunk(chunk)
             }
         },
-        [onChunkReceived, targetAgent, addTurn, updateSlyData, updateChatContext]
+        [onChunkReceived, targetAgent, handleNeuroSanAgentChunk, handleLegacyAgentChunk]
```
dsargent, line 462: Again, the simplified deps array is a sign of improvement IMO.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -454,8 +465,6 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
         // Reset state, whatever happened during request
         setIsAwaitingLlm(false)
         setChatInput("")
-        lastAIMessage.current = ""
```
dsargent, line 457: These are gone. We do less "tracking as things are streamed" and more "pure rendering after the interaction" now.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -544,6 +552,109 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
         ]
     )
 
+    const getFinalAnswerErrorTurn = () => ({
```
dsargent, line 559: Final answer logic (unfortunately) also splits between legacy agents and Neuro-san agents. 

I also improved final answer handling for legacy agents so now they're on par with Neuro-san. They respond with a simple, concise answer, and everything else is in "Thinking", just like with Neuro-san agents. it does look _much_ better in Neuro AI.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -544,6 +552,109 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+
+            return updated
+        })
+    }, [targetAgent, updateChatHistory])
+
+    /**
+     * Extract the final answer from the turns for a Neuro-san agent. For Neuro-san agents, we expect the final answer
+     * to be the most recent turn messageType === ChatMessageType.AGENT_FRAMEWORK.
+     */
+    const handleFinalAnswerNeuroSanAgent = useCallback(() => {
```
dsargent, line 628: This is the more important logic since we're all about Neuro-san nowadays: defines how we obtain the final answer from the Neuro-san chat.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -544,6 +552,109 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+        // Get current turns snapshot
+        const currentTurns = turnsRef.current
+
+        // Find the most recent turn that is from the agent framework, which should be the one that contains the
```
dsargent, line 632: Modified to use `AGENT_FRAMEWORK`, which satisfies #252

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -544,6 +552,109 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+        if (hasFinalAnswer) {
+            // Update relevant turn to be the final answer
+            setTurns((prev) => prev.map((turn, i) => (i === idx ? {...turn, role: MessageRole.FinalAnswer} : turn)))
+
+            // Save final answer to chat history
+            const finalAnswerContent = finalAnswerTurn.text || JSON.stringify(finalAnswerTurn.structure, null, 2)
+            updateChatHistory(targetAgent, [new AIMessage({content: finalAnswerContent, id: uuid()})])
+        } else {
+            // No final answer found, display error
+            setTurns((prev) => [...prev, getFinalAnswerErrorTurn()])
```
dsargent, line 658: We now display a prominent error if we don't get a final `AGENT_FRAMEWORK` message, which I believe is indeed unexpected, right @Noravee ?

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -567,17 +678,6 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
                 id: uuid(),
                 role: MessageRole.User,
                 text: query,
-                alwaysShow: true,
-            })
-
-            // Add ID block for agent
```
dsargent, line 573: We no longer show these goofy ID blocks. (I can say that because I implemented them in the first place.)

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -587,42 +687,27 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
-                        id: uuid(),
-                        role: MessageRole.FinalAnswer,
-                        text: lastAIMessage.current,
-                    })
-
-                    // Record bot answer in history.
-                    if (currentResponse?.current?.length > 0) {
-                        updateChatHistory(targetAgent, [new AIMessage({content: lastAIMessage.current, id: uuid()})])
+                // Abort condition is handled elsewhere
+                if (!wasAborted) {
```
dsargent, line 699: Cleaned up and simplified this post-stream logic too.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -670,7 +755,7 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
     const shouldEnableRegenerateButton = previousUserQuery && !isAwaitingLlm
 
     // Enable Clear Chat button if not awaiting response and there is chat output to clear
-    const enableClearChatButton = !isAwaitingLlm && turns.length > 0
+    const enableClearChatButton = !isAwaitingLlm && (turns.length > 0 || agentChatHistory?.chatHistory?.length > 0)
```
dsargent, line 762: Slight tweak: we want users to be able to clear chat history even if they have no messages yet in this session.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -680,9 +765,11 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
         resetHistory(targetAgent)
         setPreviousUserQuery("")
         currentResponse.current = ""
-        lastAIMessage.current = ""
     }, [resetHistory, targetAgent])
 
+    const [optionsMenuAnchorEl, setOptionsMenuAnchorEl] = useState<null | HTMLElement>(null)
```
dsargent: For new options menu...probably should move this up top. Will do later but don't want to push commits and have GitHub dump all my annotations.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -748,69 +835,79 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
-                    <VerticalAlignBottomIcon
-                        id="autoscroll-icon"
-                        sx={{color: "var(--bs-white)", fontSize: "0.85rem"}}
-                    />
-                </LlmChatOptionsButton>
-            </Tooltip>
-            <Tooltip
-                id="wrap-tooltip"
-                title={shouldWrapOutput ? "Text wrapping enabled" : "Text wrapping disabled"}
+    const getOptionsMenuButton = () => (
```
dsargent, line 839: For the new options menu (auto-wrap, auto-scroll).

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -748,69 +835,79 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
         </Box>
     )
 
-    const getOptionsButtons = () => (
-        <>
-            <Tooltip
```
dsargent, line 753: BTW "show thinking" is no longer an options menu thing. Users never discovered it and it went unloved and unused.

Instead, each response is now followed by a light accordion with `Show thinking`.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -748,69 +835,79 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
     )
 
-    const agentIntro = (
-        <AgentIntro
-            agentDisplayName={agentDisplayName}
-            customAgentGreetings={customAgentGreetings}
-            key={targetAgent}
-            targetAgent={targetAgent}
-        />
+    const agentGreeting = customAgentGreetings[targetAgent] ?? "Hi, how can I help?"
```
dsargent, line 858: Agent greetings are much simplified. We might want to go back to some form of random greeting just for variety. Not sure right now.

That randomness was a _huge_ pain for React rendering purity, I know that much.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -822,66 +919,75 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
-                    <AgentMetadata
-                        disableQueries={isAwaitingLlm}
-                        handleSend={handleSend}
-                        currentUser={currentUser}
-                        id={`${id}-agent-metadata-display`}
-                        neuroSanURL={neuroSanURL}
-                        targetAgent={targetAgent}
                     />
                 )}
+                <Box sx={{marginBottom: "0.5rem", marginTop: "1rem", color: "var(--bs-gray)"}}>
```
dsargent, line 953: We now just display the agent and description as a "prompt" for the user.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -822,66 +919,75 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+                    includeAgentMessages={!givesFinalAnswer(targetAgent)}
                     shouldWrapOutput={shouldWrapOutput}
                     turns={turns}
-                    userImage={userImage}
                 />
-                {/*For legacy agents, intro goes comes after the conversation as it's a continuous chat stream*/}
-                {isLegacyAgentType(targetAgent) && agentIntro}
+                {!isAwaitingLlm && turns.length > 0 && (
+                    // Only show thinking once streaming is complete
+                    <Thinking
```
dsargent, line 987: "Thinking" is now its own component.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
Noravee: That is correct.

packages/ui-common/components/Common/AccordionLite.tsx
dsargent: Actually, we _could_ even remove `MUIAccordion` in this (admittedly already huge) PR. It won't affect Neuro AI until we release `ui-common` and Neuro AI opts in, at which point it can be modified to use something else 🤔

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -544,6 +552,109 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+
+        // Extract final answer from that turn
+        const finalAnswerTurn = currentTurns[idx]
+        const hasFinalAnswer = finalAnswerTurn.text?.trim().length > 0 || !isEmpty(finalAnswerTurn.structure)
+        if (hasFinalAnswer) {
+            // Update relevant turn to be the final answer
+            setTurns((prev) => prev.map((turn, i) => (i === idx ? {...turn, role: MessageRole.FinalAnswer} : turn)))
+
+            // Save final answer to chat history
+            const finalAnswerContent = finalAnswerTurn.text || JSON.stringify(finalAnswerTurn.structure, null, 2)
```
Noravee, line 654: Does this mean if there is `text` field then anything in `structure` field will not get included in the chat history?

packages/ui-common/components/AgentChat/ChatCommon/Thinking.tsx
dsargent, line 1: It occurred to me later on that `Thinking` could also be just a wrapper around a `Conversation`, but that felt a bit far. Maybe future PR.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
dsargent: Yes, it does.

Now you are probably going to tell me we should display both if present?

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
swensel: Agreed about moving it up.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
swensel: Just wondering, why?

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
swensel: I don't think the random greetings added very much value. I support this change.

swensel (dismissed):
Great PR! Really nice UX improvements. Fixing a lot of issues. Great work @dsargent .

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
Noravee: I think, in general, we would want either `text` or `structure` so choosing one is fine. However, if LLM makes a mistake and there are both `text` and `structure`, I think `structure` field is the important part (since there can only be `structure` if the `"structure_formats` is set).

Maybe check `structure` first and if it exists then use that. What do you think?

dsargent: I see your points. This is a tricky one, as sometimes the useful info could indeed be in `structure`. On the other hand, we want to show the user nice, friendly, readable human-like text, not JSON. Think of MAUI users like ChatGPT users: in general, they don't know about JSON, APIs or any of that. Of course we have highly technical people on our team that use MAUI, but those aren't representative of the users we're targetting. 

Prioritizing `structure` is also risky because, what if `structure` just ends up being administrative stuff like "tokens used", "sustainability metrics" but the actual answer that the user would want to see is in `text`?

Another option...which I don't really like. Always display the `structure` (if present) in a separate area somewhere, like we do with "thinking" and "chat history". Then the `text` would always be displayed as part of the main conversation, as the "final answer", but there's a separate area with the contents of `structure`. That doesn't work so well if the response _only_ has `structure`, though. It's hard to come up with something that works in all cases while being friendly to non-technical users.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
dsargent: I found myself wanting to do it while testing, to get a fresh session. And was annoyed to realize I can't clear chat history without first creating a "sacrificial interaction" as the Clear button was disabled. For me this is in the category of "if a user might conceivably want to, and there's no reason to prevent it, then let them."

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
dsargent: There is an escape hatch at least: if the user opens up the "Thinking" section, everything is in there.

swensel (approved):

Noravee (approved):
lgtm.

Issue #398, opened 2026-06-09
Lines between nodes and Frontman don't line up


dsargent:
<img width="709" height="625" alt="Image" src="https://github.com/user-attachments/assets/366d4d8b-236d-4b42-9c2d-8903a50ffd34" />

PR #399, opened 2026-06-09
#398 Make lines between nodes and Frontman line up properly

dsargent:
Just this.

Before | After
:-------------------------:|:-------------------------:
image|image


swensel (approved):
Nice one. Thanks.

PR #409, opened 2026-06-17
Move Types out of `pages` dir to keep Next.js happy

dsargent:
Having non-pages under pages/ breaks the Next.js build.


apps/main/Types/Types.ts
dsargent, line 1: This was the issue. Next.js build blows up if there are things that aren't JSX pages under the `pages` dir. And it only shows up at production build time, so we didn't know about it until after merge to `main` and in deployment phase.

swensel (approved):
LGTM

Issue #316, opened 2026-05-12
Cache icon suggestions


dsargent:
For both networks + agents, to avoid repeated calls to the LLM to get suggestions.

Could be saved in Zustand store.

May need a way to clear them, or re-roll the dice if the user wants to try different ones.

swensel:
I think these should also be optional, then `OPENAI_API_KEY` won't need to be required - https://github.com/cognizant-ai-lab/neuro-san-ui/pull/330/changes#r3251111748

PR #412, opened 2026-06-17
#316 Cache icon suggestions for agents and networks

dsargent:
Technical PR only.

Uses a simple caching strategy to cache the suggested icons for agents and entire networks. The suggestions are stored in local storage using a standard Zustand store with auto persistence, and cache eviction when the item limit is hit. Other than that, no expiration.

This will save a lot of calls to the icon-suggesting services and thereby will save us a lot of provider tokens.

Another benefit: for cached networks, the custom icons load instantaneously.

Tested locally, confirmed that when the cache is warmed up, no outgoing network requests are made.

I considered using a server-side cache, which is certainly possible, but (1) that would mean either hand-rolling, 3rd party lib/memcached/nginx stuff, or doubling down on Next.js's cache which we don't want to do because we have a troubled relationship with Next.js and (2) this solution is even better, as it avoids the network traffic in the first place as well as saving on LLM tokens. This will also still work when/if we migrate the icon suggestion agents to Neuro-san.

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -95,10 +95,10 @@ const config: Config.InitialOptions = {
 
     coverageThreshold: {
         global: {
-            statements: -126,
-            branches: -158,
-            functions: -34,
-            lines: -97,
+            statements: -116,
+            branches: -156,
+            functions: -29,
+            lines: -89,
         },
     },
 
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -95,10 +95,10 @@ const config: Config.InitialOptions = {
 
     coverageThreshold: {
         global: {
-            statements: -145,
-            branches: -165,
-            functions: -43,
-            lines: -114,
+            statements: -126,
+            branches: -158,
+            functions: -34,
+            lines: -97,
         },
     },
 
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -95,10 +95,10 @@ const config: Config.InitialOptions = {
 
     coverageThreshold: {
         global: {
-            statements: -119,
-            branches: -159,
-            functions: -31,
-            lines: -92,
+            statements: -145,
+            branches: -165,
+            functions: -43,
+            lines: -114,
         },
     },
 
--- a/packages/ui-common/__tests__/components/MultiAgentAccelerator/MultiAgentAccelerator.test.tsx
+++ b/packages/ui-common/__tests__/components/MultiAgentAccelerator/MultiAgentAccelerator.test.tsx
@@ -54,13 +54,8 @@ import {
 } from "../../../components/MultiAgentAccelerator/MultiAgentAccelerator"
 import {SidebarProps} from "../../../components/MultiAgentAccelerator/Sidebar/Sidebar"
 import {MAIN_TOUR_STEPS} from "../../../components/MultiAgentAccelerator/Tour/MainTourSteps"
-import {
-    getAgentFunction,
-    getAgentNetworks,
-    getConnectivity,
-    getNetworkIconSuggestions,
-    testConnection,
-} from "../../../controller/agent/Agent"
+import {getAgentFunction, getAgentNetworks, getConnectivity, testConnection} from "../../../controller/agent/Agent"
+import {getNetworkIconSuggestions} from "../../../controller/agent/IconSuggestions"
 import {ChatMessageType, ChatResponse, ConnectivityInfo} from "../../../generated/neuro-san/NeuroSanClient"
 import {useAgentChatHistoryStore} from "../../../state/ChatHistory"
 import {useSettingsStore} from "../../../state/Settings"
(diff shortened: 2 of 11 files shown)


packages/ui-common/__tests__/components/MultiAgentAccelerator/MultiAgentAccelerator.test.tsx
```diff
@@ -82,6 +77,7 @@ let setSelectedNetwork: (network: string) => void
 jest.mock("next-auth/react")
 
 jest.mock("../../../controller/agent/Agent")
+jest.mock("../../../controller/agent/IconSuggestions")
```
dsargent, line 80: Have to mock this module too now as I split the controllers into separate modules.

packages/ui-common/__tests__/unit/Controller/agent/Agent.test.ts
```diff
@@ -111,7 +111,6 @@ describe("Controller/Agent/sendChatQuery", () => {
     const testQuery = "test query with special characters: !@#$%^&*()_+"
     const testUser = "test user"
     const chatContext = {chat_histories: [] as ChatHistory[]}
-    // TODO: ugly cast due to how openapi-typescript generates object types. What to do here?
```
dsargent, line 114: No longer applicable

packages/ui-common/__tests__/unit/Controller/agent/Agent.test.ts
dsargent, line 1: Moved this to match the path of the module under test

packages/ui-common/__tests__/unit/Controller/agent/IconSuggestions.test.ts
dsargent, line 1: Tests for refactored controller

packages/ui-common/__tests__/unit/Controller/llm/LlmChat.test.ts
dsargent, line 1: Moved to match path of controller under test

packages/ui-common/__tests__/unit/Controller/llm/Providers.test.ts
dsargent, line 1: Moved to match path of controller under test

packages/ui-common/__tests__/unit/State/IconSuggestions.test.ts
dsargent, line 1: New tests for refactored controller

packages/ui-common/controller/agent/Agent.ts
```diff
@@ -100,52 +99,6 @@ export const testConnection = async (url: string): Promise<TestConnectionResult>
     }
 }
 
-/**
```
dsargent, line 103: Moved these to their own controller module, `IconSuggestions`, since they aren't directly related to the abstract agent stuff.

packages/ui-common/controller/agent/Agent.ts
```diff
@@ -308,7 +261,7 @@ export const sendNetworkDesignerUpdate = async (
     await sendChatQuery(
         url,
         signal,
-        // Shouldn't have to pass a user message, but API behaves different without it
+        // Shouldn't have to pass a user message, but API behaves differently without it
```
dsargent, line 264: Grammar!

packages/ui-common/controller/agent/IconSuggestions.ts
dsargent, line 1: Relocated here from `Agent.ts`. Helps keep that module less cluttered and keeps things modular and easier to understand IMO.

packages/ui-common/controller/agent/IconSuggestions.ts
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * Utility function to get icon suggestions from the cache or fetch from the server if not available in the cache.
+ *
+ * @param keys - The list of keys (agent names or network names) to get suggestions for.
+ * @param storeKey - The key in the cache (zustand store) to save the suggestions under.
+ * @param fetchSuggestions - Fetch function to query the LLM for suggestions
+ * @returns A promise that resolves to a record mapping the keys to suggested MUI icon names.
+ */
+const getSuggestions = async (
```
dsargent, line 43: New write-through-cache for both types of suggestions (agent icons and network icons).

packages/ui-common/state/IconSuggestions.ts
dsargent, line 1: New Zustand store for the suggestions cache.

jest.config.ts
```diff
@@ -95,10 +95,10 @@ const config: Config.InitialOptions = {
 
     coverageThreshold: {
         global: {
-            statements: -119,
-            branches: -159,
-            functions: -31,
-            lines: -92,
+            statements: -116,
```
dsargent, line 98: Onward and upward 🎉

swensel (approved):
LGTM. You are on 🔥 with the PRs.

Issue #422, opened 2026-06-19
Settings UI gets stuck if errors accessing branding service


dsargent:
Spinners remain and user is stuck with the company input.

PR #423, opened 2026-06-19
#422 Improve error handling when Branding service call fails

dsargent:
Previously if the service failed for whatever reason, the dialog got stuck in a "spinner" state.

### DIFF
--- a/packages/ui-common/components/Settings/SettingsDialog.tsx
+++ b/packages/ui-common/components/Settings/SettingsDialog.tsx
@@ -152,12 +152,6 @@ export const SettingsDialog: FC<SettingsDialogProps> = ({id, isOpen, logoService
     const handleBrandingApply = async () => {
         setIsBrandingApplying(true)
 
-        updateSettings({
-            branding: {
-                customer: customerInput,
-            },
-        })
-
         let brandingSuggestions
         try {
             brandingSuggestions = await getBrandingSuggestions(customerInput)
@@ -170,7 +164,15 @@ export const SettingsDialog: FC<SettingsDialogProps> = ({id, isOpen, logoService
                     "branding service."
             )
             return
+        } finally {
+            setIsBrandingApplying(false)
         }
+
+        updateSettings({
+            branding: {
+                customer: customerInput,
+            },
+        })
 
         updateSettings({
             appearance: {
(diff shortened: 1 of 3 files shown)


packages/ui-common/components/Settings/SettingsDialog.tsx
```diff
@@ -152,15 +152,10 @@ export const SettingsDialog: FC<SettingsDialogProps> = ({id, isOpen, logoService
...
-        updateSettings({
-            branding: {
-                customer: customerInput,
-            },
-        })
-
         let brandingSuggestions
         try {
             brandingSuggestions = await getBrandingSuggestions(customerInput)
+            setIsBrandingApplying(false)
```
dsargent: Reset state on error

packages/ui-common/components/Settings/SettingsDialog.tsx
```diff
@@ -172,6 +167,12 @@ export const SettingsDialog: FC<SettingsDialogProps> = ({id, isOpen, logoService
             return
         }
 
+        updateSettings({
```
dsargent, line 151: Moved this down here; we shouldn't do this until we're sure the service call succeeded.

packages/ui-common/components/Settings/SettingsDialog.tsx
swensel: Shouldn't reset state on error be in the `catch`?

swensel: Your first [commit](https://github.com/cognizant-ai-lab/neuro-san-ui/pull/423/changes/6edc58fbd7251c081c2805159f4ee08be145a6c3) (previous to Devin), had a `finally`, which I think makes more sense? I'm fine with the `catch`, but just wondering why this will work as is because `setIsBrandingApplying(false)` will execute after a successful `await getBrandingSuggestions(customerInput)`?

dsargent: Fixed logic + extract guts to separate function + change `warn` to `error`.

packages/ui-common/components/Settings/SettingsDialog.tsx
```diff
@@ -239,9 +221,32 @@ export const SettingsDialog: FC<SettingsDialogProps> = ({id, isOpen, logoService
...
-        setIsBrandingApplying(false)
+    /**
+     * Handle applying branding based on customer input. Calls backend to get colors then updates settings store.
+     */
+    const handleBrandingApply = async () => {
+        setIsBrandingApplying(true)
+
+        let brandingSuggestions
+        try {
+            brandingSuggestions = await getBrandingSuggestions(customerInput)
```
swensel: This could be `let brandingSuggestions = await getBrandingSuggestions(customerInput)`. `brandingSuggestions` is only referred to in the `try` block.

swensel (approved):
LGTM.

Issue #396, opened 2026-06-08
Migrate to vitest: Phase I


dsargent:
More modern/faster/better than Jest, better ESM support (avoids transpilation step).

PR #394, opened 2026-06-06
PR feedback follow up (#393).

swensel:
# Description

Implement PR feedback from https://github.com/cognizant-ai-lab/neuro-san-ui/pull/357, https://github.com/cognizant-ai-lab/neuro-san-ui/pull/321, and https://github.com/cognizant-ai-lab/neuro-san-ui/pull/384.

PR #384, opened 2026-06-03
#389 - [FedEx] Import JSON files into MAUI

swensel:
# Description

#389 - [FedEx] Import HOCON files into MAUI.

# Screenshots

Screenshot 2026-06-26 at 3 59 53 PM

Screenshot 2026-06-26 at 4 00 02 PM

Screenshot 2026-06-26 at 3 55 08 PM

Screenshot 2026-06-26 at 3 55 19 PM

Screenshot 2026-06-26 at 3 55 44 PM

Screenshot 2026-06-26 at 3 55 52 PM

Screenshot 2026-06-26 at 3 57 32 PM

Screenshot 2026-06-26 at 3 55 19 PM

Screenshot 2026-06-26 at 3 57 49 PM

Screenshot 2026-06-26 at 3 57 59 PM

Screenshot 2026-06-26 at 3 58 19 PM

Screenshot 2026-06-26 at 3 58 43 PM

Screenshot 2026-06-26 at 3 58 56 PM

Screenshot 2026-06-26 at 3 59 02 PM

Screenshot 2026-06-26 at 3 59 11 PM

Screenshot 2026-06-26 at 3 59 26 PM

Screenshot 2026-06-26 at 3 59 33 PM

Screenshot 2026-06-26 at 3 59 41 PM

## Invalid Parsing Case

Screenshot 2026-06-26 at 3 56 56 PM

PR #443, opened 2026-06-27
#396 First phase of vitest migration

dsargent:
ℹ️ Technical PR only, no user-facing changes.

First part of phased migration to `vitest`. We have some duplication (scripts, `strictMocks` etc.) but that is temporary and will go away once we're fully migrated.

1) Adds `vitest` dependencies including `vitest` itself and a couple of associated ones
2) Adds a config for `vitest` based on our existing `jest` config
1) Similarly adds a `vitest.setup.ts` which is almost identical to the corresponding `jest` one.
3) Migrates a few very simple unit tests to `vitest` as a demo.
4) Puts in place the infrastructure so that `vitest` and `jest` can cohabit within the project: 
5) Update associated scripts and goals to support both.

ℹ️ Note that `yarn update-coverage` now has two flavors, depending which one you want to update. But it's moot for now, temporarily, as the source of truth for the thresholds is in `build_scripts/RunCombinedTests.sh`.

⚠️ `yarn test` now runs both, merges the coverage reports, and asserts that _combined_ coverage across both `jest` and `vitest` doesn't fall.

So as we migrate tests, we update the corresponding config (see Files Changed) and they will start running inside `vitest` rather than `jest`.

Some rough edges: the two tools don't measure coverage exactly the same, even when using the same `istanbul` engine, so there is a "little daylight" between the numbers. That is fine and expected, as long as the trend is in the right direction. But this means that the numbers may not match up exactly right: we may move a test from `jest` to `vitest` and the coverage changes slightly, in either direction.

Things may get bumpy during the transition, and it will be a little awkward juggling them in parallel, which should help motivate us to migrate the tests over more quickly 🙂 

Next steps:

1) Migrate the rest of the tests including RTL (component) tests
2) Once migration is done, remove migration scaffolding and all Jest-related items.


.github/workflows/test.yml
```diff
@@ -66,7 +66,7 @@ jobs:
               # v1.0.7
               uses: cognizant-ai-lab/build-common/actions/setup-node-env@3ead7f681f92044709ffc3a533f41c3f48230cfd
               with:
-                  node-version: ${{ vars.NODEJS_VERSION }}
+                  node-version: "22.17.1"
```
dsargent, line 69: @donn-leaf if you want to address this in a separate PR please feel free.

This is the pinned magic version that allows us to import TS in jest config files. Higher Node.js versions break that. This is a regression that happened during the migration to `build-common`.

__tests__/common/vitest/strictMocks.ts
dsargent, line 1: Replicated version for vitest. Once migration is completed we will have only one.

__tests__/common/vitest/TestUtils.ts
dsargent, line 1: Replicated version for vitest. Once migration is completed we will have only one.

build_scripts/vitest/UpdateCoverage.sh
dsargent, line 1: Replicated version for vitest. Once migration is completed we will have only one.

build_scripts/CommitCheck.sh
```diff
@@ -50,4 +50,4 @@ run_check "tsc" yarn tsc
 run_check "knip" yarn knip
 run_check "prettier" yarn check-format
 run_check "eslint" yarn lint --max-warnings 0
-run_check "test" jest --config ./jest_quiet.config.ts
```
dsargent, line 53: Rather than bothering with a separate "quite config" just for vitest, just use 'yarn test`. This script seems to hide the spew anyway.

build_scripts/RunCombinedTests.sh
dsargent, line 1: New hacky script, migration only: combines coverage from jest + vitest and fails if _overall_ coverage drops.

It's not an exact science though: you'd thinking moving a test from jest to vitest would be a net zero, but it's not always the case. Even with both tools configured to use the `istanbul` test coverage engine, there are very often deltas due to Jest running everything through `babel`, as of course Jest doesn't support ESM.

This also means that with vitest, we get more fidelity in the test coverage stats, since it's at the Typescript level, rather than being run through the transpiler first. Not to mention a lot faster due to not having the extra step.

build_scripts/RunCombinedTests.sh
```diff
@@ -0,0 +1,102 @@
...
+# Update these as coverage (hopefully) improves. The build will fail if the combined coverage report exceeds
+# these thresholds.
+# Line coverage matches the original Jest-only baseline, but statement/branch/function
+# counters differ slightly due to Vitest/Vite Istanbul instrumentation.
+# For reference, the jest.config numbers prior to any migration were:
+# statements: -112,
+# branches: -152,
+# functions: -27,
+# lines: -85,
+STATEMENTS=132
```
dsargent, line 22: @swensel this is the important part fo you. Updating coverage in `jest.config.ts` -> during migration, update it here instead. Not automated, sorry. Not worth it as it's a temporary thing for migration.

build_scripts/RunCombinedTests.sh
```diff
@@ -0,0 +1,102 @@
...
+set -o nounset
+set -o pipefail
+
+# Combined Jest+Vitest/NYC baseline during migration.
+# Update these as coverage (hopefully) improves. The build will fail if the combined coverage report exceeds
+# these thresholds.
+# Line coverage matches the original Jest-only baseline, but statement/branch/function
+# counters differ slightly due to Vitest/Vite Istanbul instrumentation.
+# For reference, the jest.config numbers prior to any migration were:
+# statements: -112,
```
dsargent, line 18: The new values are close but not exactly the same, probably due to different coverage reporting strategies between the tools.

The "uncovered lines" threshold agrees _exactly_ between Jest and Vitest, reassuringly, which it probably should: that should be stable across tools, however they measure everything else.

packages/dev-common/Configs/eslint.config.mjs
```diff
@@ -283,7 +283,13 @@ export default defineConfig([
                 {
                     includeInternal: true,
                     includeTypes: true,
-                    devDependencies: ["**/__tests__/**", "jest*.*", "**/eslint.config.mjs", "**/knip.config.ts"],
+                    devDependencies: [
+                        "**/__tests__/**",
+                        "jest*.*",
+                        "vitest*.*",
```
dsargent, line 289: Add vitest here

packages/dev-common/package.json
```diff
@@ -18,6 +18,7 @@
         "eslint-plugin-react-hooks": "*",
         "eslint-plugin-testing-library": "*",
         "eslint-plugin-unicorn": "*",
+        "eslint-plugin-vitest ": "*",
```
dsargent: Add linter plugin for vitest here.

TBH I didn't focus on linting for this initial PR. It's already a lot.

packages/ui-common/__tests__/unit/Utils/Authentication.test.ts
dsargent, line 1: I migrated the tests under `Utils` to vitest as a pilot/demo/spearhead, and so we can see how it looks.

Notably, these are "just unit tests", not RTL/React, so that will be a whole other thing.

packages/ui-common/__tests__/unit/Utils/Authentication.test.ts
```diff
@@ -16,16 +16,20 @@ limitations under the License.
...
 import * as BrowserNavigation from "../../../utils/BrowserNavigation"
-import {navigateToUrl} from "../../../utils/BrowserNavigation"
 
-// Mock the next-auth/react module
-jest.mock("next-auth/react")
+vi.mock("next-auth/react")
+
+const mockedSignOut = vi.mocked(signOut)
+
+let navigateToUrlSpy: MockInstance<typeof BrowserNavigation.navigateToUrl>
```
dsargent, line 32: Not convinced about this. It is better in that it preserves type safety and avoids the ugly unsafe casting, _but_...apparently URL navigation is impossible in JSDom anyway, so not sure exactly what we're testing here...

packages/ui-common/__tests__/unit/Utils/BrowserNavigation.test.ts
dsargent, line 1: Another migrated test

packages/ui-common/__tests__/unit/Utils/zIndexLayers.test.ts
```diff
@@ -1,4 +1,6 @@
 import {Theme} from "@mui/material/styles"
+// eslint-disable-next-line no-shadow
```
dsargent, line 2: We have to live with these "no-shadow" directives everywhere during migration.
The issue is that both Jest and Vitest want to add a whole bunch of globals to the global namespace, for things like `expect`, `it`, `beforeEach` etc. That's how you can get away with using those in tests without importing them.

The downside of that: when you have _both_ tools active, they fight over the global namespace. So the best way I found to deal with that:

1) Let Jest control the global namespace for now
2) Tell Vitest not to use globals
3) Manually import Vitest's identifiers in each test.

It's ugly, but the alternatives are worse IMO.

Once we're done migrating, this all goes away, we hand the global namespace over to Vitest, and we can `expect`, `describe` etc. with impunity, without importing anything, just as we do today.

packages/ui-common/package.json
```diff
@@ -47,7 +47,8 @@
         "openapi-typescript": "7.8.0",
         "react": "catalog:default",
         "react-dom": "catalog:default",
-        "typescript": "catalog:default"
+        "typescript": "catalog:default",
+        "vitest": "catalog:default"
```
dsargent, line 51: Add dependency for vitest

.gitignore
```diff
@@ -7,7 +7,7 @@
 **/tsconfig.tsbuildinfo
 
 # Coverage directory used by tools like lcov
-**/coverage/
+**/coverage*/
```
dsargent, line 10: During migration, we generate several coverage dirs, so exclude them all

.yarnrc.yml
```diff
@@ -33,11 +33,14 @@ catalogs:
         "lodash-es": "4.18.1"
         "next-auth": "5.0.0-beta.30"
         "notistack": "3.0.2"
+        "nyc": "18.0.0"
```
dsargent, line 36: Used for merging coverage

.yarnrc.yml
```diff
@@ -33,11 +33,14 @@ catalogs:
...
         "react-joyride": "3.1.0"
         "react-markdown": "10.1.0"
         "rehype-raw": "7.0.0"
         "rehype-slug": "6.0.0"
         "typescript": "5.9.3"
+        "vitest": "4.1.9"
```
dsargent, line 42: vitest deps

jest.config.ts
```diff
@@ -20,6 +20,8 @@ limitations under the License.
 // https://nextjs.org/docs/pages/building-your-application/optimizing/testing#jest-and-react-testing-library
 import type {Config} from "@jest/types"
 
+import {TESTS_MIGRATED_TO_VITEST} from "./vitest_migration"
```
dsargent, line 23: This allows jest to only runs tests that haven't yet been migrated to Vitest.

But it also breaks the build if we don't use the magic Node.js version (see changes in `test.yml`.)

jest.config.ts
```diff
@@ -76,7 +78,7 @@ const config: Config.InitialOptions = {
     verbose: false,
     setupFilesAfterEnv: ["<rootDir>/jest.setup.ts"],
     testEnvironment: "jsdom",
-    collectCoverage: true,
+    collectCoverage: false,
```
dsargent, line 81: Disable this: will speed up jest tests and during migration we should only pay attention to the _combined_ coverage across both tools.

jest.config.ts
```diff
@@ -93,12 +96,14 @@ const config: Config.InitialOptions = {
 
     coverageReporters: ["text-summary"],
 
+    // These values are meaningless during migration to vitest as we are running tests in both jest and vitest.
```
dsargent, line 99: See comment!

jest.config.ts
```diff
@@ -114,7 +119,13 @@ const config: Config.InitialOptions = {
     testMatch: ["<rootDir>/**/__tests__/**/*.(test).{ts,tsx}"],
 
     // Exclude these files from Jest scanning
-    testPathIgnorePatterns: ["<rootDir>/node_modules/", "<rootDir>/dist", "<rootDir>/.next/"],
+    testPathIgnorePatterns: [
+        "<rootDir>/node_modules/",
+        "<rootDir>/dist",
+        "<rootDir>/.next/",
+        // Exclude tests that have been migrated to Vitest.
+        ...TESTS_MIGRATED_TO_VITEST.map((testPath) => `<rootDir>/${testPath}`),
```
dsargent, line 127: This excludes tests that have been migrated to Vitest.

knip.config.ts
```diff
@@ -52,6 +52,9 @@ const config: KnipConfig = {
         // Used internally by eslint
         "globals",
 
+        // Used by RunCombinedTests.sh during Jest-to-Vitest migration
```
dsargent, line 55: We can remove this after migration (unless we happen to want to use `nyc` for something else, long term.)

package.json
```diff
@@ -37,6 +40,8 @@
         "@types/dom-speech-recognition": "0.0.6",
         "@types/jest": "catalog:default",
         "@types/node": "catalog:default",
+        "@vitejs/plugin-react": "^6.0.3",
```
dsargent, line 43: Various Vitest deps added to dev deps.

package.json
```diff
@@ -22,8 +22,11 @@
         "lint-rules": "./build_scripts/eslint_list_rules.sh",
         "lint-summary": "eslint --format summary",
         "start": "yarn workspace @cognizant-ai-lab/neuro-san-web start",
-        "test": "jest --config ./jest.config.ts",
-        "update-coverage": "./build_scripts/UpdateCoverage.sh"
+        "test": "./build_scripts/RunCombinedTests.sh",
```
dsargent, line 25: Note changed goals!

`test` -- run everything, assert on combined coverage. Use this.
`test:jest` -- run only jest tests (why?!)
`test:vitest` -- run only vitest test (why?!)
`update-coverage:jest` -- update coverage only for jest in `jest.config.ts` (pointless?)
`update-coverage:vitest` -- update coverage only for Vitest in `vitest.config.ts` (pointless?)

vitest.config.ts
dsargent, line 1: Config for Vitest.

vitest.config.ts
```diff
@@ -0,0 +1,44 @@
...
+import {defineConfig} from "vitest/config"
+
+import {TESTS_MIGRATED_TO_VITEST} from "./vitest_migration"
+
+export default defineConfig({
+    plugins: [react()],
+    test: {
+        coverage: {
+            reportsDirectory: "coverage-vitest",
+            enabled: false,
```
dsargent, line 11: Coverage disabled by default for both tools. The `RunCombinedTests.sh` script enables it and auto merges.

vitest.config.ts
```diff
@@ -0,0 +1,44 @@
...
+                statements: -13,
+                branches: -7,
+                functions: -9,
+                lines: -9,
+            },
+        },
+        environment: "jsdom",
+        exclude: ["node_modules", "dist", ".next", "coverage"],
+        globals: false,
+        include: TESTS_MIGRATED_TO_VITEST,
```
dsargent, line 41: We only run tests that have been migrated to Vitest.

vitest.config.ts
```diff
@@ -0,0 +1,44 @@
...
+                "**/coverage/**",
+                "**/dist/**",
+                "**/generated/**",
+                "**/knip.config.ts",
+                "**/next-env.d.ts",
+                "**/next.config.ts",
+                "apps/main/app/api/auth/[...nextauth]/route.ts",
+            ],
+            // TODO: Uncomment this once we have migrated all tests to vitest.
+            // include: ["**/*.{js,jsx,ts,tsx}"],
```
dsargent, line 24: This `include` is rather important. It defines the "universe" for coverage -- what should be included. Since only a handful of tests are initially migrated, it doesn't make sense for Vitest to consider the "universe" (denominator!) for coverage to be "the entire source tree".

Once we are migrated, we will expand this is to cover the entire source tree, like currently do for Jest.

In DB parlance this is an "inner join" but with that line uncommented you'd get an "outer join".

vitest.setup.ts
dsargent, line 1: Cloned from the Jest set-up.

Post migration, there's probably a lot of clean-up we can do in here. Crutches that Jest needed but Vitest doesn't.

vitest_migration.ts
dsargent, line 1: Opt-in list of tests that have been migrated to Vitest. As we migrate tests, add them here.

Note: no wildcards. They have to be named explicitly. Wildcards get messy in the config. Hacky, but it's temporary for migration so I don't really care.

yarn.lock
dsargent, line 1: I considered regen'ing this, but, the only dep changes are dev deps, so maybe not?

Or perhaps after migration is complete, to trim it down. 👍🏻

build_scripts/RunCombinedTests.sh
```diff
@@ -0,0 +1,102 @@
...
+  --coverage.thresholds.branches=-999999 \
+  --coverage.thresholds.functions=-999999 \
+  --coverage.thresholds.lines=-999999
+
+mkdir -p coverage-combined-input
+
+cp coverage-jest/coverage-final.json coverage-combined-input/jest.json
+cp coverage-vitest/coverage-final.json coverage-combined-input/vitest.json
+
+yarn nyc merge coverage-combined-input coverage-combined/coverage-final.json 1>/dev/null
```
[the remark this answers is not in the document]
dsargent: It does (or else the CI build wouldn't have passed.)

packages/dev-common/package.json
```diff
@@ -18,6 +18,7 @@
         "eslint-plugin-react-hooks": "*",
         "eslint-plugin-testing-library": "*",
         "eslint-plugin-unicorn": "*",
+        "eslint-plugin-vitest": "*",
```
dsargent, line 21: Add linter plugin for vitest here.

TBH I didn't focus on linting for this initial PR. It's already a lot.

.github/workflows/test.yml
swensel: Also migrating to `vitest` will help address this as we move away from `jest`.

swensel (dismissed):
LGTM. I'd like to get https://github.com/cognizant-ai-lab/neuro-san-ui/pull/384 in first, if possible.

There will also be an increase in test coverage with https://github.com/cognizant-ai-lab/neuro-san-ui/pull/394. Based on the coverage changing with this `vitest` PR, the improvement will be a little less clear, but it's OK.

swensel (approved):

PR #445, opened 2026-06-28
#444 Vitest migration Phase 2

dsargent:
More migrations, this time including the slightly more challenging JSX tests.

To quote Copilot,

> This does not necessarily mean coverage got worse in a meaningful way. It means your migration baseline changed as more code is measured by Vitest’s instrumentation.
During migration, your combined thresholds are a moving baseline as files cross from Jest to Vitest. That is unfortunate, but expected.

> Compare old vs new line coverage and spot-check LCOV for the migrated components. The line delta of +1 is the strongest signal here. I would not use statement/function/branch counter deltas as a real quality signal until the migration is complete.


build_scripts/RunCombinedTests.sh
```diff
@@ -19,10 +19,10 @@ set -o pipefail
 # branches: -152,
 # functions: -27,
 # lines: -85,
-STATEMENTS=132
-BRANCHES=170
-FUNCTIONS=39
-MAX_LINES=85
+STATEMENTS=281
```
dsargent, line 22: It _looks_ bad but Copilot assures me (gaslights?) that this is normal. I'm making sure that the test count stays constant as we migrate. But there are several factors meaning that the numbers don't quite add up so it looks like we're losing coverage:

1) Jest uses (and requires) babel transpilation, so it's measuring something different for coverage: TS vs JS.
2) Overlapping coverage: some tests hit multiple modules, and some modules are hit by multiple tests. That messes up the numbers when we migrate only some of the tests that hit a particular module.
3) Even when forced to use the same coverage engine ("istanbul"), Jest and Vitest measure coverage in different ways, according to Copilot.

The `MAX_LINES` is stable (well, within 1 line!) which gives me _some_ reassurance. Once we are done migrating, I will compare the overall Vitest numbers with the pre-migration Jest numbers as a cross check.

packages/dev-common/Configs/eslint.config.mjs
```diff
@@ -555,6 +557,12 @@ export default defineConfig([
             "react/no-multi-comp": "off",
         },
     },
+    {
```
dsargent, line 560: Exclude this Jest linter rule from the Vitest tests: doesn't apply. Note how we use a `files` filter block for ESLint, and feed it the list of migrated tests (which I had to change to JS for this purpose since ESLint requires it). Single point of truth for list of migrated tests. I try to do things the right way even when it's a temporary migration thing.

packages/dev-common/Configs/eslint.config.mjs
```diff
@@ -15,14 +15,16 @@ import eslintPluginUnicorn from "eslint-plugin-unicorn"
...
+
 /** @type {import("eslint").Linter.Config[]} */
 export default defineConfig([
     globalIgnores([
         "**/*.d.mts",
         "**/*.d.ts",
         "**/.next",
         "**/babel.jest.config.cjs",
-        "**/coverage",
+        "**/coverage*/",
```
dsargent, line 27: We now (temporarily) have several `coverage-foo` dirs. Will clean up after migration.

packages/ui-common/__tests__/components/Authentication/Auth.test.tsx
```diff
@@ -28,7 +30,8 @@ const AUTH_ELEMENT = (
     </Auth>
 )
```
dsargent, line 32: This was all kinds of messed up. We were silently hooking into a mock from a different workspace (`apps/main` vs `packages-ui-common`) due to the broad search patterns in `jest.config.ts`.

Now, ~homie~ Vitest don't play that, so now we explicitly mock the library here.

packages/ui-common/__tests__/components/ChatBot/ChatBot.test.tsx
```diff
@@ -31,17 +33,17 @@ jest.mock("../../../components/AgentChat/ChatCommon/ChatCommon", () => ({
 }))
 
 // Mock MUI theming
-jest.mock("@mui/material/styles", () => ({
-    ...jest.requireActual("@mui/material/styles"),
-    useColorScheme: jest.fn(),
+vi.mock("@mui/material/styles", async () => ({
+    ...(await vi.importActual("@mui/material/styles")),
```
dsargent, line 37: Some slight differences in Jest vs Vitest. Vitest seems stricter about the shape matching. Also they use `importActual` rather than `requireActual` 🤷🏻 

dding the `async` + `await` here makes tsc happy at least.

packages/ui-common/__tests__/components/ChatBot/ChatBot.test.tsx
```diff
@@ -31,17 +33,17 @@ jest.mock("../../../components/AgentChat/ChatCommon/ChatCommon", () => ({
...
 }))
 
-jest.mock("@mui/material/Grow", () => ({
+vi.mock("@mui/material/Grow", () => ({
     __esModule: true,
     default: ({children, in: inProp}: {children: unknown; in: boolean}) => (inProp ? children : null),
 }))
 
-const mockUseAuthentication = useAuthentication as jest.Mock
+const mockUseAuthentication = vi.mocked(useAuthentication)
```
dsargent, line 46: Note the safer, strongly-typed mock. This way if the test ever tries to do something with `mockUseAuthentication` that is not in line with the real `useAuthentication`, tsc will complain, which is great 🎉 In Jest, we were casting all over the place and doing careless things like that everywhere, since by casting we "silence" tsc and it cannot help us.

build_scripts/RunCombinedTests.sh
```diff
@@ -19,10 +19,10 @@ set -o pipefail
...
+BRANCHES=364
+FUNCTIONS=120
+MAX_LINES=86
```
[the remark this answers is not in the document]
dsargent: We're hoping this is okay...

packages/dev-common/Configs/eslint.config.mjs
```diff
@@ -555,6 +557,12 @@ export default defineConfig([
             "react/no-multi-comp": "off",
         },
     },
+    {
+        files: TESTS_MIGRATED_TO_VITEST,
+        rules: {
+            "jest/expect-expect": "off",
+        },
+    },
```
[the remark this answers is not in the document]
dsargent: It is working fine, maybe because we always runs tests from the repo root.

packages/ui-common/__tests__/components/ChatBot/ChatBot.test.tsx
```diff
@@ -31,17 +33,29 @@ jest.mock("../../../components/AgentChat/ChatCommon/ChatCommon", () => ({
...
+
+const getUseColorSchemeMockReturnValue = (mode: "light" | "dark", systemMode: "light" | "dark" = "light") =>
```
dsargent, line 48: Extracted this function to avoid copy-pasting the fake `useColorScheme` result across tests.

packages/ui-common/__tests__/components/ChatBot/ChatBot.test.tsx
```diff
@@ -31,17 +33,29 @@ jest.mock("../../../components/AgentChat/ChatCommon/ChatCommon", () => ({
...
+    ({
+        allColorSchemes: ["light", "dark"],
+        mode,
+        systemMode,
+        lightColorScheme: "light",
+        darkColorScheme: "light",
+        colorScheme: "light",
+        setMode: vi.fn(),
+        setColorScheme: vi.fn(),
+    }) satisfies ReturnType<typeof useColorScheme>
```
dsargent, line 58: Neat way to avoid specifying what the actual return type _is_, which is good because in this case the hook returns internal MUI "system" types that we don't want to know about or depend on.

packages/ui-common/__tests__/components/ChatBot/ChatBot.test.tsx
```diff
@@ -59,12 +73,12 @@ describe("ChatBot", () => {
             data: {
                 user: {
                     name: "Test User",
+                    image: "",
```
dsargent, line 76: Stricter mocks = can no longer get away with skipping fields.

packages/ui-common/__tests__/components/ChatBot/ChatBot.test.tsx
```diff
@@ -59,12 +73,12 @@ describe("ChatBot", () => {
...
                 },
             },
         })
-        ;(useColorScheme as jest.Mock).mockReturnValue({
-            mode: "light",
-        })
+
+        vi.mocked(useColorScheme).mockReturnValue(getUseColorSchemeMockReturnValue("light"))
```
dsargent, line 81: Type-safe mock, no casting!

packages/ui-common/__tests__/components/ErrorPage/ErrorBoundary.test.tsx
```diff
@@ -18,12 +20,12 @@ const ErrorChild = ({shouldThrow}: {shouldThrow: boolean}) => {
 }
 
 describe("ErrorBoundary", () => {
-    let consoleErrorSpy: jest.SpyInstance<void, [message?: unknown, ...optionalParams: unknown[]]>
+    let consoleErrorSpy: MockInstance
```
dsargent, line 23: Simplify -- for Vitest we only need this.

packages/ui-common/__tests__/components/Settings/FadingCheckmark.test.tsx
dsargent, line 1: Simple migration to VItest.

packages/ui-common/__tests__/components/Settings/SettingsDialog.test.tsx
dsargent, line 1: Another simple one.

packages/ui-common/__tests__/components/Settings/SettingsRow.test.tsx
dsargent, line 1: And another simple migration

packages/ui-common/__tests__/unit/AgentChat/Utils.test.ts
dsargent, line 1: Simple migration.

packages/ui-common/__tests__/unit/Controller/agent/Agent.test.ts
```diff
@@ -282,10 +284,10 @@ describe("Controller/Agent/sendNetworkDesignerRequest", () => {
 
     it("calls sendChatQuery with the designer agent ID and correct sly_data", async () => {
         const signal = new AbortController().signal
-        const onChunk = jest.fn()
+        const onChunk = vi.fn()
         const updated: AgentNetworkDefinitionEntry[] = [{origin: "myAgent", tools: [], instructions: "Do stuff."}]
 
-        ;(sendLlmRequest as jest.Mock).mockImplementation((callback) => {
+        const sendRequestMock = vi.mocked(sendLlmRequest).mockImplementation(async (callback) => {
```
dsargent, line 290: Safe, typed mock.

packages/ui-common/__tests__/unit/Controller/agent/Agent.test.ts
```diff
@@ -300,7 +302,7 @@ describe("Controller/Agent/sendNetworkDesignerRequest", () => {
         )
 
         expect(sendLlmRequest).toHaveBeenCalledTimes(1)
-        const callArgs = (sendLlmRequest as jest.Mock).mock.calls[0]
+        const callArgs = sendRequestMock.mock.calls[0]
```
dsargent, line 305: No casting, cleaner, safer.

packages/ui-common/__tests__/unit/Controller/agent/Agent.test.ts
```diff
@@ -327,10 +329,10 @@ describe("Controller/Agent/sendNetworkDesignerRequest", () => {
...
             undefined,
             TEST_USERNAME,
-            jest.fn()
+            vi.fn()
         )
 
-        const requestBody = (sendLlmRequest as jest.Mock).mock.calls[0][3]
-        expect(requestBody.sly_data).not.toHaveProperty("agent_network_name")
+        const requestBody = sendLlmRequestMock.mock.calls[0][3]
+        expect(requestBody["sly_data"]).not.toHaveProperty("agent_network_name")
```
dsargent: Interesting tid-bit: now that we are using safe, typed, mocks, tsc was able to detect that `sly_data` requires an index signature here, so I had to fix it.

packages/ui-common/__tests__/unit/Theme/Theme.test.ts
```diff
@@ -1,6 +1,11 @@
+// eslint-disable-next-line no-shadow
+import {describe, expect, it} from "vitest"
+
+import {withStrictMocks} from "../../../../../__tests__/common/vitest/strictMocks"
 import {adjustBrightness, isLightColor} from "../../../Theme/Theme"
 
 describe("adjustBrightness", () => {
+    withStrictMocks()
```
dsargent, line 8: Wasn't using strict mocks!

jest.config.ts
```diff
@@ -100,10 +100,10 @@ const config: Config.InitialOptions = {
     // Instead, use the combined coverage nunmbers from RunCombinedTests.sh.
     coverageThreshold: {
         global: {
-            statements: -805,
-            branches: -506,
-            functions: -87,
-            lines: -241,
+            statements: -168,
```
dsargent, line 103: Meaningless numbers during migration, ignore.

package.json
```diff
@@ -37,6 +37,7 @@
         "@jest/types": "catalog:default",
         "@next/eslint-plugin-next": "16.2.9",
         "@testing-library/jest-dom": "6.9.1",
+        "@testing-library/react": "catalog:default",
```
dsargent, line 40: Can't remember why I had to add this 🤔

vitest.setup.ts
```diff
@@ -126,6 +129,11 @@ if (global.SVGPathElement === undefined) {
     global.SVGPathElement = FakeSVGPathElement as unknown as typeof SVGPathElement
 }
 
+// We've disabled globals for Vitest during migration so have to do this here
```
dsargent, line 132: I had to add this for RTL tests. Will go away once we're done with migration as Vitest will access the global for us.

RTL silently does this for us in Jest because the globals are all present and imported.

vitest_migration.js
dsargent, line 1: Renamed from `vitest_migration.ts` so we can reuse in the ESLint config (which is necessarily CJS.)

vitest_migration.ts
dsargent, line 1: Not deleted, renamed. See above.

build_scripts/RunCombinedTests.sh
swensel: Yeah this is a little confusing and it's going to make the coverage improvement less clear in https://github.com/cognizant-ai-lab/neuro-san-ui/pull/394.

swensel: I'm also not sure I trust Copilot.

swensel (approved):
LGTM.

PR #453, opened 2026-07-02
#444 Vitest migration: Phase 4

dsargent:
Next phase of migration. Covers all tests under `components/MultiAgentAccelerator` and subdirectories.

This one was a bit educational so I'll annotate accordingly. Some lessons learned for future tests.

### DIFF
--- a/build_scripts/RunCombinedTests.sh
+++ b/build_scripts/RunCombinedTests.sh
@@ -19,9 +19,9 @@ set -o pipefail
 # branches: -152,
 # functions: -27,
 # lines: -85,
-STATEMENTS=587
-BRANCHES=600
-FUNCTIONS=227
+STATEMENTS=1476
+BRANCHES=1299
+FUNCTIONS=497
 MAX_LINES=86
 
 rm -rf coverage-jest coverage-vitest coverage-combined-input coverage-combined
--- a/packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentFlow.test.tsx
+++ b/packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentFlow.test.tsx
@@ -19,8 +19,10 @@ import {act, fireEvent, render, screen, waitFor} from "@testing-library/react"
 import {userEvent, UserEvent} from "@testing-library/user-event"
 import {NodePositionChange, NodeRemoveChange, ReactFlowProvider} from "@xyflow/react"
 import {FC, useEffect} from "react"
+// eslint-disable-next-line no-shadow
+import {beforeEach, describe, expect, it, MockInstance, vi} from "vitest"
 
-import {withStrictMocks} from "../../../../../../__tests__/common/strictMocks"
+import {withStrictMocks} from "../../../../../../__tests__/common/vitest/strictMocks"
 import {cleanUpAgentName} from "../../../../components/AgentChat/Common/Utils"
 import {AgentConversation} from "../../../../components/MultiAgentAccelerator/AgentConversations"
 import {
@@ -32,60 +34,22 @@ import {
 import {AgentNetworkDefinitionEntry} from "../../../../components/MultiAgentAccelerator/const"
 import {ThoughtBubbleEdgeShape} from "../../../../components/MultiAgentAccelerator/ThoughtBubbles/ThoughtBubbleEdge"
 import {sendChatQuery} from "../../../../controller/agent/Agent"
-import {ChatMessageType, ConnectivityInfo} from "../../../../generated/neuro-san/NeuroSanClient"
+import {ChatMessageType, ChatResponse, ConnectivityInfo} from "../../../../generated/neuro-san/NeuroSanClient"
 import {useTempNetworksStore} from "../../../../state/TemporaryNetworks"
 import {PALETTES} from "../../../../Theme/Palettes"
 
-jest.mock("../../../../controller/agent/Agent")
-
-jest.mock("notistack", () => ({
-    ...jest.requireActual("notistack"),
-    enqueueSnackbar: jest.fn(),
-}))
-
-// Handle to the auto-mocked sendChatQuery so tests can stub it without repeating the cast.
-const mockSendChatQuery = sendChatQuery as jest.Mock
-
+//#region Constants
 const TEST_AGENT_MUSIC_NERD_PRO = "Music Nerd Pro"
 
 const mockPlasmaEdgeTestId = "mock-plasma-edge"
 const mockThoughtBubbleEdgeTestId = "mock-thought-bubble-edge"
 const mockThoughtBubbleOverlayTestId = "mock-thought-bubble-overlay"
 
-jest.mock("@mui/material/styles", () => ({
-    ...jest.requireActual("@mui/material/styles"),
-    useColorScheme: jest.fn(),
-}))
-
-jest.mock("../../../../components/MultiAgentAccelerator/AgentFlow/PlasmaEdge", () => ({
-    PlasmaEdge: () => <g data-testid={mockPlasmaEdgeTestId} />,
-}))
-
-jest.mock("../../../../components/MultiAgentAccelerator/ThoughtBubbles/ThoughtBubbleEdge", () => ({
-    ThoughtBubbleEdge: () => <g data-testid={mockThoughtBubbleEdgeTestId} />,
-}))
-
-// Provide a mutable implementation for the ThoughtBubbleOverlay mock so individual
-// tests can swap the implementation without attempting to redefine the module
-// export (which can throw "Cannot redefine property" errors).
-type ThoughtBubbleOverlayProps = {
-    onBubbleHoverChange?: (id: string) => void
-}
-
-const defaultMockThoughtBubbleOverlay: FC<ThoughtBubbleOverlayProps> = () => (
-    <div data-testid={mockThoughtBubbleOverlayTestId} />
-)
-
-let __MockThoughtBubbleOverlayImpl: FC<ThoughtBubbleOverlayProps> = defaultMockThoughtBubbleOverlay
-jest.mock("../../../../components/MultiAgentAccelerator/ThoughtBubbles/ThoughtBubbleOverlay", () => ({
-    ThoughtBubbleOverlay: (props: ThoughtBubbleOverlayProps) => __MockThoughtBubbleOverlayImpl(props),
-}))
-
-const NETWORK: ConnectivityInfo[] = [
+const NETWORK = [
     {
         origin: "agent1",
-        tools: ["agent2", "agent3"],
         display_as: "llm_agent",
+        tools: ["agent2", "agent3"],
     },
     {
         origin: "agent2",
@@ -95,7 +59,62 @@ const NETWORK: ConnectivityInfo[] = [
         origin: "agent3",
         tools: [],
     },
-]
+] satisfies ConnectivityInfo[]
+
+//#endregion Constants
+
+//#region Types
+
+// Provide a mutable implementation for the ThoughtBubbleOverlay mock so individual
+// tests can swap the implementation without attempting to redefine the module
+// export (which can throw "Cannot redefine property" errors).
+type ThoughtBubbleOverlayProps = {
+    onBubbleHoverChange?: (id: string) => void
+}
+//#endregion Types
+
+//#region Mocks
+
+vi.mock("../../../../controller/agent/Agent")
+
+vi.mock("notistack", async (importOriginal) => {
+    const actual = await importOriginal<typeof import("notistack")>()
+
+    return {
+        ...actual,
+        enqueueSnackbar: vi.fn(),
+    }
+})
+
+vi.mock("@mui/material/styles", async (importOriginal) => {
+    const actual = await importOriginal<typeof import("@mui/material/styles")>()
+
+    return {
+        ...actual,
+        useColorScheme: vi.fn(),
+    }
+})
+
+// Handle to the auto-mocked sendChatQuery so tests can stub it without repeating the cast.
+const mockSendChatQuery = vi.mocked(sendChatQuery)
+
+vi.mock("../../../../components/MultiAgentAccelerator/AgentFlow/PlasmaEdge", () => ({
+    PlasmaEdge: () => <g data-testid={mockPlasmaEdgeTestId} />,
+}))
+
+vi.mock("../../../../components/MultiAgentAccelerator/ThoughtBubbles/ThoughtBubbleEdge", () => ({
+    ThoughtBubbleEdge: () => <g data-testid={mockThoughtBubbleEdgeTestId} />,
+}))
+
+const defaultMockThoughtBubbleOverlay: FC<ThoughtBubbleOverlayProps> = () => (
+    <div data-testid={mockThoughtBubbleOverlayTestId} />
+)
+let __MockThoughtBubbleOverlayImpl: FC<ThoughtBubbleOverlayProps> = defaultMockThoughtBubbleOverlay
+vi.mock("../../../../components/MultiAgentAccelerator/ThoughtBubbles/ThoughtBubbleOverlay", () => ({
+    ThoughtBubbleOverlay: (props: ThoughtBubbleOverlayProps) => __MockThoughtBubbleOverlayImpl(props),
+}))
+
+//#endregion Mocks
 
 describe("AgentFlow", () => {
     let user: UserEvent
@@ -104,7 +123,14 @@ describe("AgentFlow", () => {
 
     beforeEach(() => {
         user = userEvent.setup()
-        ;(useColorScheme as jest.Mock).mockReturnValue({
+        vi.mocked(useColorScheme).mockReturnValue({
+            colorScheme: undefined,
+            darkColorScheme: undefined,
+            lightColorScheme: undefined,
+            setColorScheme: vi.fn(),
+            setMode: vi.fn(),
+            systemMode: undefined,
+            allColorSchemes: ["light", "dark"],
             mode: "light",
         })
         useTempNetworksStore.getState().setTempNetworks([])
@@ -141,7 +167,7 @@ describe("AgentFlow", () => {
         isAwaitingLlm: false,
         isStreaming: false,
         thoughtBubbleEdges: new Map(),
-        setThoughtBubbleEdges: jest.fn(),
+        setThoughtBubbleEdges: vi.fn(),
     }
 
     const renderAgentFlowComponent = (overrides = {}, mode: PaletteMode = "light") => {
@@ -179,7 +205,7 @@ describe("AgentFlow", () => {
     // Simulates React's functional-setState pattern so tests can inspect the resulting Map.
     const createThoughtBubbleEdgesStore = () => {
         let map = new Map<string, {edge: ThoughtBubbleEdgeShape; timestamp: number}>()
-        const mockSetThoughtBubbleEdges = jest.fn((updater: unknown) => {
+        const mockSetThoughtBubbleEdges = vi.fn((updater: unknown) => {
             if (typeof updater === "function") {
                 map = (updater as (prev: typeof map) => typeof map)(map)
             }
@@ -204,7 +230,7 @@ describe("AgentFlow", () => {
         })
 
         it("Should show the Edit button on a temporary network and invoke onEnterEditMode when clicked", async () => {
-            const onEnterEditMode = jest.fn()
+            const onEnterEditMode = vi.fn()
             renderAgentFlowComponent({
                 networkDisplayName: "Temp Net",
                 isSelectedNetworkTemporary: true,
@@ -277,7 +303,7 @@ describe("AgentFlow", () => {
         it("Should allow switching to heatmap display and not show radial guides with linear display mode", async () => {
             const {container} = renderAgentFlowComponent()
 
-            let radialGuides = container.querySelector("#test-flow-id-radial-guides")
+            const radialGuides = container.querySelector("#test-flow-id-radial-guides")
 
             // Radial guides should be present in radial layout
             expect(radialGuides).toBeInTheDocument()
@@ -299,8 +325,8 @@ describe("AgentFlow", () => {
             await user.click(heatmapButton)
 
             // Radial guides should still not be present in linear layout
-            radialGuides = container.querySelector("#test-flow-id-radial-guides")
-            expect(radialGuides).not.toBeInTheDocument()
+            const radialGuidesAfterClick = container.querySelector("#test-flow-id-radial-guides")
+            expect(radialGuidesAfterClick).not.toBeInTheDocument()
         })
 
         it("Should handle isStreaming false", () => {
@@ -406,7 +432,7 @@ describe("AgentFlow", () => {
                         isAwaitingLlm={false}
                         isStreaming={false}
                         thoughtBubbleEdges={new Map()}
-                        setThoughtBubbleEdges={jest.fn()}
+                        setThoughtBubbleEdges={vi.fn()}
                     />
                 </ReactFlowProvider>
             )
@@ -494,7 +520,7 @@ describe("AgentFlow", () => {
                             },
                         ]}
                         thoughtBubbleEdges={new Map()}
-                        setThoughtBubbleEdges={jest.fn()}
+                        setThoughtBubbleEdges={vi.fn()}
                     />
                 </ReactFlowProvider>
             )
@@ -615,7 +641,7 @@ describe("AgentFlow", () => {
         })
 
         it("Should call setThoughtBubbleEdges when conversations with text are added", () => {
-            const mockSetThoughtBubbleEdges = jest.fn()
+            const mockSetThoughtBubbleEdges = vi.fn()
             const conversationsWithText = [
                 {
                     id: "conv-with-text",
@@ -689,7 +715,7 @@ describe("AgentFlow", () => {
         })
 
         it("Should prevent duplicate thought bubbles using thoughtBubbleEdges", () => {
-            const mockSetThoughtBubbleEdges = jest.fn()
+            const mockSetThoughtBubbleEdges = vi.fn()
             const existingEdgesMap = new Map<string, {edge: ThoughtBubbleEdgeShape; timestamp: number}>([
                 [
                     "conv-1",
@@ -774,7 +800,7 @@ describe("AgentFlow", () => {
         })
 
         it("Should clean up thought bubbles via removeThoughtBubbleEdgeHelper during timeout", () => {
-            jest.useFakeTimers()
+            vi.useFakeTimers()
             const {mockSetThoughtBubbleEdges, getThoughtBubbleEdgesMap} = createThoughtBubbleEdgesStore()
 
             const conversationsWithText: AgentConversation[] = [
@@ -805,7 +831,7 @@ describe("AgentFlow", () => {
 
             // Fast-forward time by 11 seconds (past THOUGHT_BUBBLE_TIMEOUT_MS of 10 seconds)
             act(() => {
-                jest.advanceTimersByTime(11000)
+                vi.advanceTimersByTime(11000)
             })
 
             // The bubble should have been removed from the map after expiry
@@ -832,8 +858,8 @@ describe("AgentFlow", () => {
         })
 
         it("Should prevent expired bubbles from being removed when hovered", async () => {
-            jest.useFakeTimers()
-            const mockSetThoughtBubbleEdges = jest.fn()
+            vi.useFakeTimers()
+            const mockSetThoughtBubbleEdges = vi.fn()
 
             // Create a conversation that will be added as a thought bubble
             const conversationsWithText: AgentConversation[] = [
@@ -874,7 +900,7 @@ describe("AgentFlow", () => {
 
             // Fast-forward time by 11 seconds to trigger cleanup (past the 10-second timeout)
             act(() => {
-                jest.advanceTimersByTime(11000)
+                vi.advanceTimersByTime(11000)
             })
 
             // The bubble should not be removed because it's being hovered
@@ -885,8 +911,8 @@ describe("AgentFlow", () => {
         })
 
         it("Should drop expired bubbles first when overflow limit is reached", () => {
-            jest.useFakeTimers()
-            const mockSetThoughtBubbleEdges = jest.fn()
+            vi.useFakeTimers()
+            const mockSetThoughtBubbleEdges = vi.fn()
 
             // Create 5 conversations to fill MAX_THOUGHT_BUBBLES (5) with bubbles whose startedAt is the current
             // fake time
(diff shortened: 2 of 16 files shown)


build_scripts/RunCombinedTests.sh
```diff
@@ -19,9 +19,9 @@ set -o pipefail
 # branches: -152,
 # functions: -27,
 # lines: -85,
-STATEMENTS=587
-BRANCHES=600
-FUNCTIONS=227
+STATEMENTS=1476
+BRANCHES=1299
+FUNCTIONS=497
```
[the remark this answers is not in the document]
dsargent: Expected during migration.

dsargent, line 24: "Lines" is still holding steady.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentFlow.test.tsx
```diff
@@ -32,70 +34,87 @@ import {
...
 
 const mockPlasmaEdgeTestId = "mock-plasma-edge"
 const mockThoughtBubbleEdgeTestId = "mock-thought-bubble-edge"
 const mockThoughtBubbleOverlayTestId = "mock-thought-bubble-overlay"
 
-jest.mock("@mui/material/styles", () => ({
-    ...jest.requireActual("@mui/material/styles"),
-    useColorScheme: jest.fn(),
-}))
+const NETWORK = [
```
dsargent, line 48: Consolidated items more neatly into `#region`s.

Unfortunately causes a lot of diff noise, but better for maintenance and readability long term.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentFlow.test.tsx
```diff
@@ -32,70 +34,87 @@ import {
...
 type ThoughtBubbleOverlayProps = {
     onBubbleHoverChange?: (id: string) => void
 }
+//#endregion Types
+
+//#region Mocks
+
+vi.mock("../../../../controller/agent/Agent")
+
+vi.mock("notistack", async (importOriginal) => {
```
dsargent, line 80: This is how the vitest doc (and Copilot) says to do partial mocking. Compared to Jest's `requireActual`. One interesting thing is Vitest's is async, hence the `await importOriginal`, unlike Jest's.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentFlow.test.tsx
```diff
@@ -104,7 +123,14 @@ describe("AgentFlow", () => {
 
     beforeEach(() => {
         user = userEvent.setup()
-        ;(useColorScheme as jest.Mock).mockReturnValue({
+        vi.mocked(useColorScheme).mockReturnValue({
```
dsargent, line 126: The component under test _does_ do some dark theme shenanigans, so I left in the MUI theme mockery. But I made it strongly typed, so that means we can't get away with a wrong-shaped mock any more; all fields have to be mocked if they are required in the thing being mocked.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentFlow.test.tsx
```diff
@@ -299,8 +325,8 @@ describe("AgentFlow", () => {
             await user.click(heatmapButton)
 
             // Radial guides should still not be present in linear layout
-            radialGuides = container.querySelector("#test-flow-id-radial-guides")
-            expect(radialGuides).not.toBeInTheDocument()
+            const radialGuidesAfterClick = container.querySelector("#test-flow-id-radial-guides")
```
dsargent, line 328: Better not to reuse local variables. `let` is always a last resort, as it violates immutability, though in rare cases it is required.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentFlow.test.tsx
```diff
@@ -1744,29 +1770,32 @@ describe("AgentFlow", () => {
          */
         const mockInFlightDockApply = () => {
             let release!: () => void
+
             mockSendChatQuery.mockImplementation(
                 (_url, _signal, _query, _agent, chunkCallback) =>
-                    new Promise<void>((resolve) => {
+                    new Promise<ChatResponse>((resolve) => {
```
dsargent, line 1776: `void` was incorrect here; the function being mocked actually returns `Promise<ChatResponse>`, so now the mock has to as well.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentFlow.test.tsx
```diff
@@ -1744,29 +1770,32 @@ describe("AgentFlow", () => {
...
                         release = () => {
                             chunkCallback(makeDockReservationChunk(DOCK_DEFAULT_RES, DOCK_NETWORK_NAME))
-                            resolve()
+                            resolve({} satisfies ChatResponse)
```
dsargent, line 1779: Here we use `satisfies` for type safety. Does `{}` in fact "satisfy" `ChatResponse`? Yes, because in the definition of `ChatResponse` (inferred from the Swagger API definitions), all fields are optional, so `{}` is a shape-compatible substitute.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentFlow.test.tsx
```diff
@@ -1834,12 +1863,12 @@ describe("AgentFlow", () => {
         })
 
         it("aborts an in-flight dock request if the close button is clicked during streaming", async () => {
-            const consoleErrorSpy = jest.spyOn(console, "error").mockImplementation()
-            const onExitEditMode = jest.fn()
+            const consoleErrorSpy = vi.spyOn(console, "error").mockImplementation(vi.fn())
```
dsargent, line 1866: Small change throughout all these migrations: in Vitest you can't have `.mockImplementation()` with no parameter; you have to supply a function, unlike Jest, so, we feed it `vi.fn()`. Simple.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentFlow.test.tsx
```diff
@@ -1834,12 +1863,12 @@ describe("AgentFlow", () => {
...
+            const onExitEditMode = vi.fn()
             let capturedSignal: AbortSignal | undefined
             mockSendChatQuery.mockImplementation(
                 (_url: string, signal: AbortSignal) =>
-                    new Promise<void>((_resolve, reject) => {
+                    new Promise<ChatResponse>((_resolve, reject) => {
```
dsargent, line 1871: Again, shape of mocked return Promise was incorrect.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentFlow.test.tsx
```diff
@@ -1999,16 +2030,16 @@ describe("AgentFlow", () => {
...
             // Re-initialise userEvent with advanceTimers so its internal delays stay in sync with
             // fake timers (the module-level `user` is bound to real timers and would stall here).
-            const localUser = userEvent.setup({advanceTimers: jest.advanceTimersByTime.bind(jest)})
-            mockSendChatQuery.mockImplementation(
-                (_url: string, signal: AbortSignal) =>
-                    new Promise<void>((_resolve, reject) => {
-                        signal.addEventListener("abort", () => reject(new DOMException("Aborted", "AbortError")))
-                    })
-            )
+            const localUser = userEvent.setup({advanceTimers: (ms) => vi.advanceTimersByTime(ms)})
```
dsargent, line 2037: Copilot suggested this alternative syntax for the mocked `UserEvent`, which to me is more intuitive than the weird `.bind(jest)` thing.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentFlow.test.tsx
```diff
@@ -1999,16 +2030,16 @@ describe("AgentFlow", () => {
...
+            mockSendChatQuery.mockImplementation((_url: string, signal: AbortSignal) => {
+                return new Promise<ChatResponse>((_resolve, reject) => {
```
dsargent, line 2039: Again, returned Promise had the wrong shape. These issues were all surfaced by having strongly-typed mocks.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentFlow.test.tsx
```diff
@@ -2019,20 +2050,22 @@ describe("AgentFlow", () => {
             })
 
             const promptField = screen.getByPlaceholderText(PROMPT_PLACEHOLDER)
-            await localUser.type(promptField, "add a node")
-            await localUser.click(screen.getByRole("button", {name: APPLY_BUTTON}))
+            await vi.waitFor(async () => localUser.type(promptField, "add a node"))
```
dsargent, line 2053: Couple of points here:

1) Copilot pointed out that when fake timers are active, it's as if "time isn't passing" so you almost never want to use the `findBy` family of queries, since they include an implicit `await` on time passing.

2) Instead of `await` on the `localUser` events, we have to use `await vi.waitFor` because it is fake-timer-aware. The test hangs without this as as far as regular `await` is concerned, no time ever passes so it will wait forever for the thing it's awaiting.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentFlow.test.tsx
```diff
@@ -2019,20 +2050,22 @@ describe("AgentFlow", () => {
...
 
             // Advance past the 120-second dock apply timeout
             await act(async () => {
-                jest.advanceTimersByTime(121_000)
+                vi.advanceTimersByTime(121_000)
             })
 
             // The user sees an error banner explaining the request timed out...
-            expect(await screen.findByText(/timed out/iu)).toBeInTheDocument()
+            screen.getByText(/timed out/u)
```
dsargent, line 2064: Avoid using `findBy...` when fake timers are running.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentFlow.test.tsx
```diff
@@ -2064,9 +2097,10 @@ describe("AgentFlow", () => {
                 // Send low-expiry first, then high-expiry (high should win)
                 chunkCallback(makeChunk(FIRST_RES, LOW_EXPIRY))
                 chunkCallback(makeChunk(SECOND_RES, HIGH_EXPIRY))
+                return {} satisfies ChatResponse
```
dsargent, line 2100: With typed mocks, we can't get away with not returning anything because that's not how the function being mocked works. So Vitest's typed mocks force us to adhere more closely to the contract of the functions being mocked, which is better all round.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentFlow.test.tsx
```diff
@@ -2213,12 +2247,11 @@ describe("AgentFlow", () => {
         })
 
         it("Stop & discard aborts the request, hides backdrop, shows a cancel banner, restores prompt", async () => {
-            mockSendChatQuery.mockImplementation(
-                (_url: string, signal: AbortSignal) =>
-                    new Promise<void>((_resolve, reject) => {
-                        signal.addEventListener("abort", () => reject(new DOMException("Aborted", "AbortError")))
-                    })
-            )
+            mockSendChatQuery.mockImplementation((_url: string, signal: AbortSignal) => {
```
dsargent, line 2250: Same again, wrong shape of returned Promise.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentFlow.test.tsx
```diff
@@ -2280,16 +2314,21 @@ describe("AgentFlow", () => {
                 currentUser: "test-user",
             })
 
-            await localUser.type(screen.getByPlaceholderText(PROMPT_PLACEHOLDER), "add a node")
-            await localUser.click(screen.getByRole("button", {name: APPLY_BUTTON}))
+            const editInput = screen.getByPlaceholderText(PROMPT_PLACEHOLDER)
```
dsargent, line 2317: Extract explanatory variable names + use `vi.waitFor` for fake timers rather than the perilous `findBy` queries.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentFlow.test.tsx
```diff
@@ -2280,16 +2314,21 @@ describe("AgentFlow", () => {
...
+            const applyButton = screen.getByRole("button", {name: APPLY_BUTTON})
+            await vi.waitFor(() => localUser.click(applyButton))
+
+            screen.getByText(APPLIED_BANNER)
 
             // Once the auto-dismiss timer fires, the banner disappears
-            act(() => {
-                jest.advanceTimersByTime(DOCK_BANNER_AUTO_DISMISS_MS + 100)
+            await act(async () => {
+                await vi.advanceTimersByTimeAsync(DOCK_BANNER_AUTO_DISMISS_MS + 100)
```
dsargent, line 2328: Copilot recommended `advanceTimersByTimeAsync` for some of these situations but I'm still not sure where to use that vs. the non-async version. Tests pass so 🤷🏻

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentNode.test.tsx
```diff
@@ -18,24 +18,30 @@ import {hexToRgb} from "@mui/material/styles"
...
-    ...jest.requireActual("@xyflow/react"),
-    Handle: (props: {type: unknown; id: unknown; style: CSSProperties}) => (
-        <div
-            data-testid={`handle-${props.type}-${props.id}`}
-            style={props.style}
-        />
-    ),
-}))
+// Mock the Handle component since we don't want to invite @xyflow/react to this party
+vi.mock("@xyflow/react", async (importOriginal) => {
```
dsargent, line 32: Partial mocking with Vitest

packages/ui-common/__tests__/components/MultiAgentAccelerator/Sidebar/Sidebar.test.tsx
```diff
@@ -579,21 +585,22 @@ describe("SideBar", () => {
 
         // Flush pending state updates; findByText succeeds on the first poll (element
         // already present), so no fake-timer conflict.
-        const treeItem = (await screen.findByText(cleanUpAgentName(TEMPORARY_NETWORK_NAME))).closest(
-            '[role="treeitem"]'
-        )
+        const treeItem = screen.getByText(cleanUpAgentName(TEMPORARY_NETWORK_NAME)).closest('[role="treeitem"]')
```
dsargent, line 588: Simplify

packages/ui-common/__tests__/components/MultiAgentAccelerator/ThoughtBubbles/ThoughtBubbleEdge.test.tsx
```diff
@@ -94,7 +96,7 @@ describe("ThoughtBubbleEdge", () => {
         expect(path).toHaveAttribute("d")
         expect(path).toHaveAttribute("fill", "none")
 
-        // The d attribute should contain a bezier curve path
+        // The d attribute should contain a Bézier curve path
```
dsargent, line 99: Let's give the [man](https://en.wikipedia.org/wiki/Pierre_B%C3%A9zier?useskin=vector) his dues.

packages/ui-common/__tests__/components/MultiAgentAccelerator/ThoughtBubbles/ThoughtBubbleOverlay.test.tsx
dsargent, line 1: Aside: this module is _super_ copy-pastish. My IDE flagged 5 separate blocks, each of 10-20 lines, copy-pasted between tests. That is a maintenance headache, as I found out during this migration since I had to do the same work over and over.

Removing all this vibed duplication should be a priority for the thought bubbles rewrite.

packages/ui-common/__tests__/components/MultiAgentAccelerator/Sidebar/Sidebar.test.tsx
```diff
@@ -613,11 +620,11 @@ describe("SideBar", () => {
             newlyAddedTemporaryNetworks: new Set(["any-name"]),
         })
 
-        await screen.findByText("Agent Networks")
+        screen.getByText("Agent Networks")
```
dsargent, line 623: Let's try to default to `getByText` _unless_ it actually is a `findBy` scenario, meaning, "I don't expect this thing to already be in the DOM but it should be shortly". Classic example: tooltip that pops up after _n_ milliseconds.

packages/ui-common/__tests__/components/MultiAgentAccelerator/ImportNetworkModal.test.tsx
```diff
@@ -258,9 +260,10 @@ describe("ImportNetworkModal", () => {
     })
 
     it("should show an error when the file cannot be read", async () => {
-        const readSpy = jest.spyOn(FileReader.prototype, "readAsText").mockImplementation(() => {
-            const reader = readSpy.mock.contexts.at(-1)
-            reader?.dispatchEvent(new Event("error"))
+        vi.spyOn(FileReader.prototype, "readAsText").mockImplementation(function (this: FileReader) {
```
dsargent, line 263: The typed mock was (rightly) objecting here, so I had to change this around to mimic the real `readAsText`.

packages/ui-common/__tests__/components/MultiAgentAccelerator/MultiAgentAccelerator.test.tsx
dsargent, line 1: This was the big painful one in this migration batch.

packages/ui-common/__tests__/components/MultiAgentAccelerator/MultiAgentAccelerator.test.tsx
```diff
@@ -122,46 +124,45 @@ jest.mock("../../../components/MultiAgentAccelerator/AgentFlow/AgentFlow", () =>
     },
 }))
 
-jest.mock("../../../components/MultiAgentAccelerator/Sidebar/Sidebar", () => {
-    const originalModule = jest.requireActual("../../../components/MultiAgentAccelerator/Sidebar/Sidebar")
+vi.mock("../../../components/MultiAgentAccelerator/Sidebar/Sidebar", async (importOriginal) => {
```
dsargent, line 127: Partial mock

packages/ui-common/__tests__/components/MultiAgentAccelerator/MultiAgentAccelerator.test.tsx
```diff
@@ -122,46 +124,45 @@ jest.mock("../../../components/MultiAgentAccelerator/AgentFlow/AgentFlow", () =>
...
+    const originalModule =
+        await importOriginal<typeof import("../../../components/MultiAgentAccelerator/Sidebar/Sidebar")>()
+
     return {
         __esModule: true,
```
dsargent, line 132: Actually we can probably get rid of the `__esModule` stuff in most cases, when mocking with Vitest. Will add to my "post-migration clean-up" list.

packages/ui-common/__tests__/components/MultiAgentAccelerator/MultiAgentAccelerator.test.tsx
```diff
@@ -122,46 +124,45 @@ jest.mock("../../../components/MultiAgentAccelerator/AgentFlow/AgentFlow", () =>
...
             const OriginalModal = originalModule.ImportNetworkModal
             return <OriginalModal {...props} />
         },
     }
 })
 
-jest.mock("../../../components/MultiAgentAccelerator/AgentConversations")
-
-// Mock MUI theming
-jest.mock("@mui/material/styles", () => ({
```
dsargent, line 156: Component under test only uses the `theme` for z-index values, which we don't even test in this module, so I removed this and related `each` matrix testing logic.

packages/ui-common/__tests__/components/MultiAgentAccelerator/MultiAgentAccelerator.test.tsx
```diff
@@ -252,18 +253,13 @@ const renderMultiAgentAcceleratorPage = () =>
...
-        ;(getConnectivity as jest.Mock).mockResolvedValue(MOCK_CONNECTIVITY_INFO)
-        ;(testConnection as jest.Mock).mockResolvedValue({success: true, status: "ok", version: "1.0.0"})
-        ;(sendNetworkDesignerRequest as jest.Mock).mockResolvedValue([])
-
-        // make extractConversations the real implementation
-        ;(extractConversations as jest.Mock).mockImplementation(
-            jest.requireActual("../../../components/MultiAgentAccelerator/AgentConversations").extractConversations
-        )
+    beforeEach(async () => {
+        vi.mocked(getAgentNetworks).mockResolvedValue(LIST_NETWORKS_RESPONSE)
```
dsargent, line 259: These become cleaner with typed mocks (and safer).

packages/ui-common/__tests__/components/MultiAgentAccelerator/MultiAgentAccelerator.test.tsx
```diff
@@ -274,45 +270,38 @@ describe("MultiAgentAccelerator", () => {
         useTourStore.getState().reset()
     })
 
-    it.each(["dark", "light"])(
-        "should render the component with mode:%s and change the network when item is clicked in the sidebar",
-        async (mode) => {
```
dsargent, line 279: No need to do this -- component under test delegates to MUI theming.

packages/ui-common/__tests__/components/MultiAgentAccelerator/MultiAgentAccelerator.test.tsx
```diff
@@ -375,15 +363,13 @@ describe("MultiAgentAccelerator", () => {
...
-
         let resolveSend: () => void
         const sendPromise = new Promise<void>((resolve) => {
             resolveSend = resolve
         })
 
-        ;(sendNetworkDesignerRequest as jest.Mock).mockImplementation(
-            async (_neuroSanUrl, _signal, _frontman, _networkDef, _agentNetworkName, _userName, onChunk) => {
+        vi.mocked(sendNetworkDesignerRequest).mockImplementation(
+            async (_url, _signal, _agentName, _updated, _agentNetworkName, _currentUser, onChunk) => {
```
dsargent, line 372: Modified to use same param names as function being mocked, for consistency.

packages/ui-common/__tests__/components/MultiAgentAccelerator/MultiAgentAccelerator.test.tsx
```diff
@@ -122,46 +124,45 @@ jest.mock("../../../components/MultiAgentAccelerator/AgentFlow/AgentFlow", () =>
...
         ImportNetworkModal: (props: ImportNetworkModalProps) => {
             onImport = props.onImport
+
             const OriginalModal = originalModule.ImportNetworkModal
             return <OriginalModal {...props} />
         },
     }
 })
 
-jest.mock("../../../components/MultiAgentAccelerator/AgentConversations")
```
dsargent, line 153: No longer mock this globally because we only need the mock behavior in one single test (see later).

packages/ui-common/__tests__/components/MultiAgentAccelerator/MultiAgentAccelerator.test.tsx
```diff
@@ -1123,6 +1106,9 @@ describe("MultiAgentAccelerator", () => {
         })
 
         it("Should handle temporary networks expiring", async () => {
+            const now = Date.now()
```
dsargent, line 1109: This test was a real pain to migrate due to fake timers, which Vitest is much stricter about than Jest.

packages/ui-common/__tests__/components/MultiAgentAccelerator/MultiAgentAccelerator.test.tsx
```diff
@@ -1172,12 +1152,11 @@ describe("MultiAgentAccelerator", () => {
...
 
             const displayAgentName = cleanUpAgentName(TEMPORARY_NETWORK.reservation.reservation_id)
+            screen.getByText(displayAgentName)
 
-            // Make sure we see the temp network
-            const tempNetworkItem = await screen.findByText(displayAgentName)
-
-            // Click the network to select it
-            await localUser.click(tempNetworkItem)
+            await act(async () => {
```
dsargent, line 1157: Now I know more I could probably make this work with `localUser.click()` but this is the fix that Copilot suggested -- don't use click but "drive it directly" by calling the selection change function. May revisit later.

packages/ui-common/__tests__/components/MultiAgentAccelerator/MultiAgentAccelerator.test.tsx
```diff
@@ -1192,7 +1171,7 @@ describe("MultiAgentAccelerator", () => {
 
             // First time "expired check" runs, it should not be expired yet
             await act(async () => {
-                jest.runOnlyPendingTimers()
+                await vi.advanceTimersByTimeAsync(EXPIRED_NETWORKS_CHECK_INTERVAL_MS)
```
dsargent, line 1174: `runOnlyPendingTimers` was a bit of a scattergun approach. With this, we advance time by only the amount we expect for the thing we want to happen, namely, that the reaper runs a cycle.

packages/ui-common/__tests__/components/MultiAgentAccelerator/MultiAgentAccelerator.test.tsx
```diff
@@ -1203,10 +1182,12 @@ describe("MultiAgentAccelerator", () => {
...
 
-            // advanced past expiration time but still within grace period, meaning we show the network but flag it as
-            // expired and do not let the user select it
+            // Jump system time past server-side expiration, then run exactly one interval tick.
+            vi.setSystemTime(now + expirationTimeSeconds * 1000 + 1)
+
+            // Make sure reaper runs
             await act(async () => {
-                jest.advanceTimersByTime(expirationTimeSeconds * 1000)
+                await vi.advanceTimersByTimeAsync(EXPIRED_NETWORKS_CHECK_INTERVAL_MS)
```
dsargent, line 1190: Advance to the next reaper cycle. Hmm, this never failed for me, but I wonder if there's a [fencepost issue](https://en.wikipedia.org/wiki/Off-by-one_error?useskin=vector#Fencepost_error) here where we're unlucky enough to hit it right on the tick, before it has run? I don't think so; probably okay for now.

packages/ui-common/__tests__/components/MultiAgentAccelerator/MultiAgentAccelerator.test.tsx
```diff
@@ -1249,18 +1219,46 @@ describe("MultiAgentAccelerator", () => {
...
-    it("Should pass along network icon suggestions to the sidebar", async () => {
-        const iconSuggestions = [
-            {
-                tool: "copy_cat",
-                icon_url: "Copy",
-            },
-            {
-                tool: "date_time_provider",
-                icon_url: "DateTime",
+    it("Should not allow user to select an expired temporary network", async () => {
```
dsargent, line 1222: Broke this out into a separate test as part of the going back and forth with Copilot on the "expired networks" test. There's too much going on in that other test anyway.

packages/ui-common/__tests__/components/MultiAgentAccelerator/MultiAgentAccelerator.test.tsx
```diff
@@ -1542,34 +1538,44 @@ describe("MultiAgentAccelerator", () => {
         ])(
             "Should handle responding '$buttonName' to the tour prompt correctly",
             async ({buttonName, shouldStartTour, expectedStatus}) => {
-                jest.useFakeTimers()
-                const localUser = userEvent.setup({advanceTimers: jest.advanceTimersByTime.bind(jest)})
+                vi.useFakeTimers()
```
dsargent, line 1541: Same issues in this test with fake timers and `Vitest`, same solutions: use `vi.waitFor()`, avoid `findBy()` while fake timers are running.

packages/ui-common/components/MultiAgentAccelerator/const.ts
```diff
@@ -73,3 +73,6 @@ export const GRACE_PERIOD_MS = 5 * 60 * 1000 // 5 minutes
 
 // We show the tour modal after this amount of time so as not to "pounce" on the user when they first open the app
 export const SHOW_TOUR_DELAY_MS = 5000
+
+// Check for expired networks every this many milliseconds
+export const EXPIRED_NETWORKS_CHECK_INTERVAL_MS = 10 * 1000
```
dsargent, line 78: Moved to here for use in tests (and to be consistent with other consts.)

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentFlow.test.tsx
swensel: > One interesting thing is Vitest's is async, hence the await importOriginal, unlike Jest's.

Yeah I also noticed that in one of the previous PRs.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentFlow.test.tsx
swensel: Good change.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentFlow.test.tsx
swensel: Agreed.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentFlow.test.tsx
swensel: Good to know.

swensel (approved):
LGTM.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentFlow.test.tsx
dsargent: I've done some more debugging on this and now I think Copilot was full of it, maybe hallucinating.

It's actually due to an RTL bug: https://github.com/testing-library/user-event/issues/1115

Will adapt accordingly in the latest PR. What we have here is "correct" and "works", but I don't think it's the valid long-term solution.

Issue #183, opened 2026-03-09
Limit yarn lockfile to only have production deps


dsargent:
Dependabot looks in there for CVEs and we aren't really that interested in CVEs that only affect local dev libraries.

PR #460, opened 2026-07-02
#444 Vitest post-migration clean-up

dsargent:
Follow-up to migration: clean up all Jest-related items:

1) Dependencies
2) Configs
3) Babel
4) Mocks

It's a large number of modified files, but mostly shallow, simple import changes and only affects test files, so, low risk.

I had to decide: do we go with "globals", like we've always done with Jest, or move to implicit includes (like we were doing just prior to this PR during migration?)

With globals, you can just use `it`, `describe`, `expect` etc. without importing them. It's convenient, but it's "invisible magic" and less explicit. Without globals, you have to import all those test functions in every test suite, whenever you want to use them, which is annoying. I went the "globals" route to match how we've been doing things with Jest. But open to discussion; I don't feel strongly one way or the other. One thing with Vitest globals, a small handful of tests misbehaved and I had to fix them. But not a huge deal in the scale of things.

Next PR: optimization.
After that: linting (for Vitest).


__tests__/common/strictMocks.ts
dsargent, line 1: Moved this to be the "default implementation" and removed imports since we're using Vitest globals.

__tests__/common/TestUtils.ts
```diff
@@ -14,8 +14,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import {vi} from "vitest"
```
dsargent, line 17: The huge majority of changes in this PR are just removing the Vitest imports, since the config is changed to use globals. Not a no-brainer, worth pondering how we want this to be for the future.

apps/main/__tests__/__mocks__/next-auth/react.js
dsargent, line 1: Was only used by Jest, so, deleted.

apps/main/__tests__/__mocks__/styleMock.js
dsargent, line 1: Was only used by Jest, so, deleted.

apps/main/__tests__/pages/api/handlers/Handlers.test.ts
dsargent, line 1: Remove Vitest imports (due to globals) and adjust `strictMocks` import path.

I won't bother commenting all of the many occurrences of this pattern.

apps/main/__tests__/pages/IndexPage.test.tsx
```diff
@@ -47,12 +45,10 @@ const createMockRouter = (overrides: Partial<NextRouter> = {}): NextRouter => ({
 
 // Mock dependencies
 vi.mock("../../../../packages/ui-common/state/Environment", () => ({
-    __esModule: true,
```
dsargent, line 50: `__esModule` not usually required for Vitest, as it's ESM native, unlike Jest.

build_scripts/vitest/UpdateCoverage.sh
dsargent, line 1: Not deleted. Moved to "default" position directly under `build_scripts`. The Jest one was deleted.

build_scripts/RunCombinedTests.sh
dsargent, line 1: This one _is_ deleted. No longer required now that migration is done.

build_scripts/UpdateCoverage.sh
dsargent, line 1: GitHub sees it as modified, but I actually deleted the Jest version and moved the Vitest version to where it was, making the Vitest version the default and only one.

But GitHub shows it as diffs which is handy for this PR.

packages/dev-common/Configs/eslint.config.mjs
dsargent, line 1: Remove Jest-isms, modify for Vitest.

Note: linting is still incomplete for Vitest. In a coming PR I will add the plugins and configs for Vitest to bring it as close to parity with what we had for Jest as I can.

packages/dev-common/Configs/eslint.config.mjs
```diff
@@ -164,7 +150,7 @@ export default defineConfig([
             // with other plugins we're using?
             // Also, the place for rules we want to explicitly enable.
             //
-            // How to see which rules are enabled: ./node_modules/.bin/eslint --print-config jest.config.ts
+            // How to see which rules are enabled: yarn lint-rules
```
dsargent, line 153: Simplify

packages/dev-common/Configs/eslint.config.mjs
```diff
@@ -285,13 +271,7 @@ export default defineConfig([
...
                     includeInternal: true,
                     includeTypes: true,
-                    devDependencies: [
-                        "**/__tests__/**",
-                        "jest*.*",
-                        "vitest*.*",
-                        "**/eslint.config.mjs",
-                        "**/knip.config.ts",
-                    ],
+                    devDependencies: ["**/__tests__/**", "vitest*.*", "**/eslint.config.mjs", "**/knip.config.ts"],
```
dsargent, line 274: Remove Jest items

packages/dev-common/Configs/eslint.config.mjs
```diff
@@ -557,12 +529,6 @@ export default defineConfig([
             "react/no-multi-comp": "off",
         },
     },
-    {
```
dsargent, line 560: Not needed now that migration is done.

packages/dev-common/package.json
```diff
@@ -11,8 +11,6 @@
         "eslint": "*",
         "eslint-config-prettier": "*",
         "eslint-plugin-import-x": "*",
-        "eslint-plugin-jest": "*",
```
dsargent, line 14: Will need to find equivalents for Vitest.

packages/ui-common/__tests__/components/Authentication/Auth.test.tsx
dsargent, line 1: Somehow this one was still referencing Jest. No idea how it slipped through, but I fixed it now.

I think it's because it was only using Jest for casts, which are invisible at runtime (erased), so the tests were passing just fine. Only when I actually removed Jest entirely did tsc start complaining.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentFlow.test.tsx
```diff
@@ -122,6 +121,13 @@ describe("AgentFlow", () => {
     withStrictMocks()
 
     beforeEach(() => {
+        // This has nothing to do with Jest itself and everything to do with a bug in React Testing Library.
```
dsargent, line 124: This was one of the issues I had with globals. See comment + link.
Actually I think it has more to do with the linked bug, and the broken interaction between Vitest fake timers and RTL UserEvent. So, not a "black mark" against using globals. We should make that decision independently.

Bottom line, my comments in previous PRs about avoiding `findBy` when fake timers are running and being sure to use `vi.waitFor` -- not applicable. Those were Copilot hallucinations, and the real issue was the bug linked in this comment. 🤦🏻 Copilot made me waste a _lot_ of time on that. Copilot is like a friend who occasionally helps you spot a bargain, and saves you a dollar or two on purchases, but then randomly just steals $500 from your account.

Image

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentFlow.test.tsx
```diff
@@ -122,6 +121,13 @@ describe("AgentFlow", () => {
...
+        // See: https://github.com/testing-library/user-event/issues/1115#issuecomment-1565730917
+        // @ts-expect-error -- it's an ugly workaround to be removed when the above issue is fixed in RTL.
+        globalThis["jest"] = {
```
dsargent, line 127: There is some deep, unholy coupling between RTL and Jest, so we have to trick it with this fake Jest global. Once the linked bug is fixed, we can remove all this.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentFlow.test.tsx
```diff
@@ -2050,18 +2058,18 @@ describe("AgentFlow", () => {
...
             const promptField = screen.getByPlaceholderText(PROMPT_PLACEHOLDER)
-            await vi.waitFor(async () => localUser.type(promptField, "add a node"))
+            await localUser.type(promptField, "add a node")
 
             const applyButton = screen.getByRole("button", {name: APPLY_BUTTON})
-            await vi.waitFor(async () => localUser.click(applyButton))
+            await localUser.click(applyButton)
 
             // Advance past the 120-second dock apply timeout
-            await act(async () => {
```
dsargent, line 2059: This was pointlessly async. Not even using `advanceTimersByTimeAsync`.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentFlow.test.tsx
```diff
@@ -2050,18 +2058,18 @@ describe("AgentFlow", () => {
             })
 
             const promptField = screen.getByPlaceholderText(PROMPT_PLACEHOLDER)
-            await vi.waitFor(async () => localUser.type(promptField, "add a node"))
+            await localUser.type(promptField, "add a node")
 
             const applyButton = screen.getByRole("button", {name: APPLY_BUTTON})
-            await vi.waitFor(async () => localUser.click(applyButton))
+            await localUser.click(applyButton)
```
dsargent, line 2064: Get rid of the `vi.waitFor` (pointless non-suggestion from Copilot) and the gratuitous `async`-ness. Simpler now.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentFlow.test.tsx
```diff
@@ -2032,9 +2038,11 @@ describe("AgentFlow", () => {
         it("shows a timeout error banner that persists when the dock apply request times out", async () => {
             vi.useFakeTimers()
 
-            // Re-initialise userEvent with advanceTimers so its internal delays stay in sync with
-            // fake timers (the module-level `user` is bound to real timers and would stall here).
-            const localUser = userEvent.setup({advanceTimers: (ms) => vi.advanceTimersByTime(ms)})
+            // Create a custom UserEvent that is synced with Vitest's fake timers
+            const localUser = userEvent.setup({
+                advanceTimers: vi.advanceTimersByTime,
```
dsargent, line 2043: Even simpler way to feed it the fake timers.

Aside, dealing with tests that use fake timers was probably the most difficult part of this whole migration. We should probably figure out better ways of doing that: still covering everything important, while dealing with code that depends on real-time events happening and async timers.

packages/ui-common/__tests__/components/MultiAgentAccelerator/Sidebar/Sidebar.test.tsx
```diff
@@ -136,6 +134,13 @@ describe("SideBar", () => {
     })
 
     beforeEach(() => {
+        // This has nothing to do with Jest itself and everything to do with a bug in React Testing Library.
```
dsargent, line 137: Same deal as before, same bug, same work-around.

packages/ui-common/__tests__/components/MultiAgentAccelerator/MultiAgentAccelerator.test.tsx
```diff
@@ -256,6 +250,13 @@ describe("MultiAgentAccelerator", () => {
     let user: ReturnType<typeof userEvent.setup>
 
     beforeEach(async () => {
+        // This has nothing to do with Jest itself and everything to do with a bug in React Testing Library.
```
dsargent, line 253: Same old bug.

packages/ui-common/__tests__/components/MultiAgentAccelerator/MultiAgentAccelerator.test.tsx
dsargent, line 1109: This test is an absolute beast. Too long to reason about and maintain, fragile. Should be broken up.

packages/ui-common/__tests__/components/MultiAgentAccelerator/MultiAgentAccelerator.test.tsx
```diff
@@ -1148,15 +1153,20 @@ describe("MultiAgentAccelerator", () => {
...
+            const temporaryNetworkNode = document.querySelector(
+                `[data-itemid="${CSS.escape(expectedNetworkName)}"]`
+            ) satisfies HTMLElement
+
             expect(temporaryNetworkNode).not.toBeNull()
 
             const displayAgentName = cleanUpAgentName(TEMPORARY_NETWORK.reservation.reservation_id)
             screen.getByText(displayAgentName)
 
-            await act(async () => {
```
dsargent, line 1157: Copilot suggested using this (instead of simulating a click) to "fix the issue with fake timers". But of course, was a hallucination: the real issue was the bug I linked in the comments, and this just makes it a worse test, so I changed it back to using clicks like a real user would.

packages/ui-common/__tests__/components/MultiAgentAccelerator/MultiAgentAccelerator.test.tsx
```diff
@@ -1170,46 +1180,69 @@ describe("MultiAgentAccelerator", () => {
             chatCommonMock.mockClear()
 
             // First time "expired check" runs, it should not be expired yet
-            await act(async () => {
-                await vi.advanceTimersByTimeAsync(EXPIRED_NETWORKS_CHECK_INTERVAL_MS)
+            act(() => {
```
dsargent, line 1183: Simplify, get rid of Copilot's pointless `async` obsession.

packages/ui-common/__tests__/components/MultiAgentAccelerator/MultiAgentAccelerator.test.tsx
```diff
@@ -1170,46 +1180,69 @@ describe("MultiAgentAccelerator", () => {
...
                 expect.objectContaining({
                     selectedNetwork: expectedNetworkName,
                 })
             )
 
+            // Reset mock calls so we have a clean slate
             chatCommonMock.mockClear()
 
-            // Jump system time past server-side expiration, then run exactly one interval tick.
-            vi.setSystemTime(now + expirationTimeSeconds * 1000 + 1)
```
dsargent, line 1186: Another Copilot suggestion to get around the fake timers/RTL bug. Pointless, so reverted back to how it was, using simple `advanceTimersByTime()`.

packages/ui-common/__tests__/components/MultiAgentAccelerator/MultiAgentAccelerator.test.tsx
```diff
@@ -1170,46 +1180,69 @@ describe("MultiAgentAccelerator", () => {
...
+                )
+            })
 
             // ...but should still be in the store since we're within the grace period
             expect(useTempNetworksStore.getState().tempNetworks.length).toBe(1)
 
-            // Jump system time past grace period, then run exactly one interval tick.
-            vi.setSystemTime(now + expirationTimeSeconds * 1000 + GRACE_PERIOD_MS + 1)
+            // Requery for stability
+            const expiredTemporaryNetworkNode = document.querySelector(
```
dsargent, line 1216: These make the test a little more predictable and less flaky.

packages/ui-common/__tests__/components/MultiAgentAccelerator/MultiAgentAccelerator.test.tsx
dsargent, line 1266: This works, because the Tooltip popover gets portalized in the global DOM, but it's prone to false negatives: the word "Expired" could easily appear at some random place in the DOM by chance. We should improve this, with `within` type queries.

__tests__/common/TestUtils.ts
dsargent, line 1: Same changes as for `strictMocks`.

packages/ui-common/package.json
```diff
@@ -30,8 +30,6 @@
         "zustand": "4.5.7"
     },
     "devDependencies": {
-        "@babel/core": "catalog:default",
```
dsargent, line 33: Not needed -- were used for Jest 🎉

.gitignore
```diff
@@ -7,13 +7,13 @@
 **/tsconfig.tsbuildinfo
 
 # Coverage directory used by tools like lcov
-**/coverage*/
+coverage/
```
dsargent, line 10: Simplify, as the only directory we expect is `<project root>/coverage`.

.yarnrc.yml
```diff
@@ -2,13 +2,9 @@ nodeLinker: node-modules
 
 # Define common dependencies versions to be used throughout the monorepo.
 catalogs:
-    # The idea is that once https://github.com/yarnpkg/berry/issues/6925 we will have a catalog for fixed requirements
-    # and another for requirements ranges, typically used in peerDependencies.
+    # The idea is that once https://github.com/yarnpkg/berry/issues/6925 is fixed we will have a catalog for
+    # fixed requirements and another for requirements ranges, typically used in peerDependencies.
     default:
-        "@babel/core": "7.29.0"
```
dsargent, line 8: Remove Jest-related deps.

babel.jest.config.cjs
dsargent, line 1: No longer needed

jest.config.ts
dsargent, line 1: No longer needed

jest.setup.ts
dsargent, line 1: No longer needed (but `vitest.setup.ts` was copied from this)

jest_quiet.config.ts
dsargent, line 1: No longer needed

knip.config.ts
```diff
@@ -26,53 +26,22 @@ import {config as baseConfig} from "./packages/dev-common/Configs/knip.config"
 const config: KnipConfig = {
     ...baseConfig,
 
-    ignore: [
```
dsargent, line 29: A lot of the disables in here were for Jest stuff, so, no longer required.

Some will probably come back when I add linter plugins for Vitest.

knip.config.ts
```diff
@@ -26,53 +26,22 @@ import {config as baseConfig} from "./packages/dev-common/Configs/knip.config"
...
 
         // Used internally by eslint
         "globals",
 
-        // Used by jest
-        "@babel/core",
-        "@babel/preset-env",
-        "babel-jest",
-
-        // Used for Speech Recognition API types
```
dsargent, line 49: This is now explicitly mentioned in `tsconfig`, so no longer needs to be an exception here.

package.json
```diff
@@ -22,35 +22,23 @@
         "lint-rules": "./build_scripts/eslint_list_rules.sh",
         "lint-summary": "eslint --format summary",
         "start": "yarn workspace @cognizant-ai-lab/neuro-san-web start",
-        "test": "./build_scripts/RunCombinedTests.sh",
-        "test:jest": "jest --config ./jest.config.ts",
-        "test:vitest": "vitest run",
-        "update-coverage:jest": "./build_scripts/UpdateCoverage.sh",
-        "update-coverage:vitest": "./build_scripts/vitest/UpdateCoverage.sh"
+        "test": "vitest run",
```
dsargent, line 25: Simplify/consolidate goals so they're same as before migration.

package.json
```diff
@@ -22,35 +22,23 @@
...
+        "update-coverage": "./build_scripts/UpdateCoverage.sh"
     },
     "packageManager": "yarn@4.12.0",
     "devDependencies": {
-        "@babel/core": "catalog:default",
```
dsargent, line 33: Remove Jest-related deps.

package.json
```diff
@@ -59,16 +47,9 @@
         "eslint-plugin-vitest": "^0.5.4",
         "fake-indexeddb": "catalog:default",
         "globals": "catalog:default",
-        "jest": "catalog:default",
-        "jest-environment-jsdom": "catalog:default",
-        "jest-fail-on-console": "3.3.4",
-        "jest-silent-reporter": "0.6.0",
         "jsdom": "^29.1.1",
         "knip": "6.16.1",
-        "nyc": "catalog:default",
```
dsargent, line 68: Only used during migration, to merge coverage reports.

tsconfig.json
```diff
@@ -28,8 +29,8 @@
         "strict": true,
         "strictNullChecks": false,
         "target": "es2022",
-        "useUnknownInCatchVariables": true,
-        "incremental": true
+        "types": ["vitest/globals", "node", "dom-speech-recognition"],
```
dsargent, line 32: Had to add these to make tsc happy. Copilot said it's the right thing to do 🙄

vitest.config.ts
```diff
@@ -18,16 +15,14 @@ export default defineConfig({
                 "**/knip.config.ts",
                 "**/next-env.d.ts",
                 "**/next.config.ts",
-                "apps/main/app/api/auth/[...nextauth]/route.ts",
+                "**/vitest*.ts",
```
dsargent, line 18: Clean up, fix include and exclude paths for Vitest.

So now we're basically back to the same rules as before migration in here, except that it's Vitest instead of Jest.

vitest.config.ts
```diff
@@ -36,9 +31,9 @@ export default defineConfig({
             },
         },
         environment: "jsdom",
-        exclude: ["node_modules", "dist", ".next", "coverage"],
-        globals: false,
-        include: TESTS_MIGRATED_TO_VITEST,
+        exclude: ["**/node_modules/**", "**/dist/**", "**/.next/**", "**/coverage/**"],
```
dsargent, line 34: Tighten up the patterns and remove migration list.

Also this is where `globals` are enabled.

vitest.setup.ts
dsargent, line 1: I went to town on this file, removing the stuff that is outdated or no longer needed. A lot of it was for earlier versions of Node that lacked implementations of certain functions. Now it's _much_ cleaner and simpler. I verified that without these remaining items, one or other tests fail.

vitest.setup.ts
```diff
@@ -16,30 +16,8 @@ limitations under the License.
...
-
-import {cleanup} from "@testing-library/react"
-// eslint-disable-next-line no-shadow
-import {ReadableStream} from "node:stream/web"
-import {TextDecoder as NodeTextDecoder, TextEncoder as NodeTextEncoder} from "node:util"
-// eslint-disable-next-line no-shadow
-import {afterEach, vi} from "vitest"
 import failOnConsole from "vitest-fail-on-console"
 
-/*
```
dsargent, line 28: Modern Node versions (even ours, 22, which isn't _that_ modern) has these already.

I've been wanting to clean up the obsolete stuff in this file for a while, and the migration seemed like a good opportunity to do so. A heavy setup file is a tax on every single test, so the leaner we can make this file, the better.

vitest.setup.ts
```diff
@@ -88,52 +66,9 @@ Object.defineProperties(global.HTMLElement.prototype, {
     },
 })
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-;(global.SVGElement as any).prototype.getBBox = () => ({
```
dsargent, line 92: Same, Node already has this stuff nowadays.

vitest.setup.ts
```diff
@@ -88,52 +66,9 @@ Object.defineProperties(global.HTMLElement.prototype, {
...
-                matrixTransform: () => this as unknown as DOMPoint,
-                toJSON: () => ({x: len, y: len, z: 0, w: 1}),
-            }
-        }
-    }
-
-    global.SVGPathElement = FakeSVGPathElement as unknown as typeof SVGPathElement
-}
-
-// We've disabled globals for Vitest during migration so have to do this here
```
dsargent, line 132: Opting-in to `globals` means this happens automatically for us. RTL detects the presence of the Vitest globals in the global namespace and automatically handles post test clean-up.

vitest_migration.js
dsargent, line 1: Migration done, so no longer required. 🎉

yarn.lock
dsargent, line 1: Regen'd this due to lots of changes in deps.

But that does raise the issue of: if `yarn.lock` is supposed to be our "deployment manifesto", why does it contain, or care about, dev dependencies? We have a long-standing [ticket](https://github.com/cognizant-ai-lab/neuro-san-ui/issues/183) for this. This gets into the deeper question of, what does `yarn.lock` actually mean to us?

1) We commit to source control
2) We generate it in CI (and if the generated one differs from the committed one, fail the build)
3) At deployment time, we generate it again. If the generated one differs from the committed one, again, fail the build/deployment
4) Else, use the deps specified in the generated one for the generated Docker image.

There seems to be some circularity in the process. We should dig into that.

I guess the idea is that `yarn.lock is a definition of exactly what we will deploy to production`. But since it's such a huge mess of definitions and overrides, none of us ever looks at it, what's the point?

swensel:
> I went the "globals" route to match how we've been doing things with Jest.

I was going to suggest that as well.

swensel (approved):
LGTM. Thanks for doing this.

Issue #462, opened 2026-07-03
Vitest: post-migration optimization


dsargent:
Try to make tests run faster than Jest at least.

PR #463, opened 2026-07-03
#462 Optimize unit tests

dsargent:
## Summary
Optimizations to speed up unit test runs. Important as a "force multiplier" because we (and CI) run tests All. The. Time.

## Testing methodology

Average over 5 runs using `perf stat -r5 yarn test` because Science.

## Some optimizations for the unit tests:

1) Prefer `user.paste()` in tests over `user.type()`. Faster, and fine to use unless we're explicitly testing keyboarding.
2) Create a mock for the heavy MUI Icons when we don't care about the real behavior.
3) Optimize `PlasmaEdge` tests.
4) Fix the GitHub Action which was running tests twice every time.

## Things I tried that had no measurable difference or made things worse

1) [Deps optimizer](https://vitest.dev/config/deps.html#deps-optimizer)
2) Splitting large tests into separate files to enhance parallelism (`ChatCommon`, `AgentFlow`)
3) Setting `watch` to `false` (we already run the tests using `vitest run` which negates `watch`)
4) Enabling [experimental.fsModuleCache](https://vitest.dev/config/experimental.html#experimental-fsmodulecache). I was surprised that one didn't help, but the data don't lie.
5) Tweaking `maxWorkers` higher, lower, leaving it at default: 8 was the sweet spot for my machine. I let CI default because with `8`, tests were failing in CI probably due to concurrency issues.
6) Change `pool` to `threads` rather than the default `forks`.
7) `maxConcurrency: 8`

## Things I couldn't try
1) `isolation` false: too many tests break.
2) `happy-dom` instead of `jsdom` (reputedly 3x faster): too many tests break.

Suggestion from Copilot (which aligns with my comment in an earlier PR that the Dock should be its own component):

> The Network Editor tests are testing dock behavior, but every test pays full graph render cost. Extracting the dock into AgentFlowNetworkEditorDock would let those tests render only the dock.

Also: Move unit test suites where they belong in preparation for splitting test envs Node vs JSDom.


.github/workflows/test.yml
```diff
@@ -131,5 +131,5 @@ jobs:
               if: success() || failure()
               run: |
                   node --version
-                  yarn test --version
+                  yarn vitest --version
```
dsargent, line 134: Oops...this was running the tests twice. Fixed.

__tests__/common/MuiIconsMock.tsx
dsargent, line 1: Part  of the optimization. A new mock (suggested/vibed by Copilot) for the heavy MUI Icons dep.

We pay a high price for the `import * from @mui/icons` that we do, in order to have it so the LLM can dynamically suggest icons.

packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
```diff
@@ -156,7 +156,8 @@ describe("ChatCommon", () => {
         const userQueryInput = screen.getByPlaceholderText(`Chat with ${agent}`)
 
         // Type a query
-        await user.type(userQueryInput, query)
+        await user.click(userQueryInput)
```
dsargent, line 159: Throughout, changed `user.type` to `user.click` + `user.paste`. The rationale being, it's faster (as in, measurably), and we don't really need `user.type()` unless we're testing some specific keyboarding behavior, like hitting shift-enter in an text input or something like that. Mostly we don't care about that and just want to input the text by whatever means.

The next step would be our old frenemy `fireEvent`, which is probably even faster, but yet one small step away from how real users interact.

packages/ui-common/__tests__/components/Common/CustomerLogo.test.tsx
```diff
@@ -34,7 +35,7 @@ describe("CustomerLogo", () => {
             <CustomerLogo
                 customer="Test Customer"
                 logoSource="generic"
-                iconSuggestion="InvalidIconName" // Invalid MUI icon name
+                iconSuggestion={INVALID_MUI_ICON_NAME}
```
dsargent, line 38: Use standard "invalid icon" value.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentFlow.test.tsx
dsargent, line 1: Replace `user.type()` with `user.click()` + `user.paste()`.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentNode.test.tsx
```diff
@@ -193,7 +194,7 @@ describe("AgentNode", () => {
     })
 
     it("warns and falls back to default icon when agentIconSuggestion is invalid", async () => {
-        renderAgentNode({agentIconSuggestion: "NotARealMuiIconAtAll"})
+        renderAgentNode({agentIconSuggestion: INVALID_MUI_ICON_NAME})
```
dsargent, line 197: Use standard "invalid icon" constant.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/PlasmaEdge.test.tsx
dsargent, line 1: This was a slow test as it was actually awaiting the animation "real time". I had Copilot rewrite it and now it's super fast with no loss of coverage. I still don't know how it works because I know zilch about animation.

packages/ui-common/__tests__/components/MultiAgentAccelerator/Sidebar/Sidebar.test.tsx
```diff
@@ -112,6 +113,9 @@ const DEFAULT_PROPS: SidebarProps = {
     setSelectedNetwork: setSelectedNetworkMock,
 }
 
+// Not available in JSDom. See: https://github.com/jsdom/jsdom/issues/1695
```
dsargent, line 116: Moved to here from `Setup` as only a couple of tests need this mock.

packages/ui-common/__tests__/components/MultiAgentAccelerator/Sidebar/Sidebar.test.tsx
```diff
@@ -335,7 +339,7 @@ describe("SideBar", () => {
             // Override networkIconSuggestions to include an invalid icon name for TEST_AGENT_MUSIC_NERD
             networkIconSuggestions: {
                 ...NETWORK_ICON_SUGGESTIONS,
-                [`${TEST_AGENTS_FOLDER}/${TEST_AGENT_MUSIC_NERD}`]: "NonExistentIcon",
+                [`${TEST_AGENTS_FOLDER}/${TEST_AGENT_MUSIC_NERD}`]: INVALID_MUI_ICON_NAME,
```
dsargent, line 342: Use standard "invalid icon" constant.

packages/ui-common/__tests__/components/MultiAgentAccelerator/MultiAgentAccelerator.test.tsx
```diff
@@ -154,6 +154,9 @@ vi.mock("../../../components/MultiAgentAccelerator/Sidebar/ImportNetworkModal",
     }
 })
 
+// Not available in JSDom. See: https://github.com/jsdom/jsdom/issues/1695
```
dsargent, line 157: Moved to here from `Setup` as only a couple of tests need this mock.

packages/ui-common/__tests__/unit/MultiAgentAccelerator/AgentConversations.test.ts
dsargent, line 1: Moved these 3 suites under `unit` because they are in fact unit tests. At some point we may want to split tests by unit vs component, since, currently, we fire up `JSDom` for _all_ tests and that's expensive, rather than, say, just running the unit tests in the `node` env. But in the scheme of things, much smaller effect than the other optimizations in this PR, so I didn't do it this time.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow/PlasmaEdge.tsx
dsargent, line 1: Refactoring to make it more testable: make particle count values params so we can force them to low values in tests.

.yarnrc.yml
```diff
@@ -17,6 +17,7 @@ catalogs:
         "@types/node": "22.17.1"
         "@types/react": "19.2.14"
         "@types/react-dom": "19.2.3"
+        "@vitest/utils": "4.1.10"
```
dsargent, line 20: Fix the `yarn install` warnings.

package.json
```diff
@@ -32,8 +32,9 @@
         "@testing-library/jest-dom": "6.9.1",
         "@types/dom-speech-recognition": "0.0.6",
         "@types/node": "catalog:default",
-        "@vitejs/plugin-react": "^6.0.3",
```
dsargent, line 35: We don't need this plugin for now.

package.json
```diff
@@ -32,8 +32,9 @@
...
+        "@types/react": "catalog:default",
```
dsargent, line 35: ...but I did have to add a few deps because we're actually using these from the root when running `vitest`, so yarn was (rightfully) complaining.

package.json
```diff
@@ -50,8 +51,10 @@
         "jsdom": "^29.1.1",
         "knip": "6.16.1",
         "prettier": "3.8.4",
+        "react": "catalog:default",
         "typescript": "catalog:default",
         "typescript-eslint": "8.61.0",
+        "vite": "catalog:default",
```
dsargent, line 57: It hurt my soul to add `vite` here as a dep, but `vitest` implicitly depends on it.

vitest.config.ts
```diff
@@ -1,8 +1,6 @@
-import react from "@vitejs/plugin-react"
```
dsargent, line 1: Plugin not needed currently.

vitest.config.ts
```diff
@@ -30,10 +28,16 @@ export default defineConfig({
                 lines: -76,
             },
         },
+        // TODO: potential small optimization: consider using `node` environment for non-UI tests
```
dsargent, line 31: See comment

vitest.config.ts
```diff
@@ -30,10 +28,16 @@ export default defineConfig({
...
         environment: "jsdom",
         exclude: ["**/node_modules/**", "**/dist/**", "**/.next/**", "**/coverage/**"],
         globals: true,
-        include: ["**/__tests__/**/*.test.{ts,tsx}"],
+        // Let maxWorkers default to the number of CPU cores in CI, but set 8 for local development (best by test)
+        maxWorkers: process.env["CI"] === "true" ? undefined : 8,
```
dsargent: See comments

vitest.config.ts
```diff
@@ -1,8 +1,6 @@
-import react from "@vitejs/plugin-react"
 import {defineConfig} from "vitest/config"
 
 export default defineConfig({
-    plugins: [react()],
     test: {
         coverage: {
```
dsargent, line 5: Coverage numbers (reassuringly) not affected by this PR.

vitest.setup.ts
```diff
@@ -32,15 +33,17 @@ class ResizeObserver {
         this.callback([{target} as globalThis.ResizeObserverEntry], this)
     }
 
-    // Just a stub for testing so disable warnings
-    /* eslint-disable-next-line
-        no-empty-function,@typescript-eslint/no-empty-function,@typescript-eslint/class-methods-use-this */
-    unobserve() {}
+    /* Just a stub for testing so disable warnings */
```
dsargent, line 36: Giving it an empty body with a comment at least placates a couple of the warnings.

vitest.setup.ts
```diff
@@ -66,9 +69,6 @@ Object.defineProperties(global.HTMLElement.prototype, {
     },
 })
 
-// Not available in JSDom. See: https://github.com/jsdom/jsdom/issues/1695
```
dsargent, line 69: Removed. Only used in a couple of tests so they can mock it as needed.

__tests__/common/MuiIconsMock.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+    const getMockExport = (target: typeof moduleBase, prop: PropertyKey) => {
+        if (prop in target) {
+            return target[prop as keyof typeof moduleBase]
+        }
+
+        return typeof prop === "string" ? getMockIcon(prop) : undefined
+    }
+
+    return new Proxy(moduleBase, {
```
swensel, line 52: Interesting use of `Proxy`. 👍

dsargent: And also the `get`/`has` property functions. I didn't even know those were a thing (though they immediately rang a bell with C#, unsurprisingly).

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/PlasmaEdge.test.tsx
```diff
@@ -38,71 +54,68 @@ describe("PlasmaEdge", () => {
...
-                cb(0)
-            } catch {
-                // ignore
+        let rafCallback: FrameRequestCallback | undefined
+
+        try {
+            HTMLCanvasElement.prototype.getContext = vi.fn(
+                () => fakeCtx as CanvasRenderingContext2D
+            ) as unknown as typeof HTMLCanvasElement.prototype.getContext
+            ;(Element.prototype as unknown as {getTotalLength?: () => number}).getTotalLength = vi.fn(() => 100)
```
swensel, line 71: Bunch of `as unknown`.

swensel (approved):
LGTM. Nice improvements! 🎉

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/PlasmaEdge.test.tsx
dsargent: Yup, but the old test had just as many if not more.

It makes me nervous that `PlasmaEdge` itself is vibed, as is the test. It's a blind spot for me. I know almost nothing about 2D animation in the browser, so very reliant on Copilot for maintenance of this particular black hole in the code base. Not sure what to do about that, other than take classes in 2D browser animation 🤷🏻

Issue #441, opened 2026-06-26
byok use case: ui won't start without OPENAI_API_KEY set


donn-leaf:
In order to get a byok instance running, I had to put in a placeholder value for OPENAI_API_KEY.

Desired: ui should start without this key, at least for the byok case.

PR #464, opened 2026-07-06
#441 Make OpenAI API key optional

dsargent:
Move `OPENAI_API_KEY` env var to the "optional" section so that MAUI will still start without it, but with degraded functionality.


swensel (approved):
LGTM.

donn-leaf (approved):
Thanks

PR #467, opened 2026-07-07
Fix build failure caused by duelling @mui/system versions

dsargent:
For posterity, the issue was that we were installing two different copies of `@mui/system` due to dependency resolution issues. We were getting away with it before because the versions were all resolving down to the same one, `9.0.1` in `yarn.lock`, but we were living on borrowed time because the actual internal dependency of `@mui/material` was always pointing to `@mui/system` version `9.1.2`, despite what we specified in our catalog:

```
direct workspace/package dependency: @mui/system@9.0.1
transitive from @mui/material@9.0.1: @mui/system@9.1.2
```

We were getting away with it by the skin of our teeth because of `yarn.lock`, but as soon as I regen'd that file as part of the Vitest updates, all hell broke loose.

Our yarn catalog was calling for `9.0.1` but in fact at build time we were resolving to `9.1.2` transitively via `@mui/material`. I wish we could just not have `@mui/system` as a dependency at all and just let the other MUI libraries pull it in transitively, but if we don't specify it, yarn complains that `@mui/x-tree-view` requires it, so we're stuck.

The solution was to modify the version in the catalog so it reflects the same version that we are actually resolving to at build time. This is not an upgrade, but a realignment of the catalog to reflect what has been happening in reality.

My explanation above wasn't that great; I asked Copilot to do better:

> @mui/material@9.0.1 depends on @mui/system@^9.0.1, and after the lockfile changed, Yarn re-resolved that range from 9.0.1 to 9.1.2, while ui-common was still directly providing @mui/system@9.0.1; that created two MUI System type identities, which only surfaced during declaration emit in build:lib.
Updating the catalog so the direct @mui/system provision also resolves to 9.1.2 removed the duplicate version/type split: the repo now uses a single @mui/system version, matching what MUI Material was already choosing transitively.

We can consider adding the build step to our `test` goals for CI; it would catch weird edge cases like this, for the cost of a little extra time on every build. But only 11 seconds currently:

```
❯ time yarn build:lib 
✨ openapi-typescript 7.8.0
🚀 https://neuro-san-dev.decisionai.ml/api/v1/docs → ./generated/neuro-san/NeuroSanClient.ts [463.5ms]
yarn build:lib  11.08s user 1.48s system 157% cpu 7.963 total
```

On the other hand, it is super rare that something passes all our CI checks then fails at `yarn build:lib` time. Not impossible, as we see here, but hugely unlikely. Similar to Docker build failures: very rare that something sneaks past our CI and causes issues at Docker build time, but not unheard-of. However, Docker builds take a long time -- O(several minutes) -- so I wouldn't want to add that tax to every build.

Also move MUI deps from deps to devDeps as recommended by Copilot.

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -8,7 +8,7 @@ catalogs:
         "@langchain/core": "1.1.17"
         "@mui/icons-material": "9.0.1"
         "@mui/material": "9.0.1"
-        "@mui/system": "9.0.1"
+        "@mui/system": "9.1.2"
         "@mui/x-tree-view": "9.1.0"
         "@testing-library/dom": "10.4.1"
         "@testing-library/react": "16.3.0"
--- a/packages/ui-common/package.json
+++ b/packages/ui-common/package.json
@@ -12,10 +12,6 @@
     "dependencies": {
         "@dagrejs/dagre": "1.0.4",
         "@langchain/core": "catalog:default",
-        "@mui/icons-material": "catalog:default",
-        "@mui/material": "catalog:default",
-        "@mui/system": "catalog:default",
-        "@mui/x-tree-view": "catalog:default",
         "@xyflow/react": "catalog:default",
         "http-status": "catalog:default",
         "lodash-es": "catalog:default",
@@ -30,6 +26,10 @@
         "zustand": "4.5.7"
     },
     "devDependencies": {
+        "@mui/icons-material": "catalog:default",
+        "@mui/material": "catalog:default",
+        "@mui/system": "catalog:default",
+        "@mui/x-tree-view": "catalog:default",
         "@testing-library/dom": "catalog:default",
         "@testing-library/react": "catalog:default",
         "@testing-library/user-event": "catalog:default",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1495,7 +1495,7 @@ __metadata:
   languageName: node
   linkType: hard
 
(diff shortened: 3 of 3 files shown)


packages/ui-common/package.json
```diff
@@ -12,10 +12,6 @@
     "dependencies": {
         "@dagrejs/dagre": "1.0.4",
         "@langchain/core": "catalog:default",
-        "@mui/icons-material": "catalog:default",
```
dsargent, line 15: Copilot suggested moving these to dev deps, and I agree: defeats the purpose of having them as peer deps if we're _also_ going to have them as regular deps.

.yarnrc.yml
```diff
@@ -8,7 +8,7 @@ catalogs:
         "@langchain/core": "1.1.17"
         "@mui/icons-material": "9.0.1"
         "@mui/material": "9.0.1"
-        "@mui/system": "9.0.1"
+        "@mui/system": "9.1.2"
```
dsargent, line 11: This is the fix for the build issue: sync the version up in our catalog with what yarn was choosing anyway.

yarn.lock
```diff
@@ -1535,35 +1535,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@mui/system@npm:9.0.1":
```
dsargent, line 1538: The fact that this block is deleted is a good sign of our dependency health; we no longer force-depend on this version of `@mui/system`.

swensel (approved):
LGTM. Good find.

PR #469, opened 2026-07-07
PR feedback follow up (#393) and add more test coverage (#287).

swensel:
PR feedback from https://github.com/cognizant-ai-lab/neuro-san-ui/pull/357, https://github.com/cognizant-ai-lab/neuro-san-ui/pull/321, and https://github.com/cognizant-ai-lab/neuro-san-ui/pull/384.

Add more test coverage (#287). I did try to split this off, but it was linked to some of the changes. The one clean split was #470.

Also some other clean up, optimization, tech debt included in this this PR (i.e. switching more `fireEvent` to `userEvent` and `user.type` to `user.paste` in tests, reducing copy and pasting, enabling keyboard access, and more).

No screenshots. UI should look the same. These are behind the scenes improvements. I could include a video of keyboard actions but I don't think that would translate super well.

PR #470, opened 2026-07-07
Improve test coverage in Agent.test.ts‎ (#287).

swensel:
# Description

Improve test coverage in `Agent.test.ts‎` (#287).


packages/ui-common/__tests__/unit/Controller/agent/Agent.test.ts
```diff
@@ -75,6 +77,31 @@ describe("Controller/Agent/testConnection", () => {
         const result: TestConnectionResult = await testConnection("www.example.com")
         expect(result.success).toBe(false)
     })
+
+    it("Should abort and report failure when the request exceeds the timeout", async () => {
```
swensel, line 98: Timeout exceeded case.

packages/ui-common/__tests__/unit/Controller/agent/Agent.test.ts
```diff
@@ -218,6 +245,21 @@ describe("Controller/Agent/getConnectivity", () => {
         // Make sure getConnectivity logged the error to the console
         expect(errorSpy).toHaveBeenCalled()
     })
+
+    it("Should strip the temporary/ prefix from the network name before calling the server", async () => {
+        global.fetch = mockFetch({connections: []})
```
swensel, line 267: Test to strip the `temporary/` prefix before calling the server. 

We don't actually add `${TEMPORARY_NETWORK_FOLDER}/`, except here in this test, in order to mock what would be sent in the actual code for a selected network.

packages/ui-common/__tests__/unit/Controller/agent/Agent.test.ts
```diff
@@ -336,3 +378,81 @@ describe("Controller/Agent/sendNetworkDesignerRequest", () => {
...
+                }),
+            })
+        )
+    })
+
+    it("getBrandingSuggestions POSTs the company name and returns the suggestions", async () => {
+        const suggestions = {primary: "#123456"}
+        global.fetch = mockFetch(suggestions)
+
+        const result = await getBrandingSuggestions("Acme")
```
swensel, line 419: Test `getBrandingSuggestions` function.

packages/ui-common/__tests__/unit/Controller/agent/IconSuggestions.test.ts
```diff
@@ -47,6 +47,14 @@ describe("Controller/Agent/getNetworkIconSuggestions", () => {
 
         expect(result).toEqual(mockSuggestions)
         expect(global.fetch).toHaveBeenCalledTimes(1)
+        expect(global.fetch).toHaveBeenCalledWith(
+            "/api/networkIconSuggestions",
```
swensel, line 50: Verify that `/api/networkIconSuggestions` is what is called.

packages/ui-common/__tests__/unit/Controller/agent/IconSuggestions.test.ts
```diff
@@ -67,11 +75,22 @@ describe("Controller/Agent/getNetworkIconSuggestions", () => {
...
                 },
             ],
+            metadata: {foo: "bar"},
         }
         const result = await getAgentIconSuggestions(connectivityInfo)
 
         expect(result).toEqual(mockSuggestions)
         expect(global.fetch).toHaveBeenCalledTimes(1)
+        expect(global.fetch).toHaveBeenCalledWith(
+            "/api/agentIconSuggestions",
```
swensel, line 85: Verify that `/api/agentIconSuggestions` is what is called.

packages/ui-common/__tests__/unit/Controller/agent/IconSuggestions.test.ts
```diff
@@ -67,11 +75,22 @@ describe("Controller/Agent/getNetworkIconSuggestions", () => {
                     origin: "other_agent",
                 },
             ],
+            metadata: {foo: "bar"},
```
swensel, line 79: Pass metadata to verify that the API is called with that.

packages/ui-common/__tests__/unit/Controller/agent/IconSuggestions.test.ts
```diff
@@ -67,11 +75,22 @@ describe("Controller/Agent/getNetworkIconSuggestions", () => {
...
 
         expect(result).toEqual(mockSuggestions)
         expect(global.fetch).toHaveBeenCalledTimes(1)
+        expect(global.fetch).toHaveBeenCalledWith(
+            "/api/agentIconSuggestions",
+            expect.objectContaining({
+                method: "POST",
+                body: JSON.stringify({
+                    connectivity_info: connectivityInfo.connectivity_info,
+                    metadata: connectivityInfo.metadata,
```
swensel, line 90: Verify metadata as well.

packages/ui-common/__tests__/unit/Controller/agent/Agent.test.ts
```diff
@@ -75,6 +76,31 @@ describe("Controller/Agent/testConnection", () => {
...
+
+        // Fire the 2.5s timeout, which triggers controller.abort() → fetch rejects.
+        vi.advanceTimersByTime(2500)
+
+        const result = await resultPromise
+        expect(result.success).toBe(false)
+        expect(result.status).toContain("aborted")
+
+        vi.useRealTimers()
+    })
```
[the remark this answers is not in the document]
swensel: I'll add a try/finally to this test.

vitest.config.ts
```diff
@@ -22,10 +22,10 @@ export default defineConfig({
             provider: "istanbul",
             reporter: ["text-summary"],
             thresholds: {
-                statements: -103,
-                branches: -154,
-                functions: -25,
-                lines: -74,
+                statements: -101,
```
swensel, line 25: Small improvement. This was originally part of https://github.com/cognizant-ai-lab/neuro-san-ui/pull/394 and was split off.

I ended up with a lot more [improvement](https://github.com/cognizant-ai-lab/neuro-san-ui/pull/469/changes#diff-2ee894bf23aa44ff4ce12a3da8af19ab20180474ebf2176dbe5f2eea3f96dc92R25) in #469.

packages/ui-common/__tests__/unit/Controller/agent/Agent.test.ts
```diff
@@ -49,6 +50,14 @@ let oldFetch: typeof global.fetch
 describe("Controller/Agent/testConnection", () => {
     withStrictMocks()
 
+    beforeEach(() => {
```
dsargent, line 53: We could consider adding these to `strictMocks`.

packages/ui-common/__tests__/unit/Controller/agent/Agent.test.ts
```diff
@@ -85,6 +94,33 @@ describe("Controller/Agent/testConnection", () => {
...
+            const resultPromise = testConnection("https://slow.example.com")
+
+            // Fire the 2.5s timeout, which triggers controller.abort() → fetch rejects.
+            vi.advanceTimersByTime(2500)
+
+            const result = await resultPromise
+            expect(result.success).toBe(false)
+            expect(result.statusText).toContain("aborted")
+        } finally {
+            vi.useRealTimers()
```
dsargent: No need, done in `beforeEach` in `withStrictMocks()`.

packages/ui-common/__tests__/unit/Controller/agent/Agent.test.ts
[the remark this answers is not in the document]
dsargent: But you addressed the part that didn't need to be (`useRealTimers()`) but didn't address the part that in theory _did_ need to be, namely the `global.fetch` overwrite?!

packages/ui-common/__tests__/unit/Controller/agent/Agent.test.ts
```diff
@@ -228,6 +264,21 @@ describe("Controller/Agent/getConnectivity", () => {
         // Make sure getConnectivity logged the error to the console
         expect(errorSpy).toHaveBeenCalled()
     })
+
+    it("Should strip the temporary/ prefix from the network name before calling the server", async () => {
```
dsargent, line 266: This is a good idea for a test -- making sure our local abstraction doesn't "leak" into outbound API calls.

packages/ui-common/__tests__/unit/Controller/agent/Agent.test.ts
```diff
@@ -228,6 +264,21 @@ describe("Controller/Agent/getConnectivity", () => {
...
+        global.fetch = mockFetch({connections: []})
+
+        await getConnectivity(
+            NEURO_SAN_EXAMPLE_URL,
+            `${TEMPORARY_NETWORK_FOLDER}/${TEST_AGENT_MATH_GUY}`,
+            TEST_USERNAME
+        )
+
+        // The server doesn't know about the "temporary/" UI convention, so it must be removed from the path.
+        const calledUrl = vi.mocked(global.fetch).mock.calls[0][0] as string
```
dsargent: Usually digging into the bowels of `Mock.calls` is an absolute last resort. It's fragile, makes assumptions, and is hard to read. 

Almost always better to use the mocks for what they are intended:

```typescript
expect(global.fetch).toHaveBeenCalledWith(
    expect.stringContaining(TEST_AGENT_MATH_GUY),
    expect.anything()
)
expect(global.fetch).not.toHaveBeenCalledWith(
    expect.stringContaining(`${TEMPORARY_NETWORK_FOLDER}/`),
    expect.anything()
)
```

packages/ui-common/__tests__/unit/Controller/agent/Agent.test.ts
```diff
@@ -346,3 +397,28 @@ describe("Controller/Agent/sendNetworkDesignerRequest", () => {
...
+
+    beforeEach(() => {
+        oldFetch = global.fetch
+    })
+
+    afterEach(() => {
+        global.fetch = oldFetch
+    })
+
+    it("getBrandingSuggestions POSTs the company name and returns the suggestions", async () => {
```
dsargent, line 415: Thanks for adding tests for this, even though I wrote the implementation. In general, we should probably each focus on backfilling test coverage on our own PRs, though I do appreciate the "assist".

packages/ui-common/__tests__/unit/Controller/agent/Agent.test.ts
```diff
@@ -346,3 +397,28 @@ describe("Controller/Agent/sendNetworkDesignerRequest", () => {
...
+        const suggestions = {primary: "#123456"}
+        global.fetch = mockFetch(suggestions)
+
+        const result = await getBrandingSuggestions("Acme")
+
+        expect(result).toEqual(suggestions)
+        expect(global.fetch).toHaveBeenCalledWith(
```
dsargent: Or even tighter: `toHaveBeenCalledExactlyOnceWith(...)`

In general, it's good to use the "strongest possible assertion that should be true", without veering into "testing unimportant details". Yes, it's a balancing act.

packages/ui-common/__tests__/unit/Controller/agent/IconSuggestions.test.ts
dsargent: You did it correctly here, rather than digging into `mock.calls[0][0]` stuff.

packages/ui-common/__tests__/unit/Controller/agent/IconSuggestions.test.ts
dsargent: Good that you strengthened the assertions here.

Perhaps even more:

```typescript
// Can remove this line
//         expect(global.fetch).toHaveBeenCalledTimes(1)

        expect(global.fetch).toHaveBeenCalledExactlyOnceWith(
            "/api/agentIconSuggestions",
            expect.objectContaining({
                method: "POST",
                body: JSON.stringify({
                    connectivity_info: connectivityInfo.connectivity_info,
                    metadata: connectivityInfo.metadata,
                }),
            })
        )

        vi.mocked(global.fetch).mockClear()

        // Second time should be cached
        const cachedResult = await getAgentIconSuggestions(connectivityInfo)
        expect(cachedResult).toEqual(mockSuggestions)
        expect(global.fetch).toHaveBeenCalledTimes(0) // No additional fetch calls
```

Here I've used my favorite `toHaveBeenCalledExactlyOnceWith` and reset the mock so it's clearer in the second assertion - Times(0).

dsargent (dismissed):
Some minor comments on style/approach, but no blockers. Thanks for improving test coverage in meaningful ways.

vitest.config.ts
```diff
@@ -22,10 +22,10 @@ export default defineConfig({
...
             reporter: ["text-summary"],
             thresholds: {
-                statements: -104,
-                branches: -155,
-                functions: -25,
-                lines: -76,
+                statements: -102,
+                branches: -153,
+                functions: -24,
+                lines: -74,
```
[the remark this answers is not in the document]
dsargent: Copilot -- tell me you don't understand how coverage thresholds work without saying you don't understand how coverage thresholds work.

packages/ui-common/__tests__/unit/Controller/agent/Agent.test.ts
swensel: Hmm, I may hesitate to do that since they aren't applicable for every test, but it is an interesting thing to consider. I suppose it could be a boolean passed in `options` to `withStrictMocks`. Not sure.

packages/ui-common/__tests__/unit/Controller/agent/Agent.test.ts
swensel: I actually had added this try/finally due to a Copilot suggestion. Anyway, removing it.

packages/ui-common/__tests__/unit/Controller/agent/IconSuggestions.test.ts
swensel: I'll remember to use `toHaveBeenCalledExactlyOnceWith`, where applicable, in the future, and I've made updates in this PR to use that, instead.

packages/ui-common/__tests__/unit/Controller/agent/Agent.test.ts
swensel: I see you added another [comment](https://github.com/cognizant-ai-lab/neuro-san-ui/pull/470#discussion_r3574946867) about this. I'll look into that.

dsargent: Copilot doesn't always dig deep enough to understand that `strictMocks` does this. We still have to be in the loop.

packages/ui-common/__tests__/unit/Controller/agent/Agent.test.ts
dsargent: It's borderline. Not all tests care about `fetch`, and for those that don't, it's harmless and negligible overhead. For those that _do_ care about `fetch`, I don't want them to be making assumptions about the mock replacement being persisted across tests.

I could go either way on this one.

packages/ui-common/__tests__/unit/Controller/agent/IconSuggestions.test.ts
dsargent: It's always good to ask yourself, "what behavior do I care about testing here and under what circumstances would I want this test to fail?"

React goes through an extra render cycle with my component -- probably not. Lots of harmless things cause extra renders. Usually not an issue.

The test logic results in my "fetch" function being called an extra time? Yes, I probably want the test to tell me about that and fail.

The exact test of the "abort" message changes? Probably don't care.
Something changed in the code, and the abort exception no longer gets thrown? Absolutely, I want that to fail, and _loudly_.

packages/ui-common/__tests__/unit/Controller/agent/Agent.test.ts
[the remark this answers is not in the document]
swensel: This should restore `global.fetch` though?

```
afterEach(() => {
    global.fetch = oldFetch
})
```

packages/ui-common/__tests__/unit/Controller/agent/Agent.test.ts
swensel: I'm also on the fence. I do see the potential benefit.

packages/ui-common/__tests__/unit/Controller/agent/Agent.test.ts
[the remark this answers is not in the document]
dsargent: Yes, that'll work. I didn't realize you'd put one in `finally` and the other in `afterEach`.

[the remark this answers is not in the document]
dsargent: This type of thing is going to be more important as I'd like to get to the point where we can run tests with `isolation: false` (much faster), which means they have to be diligent at housekeeping things like this.

dsargent (approved):

Issue #472, opened 2026-07-08
Error when running tests: Not implemented: navigation to another Document


dsargent:
Above error started showing up after migration to Vitest.

Explanation: JSDom doesn't support navigation. But our tests exercise it anyway. The message was somehow not surfaced by Jest, but was always an issue.

PR #473, opened 2026-07-08
#472 Fix error message during tests

dsargent:
Fixes this error that was showing up during unit test runs:

`Not implemented: navigation to another Document`

Emitted by JSDom, which doesn't support browser navigation (and refuses to pretend it does).

Reference: https://github.com/jsdom/jsdom/issues/2112

### DIFF
--- a/packages/ui-common/__tests__/components/Common/Navbar.test.tsx
+++ b/packages/ui-common/__tests__/components/Common/Navbar.test.tsx
@@ -99,7 +99,7 @@ describe("Navbar", () => {
     })
 
     it("should redirect to email client when confirmation is clicked", async () => {
-        vi.spyOn(BrowserNavigation, "navigateToUrl")
+        vi.spyOn(BrowserNavigation, "navigateToUrl").mockImplementation(vi.fn())
 
         renderNavbar()
 
--- a/packages/ui-common/utils/BrowserNavigation.ts
+++ b/packages/ui-common/utils/BrowserNavigation.ts
@@ -18,6 +18,8 @@ limitations under the License.
  * Navigate to the specified URL by setting window.location.href
  * @param url The URL to navigate to
  */
+// See: https://github.com/jsdom/jsdom/issues/2112
+/* istanbul ignore next -- JSDom doesn't support simulated navigation */
 export const navigateToUrl = (url: string): void => {
     window.location.href = url
 }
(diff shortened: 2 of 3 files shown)


packages/ui-common/__tests__/components/Common/Navbar.test.tsx
```diff
@@ -99,7 +99,7 @@ describe("Navbar", () => {
     })
 
     it("should redirect to email client when confirmation is clicked", async () => {
-        vi.spyOn(BrowserNavigation, "navigateToUrl")
+        vi.spyOn(BrowserNavigation, "navigateToUrl").mockImplementation(vi.fn())
```
dsargent, line 102: This is the fix. Mock out the implementation so we don't attempt to navigate to another URL during test execution, which JSDom doesn't like.

packages/ui-common/utils/BrowserNavigation.ts
```diff
@@ -18,6 +18,8 @@ limitations under the License.
  * Navigate to the specified URL by setting window.location.href
  * @param url The URL to navigate to
  */
+// See: https://github.com/jsdom/jsdom/issues/2112
+/* istanbul ignore next -- JSDom doesn't support simulated navigation */
```
dsargent, line 22: Very rare that I would want to exclude something from coverage analysis, but this fits the bill: since JSDom doesn't support navigation, and has no plans to, we literally can't cover this. Or rather we can, but it just generates a JSDom error. We _can_, and do, assert that we at least called the mock version in the tests.

vitest.config.ts
```diff
@@ -32,8 +32,8 @@ export default defineConfig({
         environment: "jsdom",
         exclude: ["**/node_modules/**", "**/dist/**", "**/.next/**", "**/coverage/**"],
         globals: true,
-        // Let maxWorkers default to the number of CPU cores in CI, but set 8 for local development (best by test)
         include: ["apps/**/__tests__/**/*.test.{ts,tsx}", "packages/**/__tests__/**/*.test.{ts,tsx}"],
+        // Let maxWorkers default to the number of CPU cores in CI, but set 8 for local development (best by test)
```
dsargent, line 36: Moved comment next to line it pertains to.

swensel (approved):
LGTM. Thanks for fixing this.

dsargent:
Looks like GitHub CodeQL is having a moment. It's hung. This passes the regular CI checks and is approved so I'll force-merge.

Issue #213, opened 2026-03-12
[UN-3665] Upgrade MAUI to Node.js 24


dsargent:
**Migrated from Jira:** [UN-3665](https://leaf-ai.atlassian.net/browse/UN-3665)

## Jira Details
- **Issue Type:** Task
- **Priority:** Major
- **Status:** Nice to Haves
- **Reporter:** Sean Wensel
- **Assignee:** Sean Wensel
- **Created:** 2026-01-26 13:50:49
- **Updated:** 2026-02-02 14:13:58

### Comment by Sean Wensel (2026-02-02 14:11:09)
MAUI is tracked in [https://github.com/cognizant-ai-lab/neuro-san-ui/issues/132](https://github.com/cognizant-ai-lab/neuro-san-ui/issues/132)

PR #476, opened 2026-07-09
#213 #250 Upgrade to Node v26 and enforce Node version

dsargent:
Upgrades for Node v26. We are currently on v22 so this is an upgrade of two major versions.

Also adds a mechanism to enforce the Node.js version, via a `yarn` plugin, since yarn no longer enforces `engines` in `package.json` due to a bug (see reference in files details).

This release is backward compatible: it will accept any version of Node matching ^22 or ^26 (but not others). Once we are migrated and CI is updated, we can remove ^22 support.

---

Regarding Docker: see comments on `Dockerfile`. It will continue working for v22, but some legwork required for upgrading it to v26. That can be done on our schedule, when time permits.


.yarn/plugins/CheckNodeVersion.mjs
dsargent, line 1: New yarn plugin to check for correct Node.js version. See header comment.

packages/dev-common/Configs/knip.config.ts
dsargent, line 1: Freebie: I noticed we weren't up to date on our Knip checks. This makes sure that tsc will keep us honest; if knip ever adds new checks, or more likely, we upgrade Knip but neglect to opt-in to its new checks (like `cycle` in this case), tsc will scream.

packages/dev-common/Configs/knip.config.ts
```diff
@@ -28,6 +28,26 @@ import type {KnipConfig} from "knip"
...
+    exports: true,
+    files: true,
+    namespaceMembers: true,
+    nsExports: true,
+    nsTypes: true,
+    optionalPeerDependencies: true,
+    types: true,
+    unlisted: true,
+    unresolved: true,
+} satisfies Record<IssueType, true>) as IssueType[]
```
dsargent, line 49: Yup, `as`. Unavoidable here as `Object.keys()` is old school and just returns `string[]` with no way to convince it to do otherwise. If only it admitted a generic param...

packages/dev-common/Configs/knip.config.ts
```diff
@@ -38,25 +58,9 @@ export const config: RawConfiguration = {
...
-        "optionalPeerDependencies",
-        "types",
-        "unlisted",
-        "unresolved",
-    ],
+    include: allIssueTypes,
 
     ignore: [],
-
+    ignoreBinaries: [],
```
dsargent, line 64: Add this so "inheritors" can safely spread it, just like the other properties like `ignoreDependencies`.

.gitignore
```diff
@@ -23,7 +23,7 @@ coverage/
 **/dist/
 package.tgz
 yarn-error.log
-.yarn
+.yarn/install-state.gz
```
dsargent, line 26: Make this more specific: the new yarn plugin lives under `.yarn` which is where Copilot suggested. I don't know if that's normal and idiomatic.

.nvmrc
dsargent, line 1: Bump this

.yarn/plugins/CheckNodeVersion.mjs
```diff
@@ -0,0 +1,33 @@
...
+ */
+import pkg from "../../package.json" with {type: "json"}
+
+const nodeMajorActual = Number(process.versions.node.split(".", 1)[0])
+
+const nodeMajorsRequired = [
+    ...new Set(
+        pkg.engines.node
+            .split("||")
+            .map((range) => range.trim().match(/^\^?(?<temp1>\d+)\./u)?.[1])
```
dsargent: Allows for multiple versions, for a smooth migration. We can simplify this after we migrate.

There are of course semver parsing libraries, but that seemed _way_ overkill for this. Hopefuly yarn will fix their bug some day and this can all go away.

.yarnrc.yml
```diff
@@ -1,5 +1,9 @@
 nodeLinker: node-modules
 
+plugins:
```
dsargent, line 3: Invoke the new plugin. This will happen during _all_ yarn phases, so if you have the wrong Node.js version, you won't be able to `install`, `run`, or anything else.

.yarnrc.yml
```diff
@@ -14,7 +18,7 @@ catalogs:
         "@testing-library/react": "16.3.0"
         "@testing-library/user-event": "14.6.1"
         "@types/lodash-es": "4.17.12"
-        "@types/node": "22.17.1"
+        "@types/node": "26.1.1"
```
dsargent, line 21: Bump types. This also seems to work fine with v22 (backward compatibility).

knip.config.ts
```diff
@@ -26,6 +26,13 @@ import {config as baseConfig} from "./packages/dev-common/Configs/knip.config"
 const config: KnipConfig = {
     ...baseConfig,
 
+    ignoreBinaries: [
+        ...baseConfig.ignoreBinaries,
+
+        // Used in package.json script
+        "mktemp",
```
dsargent, line 33: We now have to use `mktemp` to give Node a storage dir to simulate `localStorage`. `knip` does have a whitelist of "OS binaries that can be assumed to be present" but for some reason `mktemp` isn't on the list, so we have to explicitly exclude it.

package.json
```diff
@@ -22,10 +22,13 @@
         "lint-rules": "./build_scripts/eslint_list_rules.sh",
         "lint-summary": "eslint --format summary",
         "start": "yarn workspace @cognizant-ai-lab/neuro-san-web start",
-        "test": "vitest run",
+        "test": "NODE_OPTIONS=\"--localstorage-file=$(mktemp --directory)/node-storage\" vitest run",
```
dsargent, line 25: Work-around for Node.js opinionated design decision in v25+.

Ignored in v22, but will be used in v26 for running tests.

Not sure if we'll be stuck with this forever or not; the Node dev team has already gone back and forth on this.

References:

https://appwrite.io/blog/post/nodejs-v25-whats-new
https://github.com/nodejs/node/issues/60303

package.json
```diff
@@ -22,10 +22,13 @@
...
         "update-coverage": "./build_scripts/UpdateCoverage.sh"
     },
     "packageManager": "yarn@4.12.0",
+    "engines": {
```
dsargent, line 29: This is where one is _supposed_ to specify the Node.js version, but Yarn ignores it due to the bug referenced elsewhere.

Here I specify both versions 22 and 26 for a smooth migration, 22 being the one we're on, and 26 being the one we want to be on.

package.json
```diff
@@ -49,7 +52,7 @@
         "eslint-plugin-unicorn": "65.0.1",
         "fake-indexeddb": "catalog:default",
         "globals": "catalog:default",
-        "jsdom": "^29.1.1",
+        "jsdom": "29.1.1",
```
dsargent, line 55: Shouldn't be floating

swensel (dismissed):
LGTM. Nice job.

package.json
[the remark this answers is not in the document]
dsargent: Worked fine on my macOS.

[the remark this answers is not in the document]
swensel: I have not found that to be true on Mac OS 26.5.2:

```
yarn test

 RUN  v4.1.10 /Users/971999/Documents/git_repo/neuro-san-ui


 Test Files  65 passed (65)
      Tests  793 passed (793)
   Start at  15:21:18
   Duration  23.86s (transform 3.83s, setup 13.40s, import 58.85s, tests 27.92s, environment 81.24s)
```

apps/main/Dockerfile
```diff
@@ -20,10 +20,10 @@
 # It uses multi-stage builds for a smaller resulting image 
 #
 
-# This is the major version of NodeJS we will enforce
-# Check in CI to see what the variable is set to if you want to see the official version of NodeJS we are targeting
-ARG NODEJS_VERSION=22
-
+# This is the major version of NodeJS we will use
```
dsargent, line 23: @donn-leaf I removed the default here and clarified the doc.

Now, if someone tries to build this without specifying the `NODEJS_VERSION` build arg, it will fail. Unfortunately it fails in an ugly way, as the first thing it tries to do is `FROM <node image based on that version>` which fails if the version is absent or invalid.

Could we make it fail in a nicer way? Not really, due to a chicken-and-egg problem. If we wanted to do something like `RUN test -n "$NODEJS_VERSION"`...that already requires us to run some base image. And a base image is exactly what we don't have yet, because we need to know the Node.js version. We _could_ do a cheeky `FROM alpine` I suppose, _just_ to check that build arg, but that's crazy talk. Way overkill IMO.

Unfortunately we can't just pass different values for `NODEJS_VERSION` because they won't work with the distroless hash in the `builder` stage. Not only that, but Google stopped releasing Debian12 distroless images as of Node v26 so we'll have to upgrade the Debian version to ≥ 13 as well as the hash tag when you migrate this to node v26.

apps/main/Dockerfile
dsargent: @donn-leaf 👀 this is the part that will fail if you try to feed it a new `NODEJS_VERSION`. If you look [here](https://console.cloud.google.com/artifacts/docker/distroless/us/gcr.io?pli=1) you can see that as of Node v26, no more Debian 12 (Bookworm) support and we'll have to move to Debian 13 (Trixie). And of course update the `sha256` hash to whatever is right for that image.

apps/main/Dockerfile
```diff
@@ -33,6 +33,7 @@ ENV NEXT_TELEMETRY_DISABLED=1
 
 WORKDIR /app
 COPY .yarnrc.yml ./
+COPY .yarn/plugins/CheckNodeVersion.mjs .yarn/plugins/CheckNodeVersion.mjs
```
dsargent, line 36: Have to copy this plugin as `yarn` now depends on it to check the Node.js version.

apps/main/Dockerfile
```diff
@@ -59,6 +60,7 @@ WORKDIR /app
 
 # Install only production dependencies
 COPY .yarnrc.yml ./
+COPY .yarn/plugins/CheckNodeVersion.mjs .yarn/plugins/CheckNodeVersion.mjs
```
dsargent, line 63: And here too as it's a separate build stage.

swensel (dismissed):

build_scripts/UpdateCoverage.sh
```diff
@@ -95,7 +95,7 @@ main() {
     new_coverage_values="${output_dir}/new_coverage_values.txt"
 
     # Run the tests with coverage
-    if ! yarn vitest run --coverage \
+    if ! yarn test --coverage \
```
dsargent, line 98: Forgot about this one. I modified it to use `yarn test` as an opaque script, so we don't have to worry about Node.js options here: they will be inherited from whatever is in `package.json`.

vitest.config.ts
```diff
@@ -25,7 +25,7 @@ export default defineConfig({
                 statements: -104,
                 branches: -155,
                 functions: -25,
-                lines: -76,
+                lines: -75,
```
dsargent, line 28: Not sure how this happened but I'll take it.

apps/main/Dockerfile
dsargent: Turns out to be not possible/advisable to have "no default". See build check rule: https://docs.docker.com/reference/build-checks/invalid-default-arg-in-from/

Instead, I'm making it default to `0` which is "always invalid" so will fail the build if the Node.js version is not specified. Which is not really in the spirit of the build check rule linked above, since you're supposed to be able to get _something_ from a `docker build` without specifying _any_ args but...I feel it's how it should be.

.github/workflows/test.yml
```diff
@@ -117,7 +117,8 @@ jobs:
               id: dockerbuild
               if: success() || failure()
               run: |
-                  docker build --quiet --check --file apps/main/Dockerfile . > /dev/null
+                  docker --version
```
dsargent, line 120: Added the `--version` output here for troubleshooting.

Also, have to pass the `NODEJS_VERSION` here now due to forcing the invalid value in the `Dockerfile`.

.yarn/plugins/CheckNodeVersion.mjs
dsargent: Made this a "proper" yarn plugin. Copilot suggested the original one, then later claimed it wasn't working because "it's not a proper yarn plugin" 🤦🏻 Now it follows the standard for plugins.

.yarn/plugins/CheckNodeVersion.mjs
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * Note: the parsing is intentionally simple-minded. It will only handle versions like `^a.b.c` or `^x.y.z || ^a.b.c`.
+ */
+
+// eslint-disable-next-line unicorn/no-exports-in-scripts, no-shadow -- required for yarn plugin
+export const name = "enforce-node-version"
+
+// eslint-disable-next-line unicorn/no-exports-in-scripts -- required for yarn plugin
+export const factory = (
+    /** @type {(name: string) => {MessageName: {UNNAMED: string}}} */
```
dsargent, line 23: We have to put these type annotations because our `tsconfig` (intentionally) includes JS and MJS files under the purview of tsc, as it does useful checks. But that means it complains about things like "explicit `any`" so I had to add type annotations.

swensel (dismissed):

.github/workflows/test.yml
```diff
@@ -66,7 +66,7 @@ jobs:
               # v1.0.7
               uses: cognizant-ai-lab/build-common/actions/setup-node-env@3ead7f681f92044709ffc3a533f41c3f48230cfd
               with:
-                  node-version: "22.17.1"
+                  node-version: ${{ env.NODEJS_VERSION }}
```
donn-leaf: Here we remove the hardcode and we'll rely on the repo level variable, which I will move to 26

donn-leaf (approved):
LGTM. I've looked at what I'll need to do deployment wise. We'll also have to update test.yml to remove the hardcoded reference to 22.17.1 at that time.

Issue #481, opened 2026-07-16
Editing UI gets stuck in "canceling" state


dsargent:
Steps to reproduce:

1. Create a temporary network
2. Click "Edit" to edit the entire network
3. Enter an edit request
4. Click "abort" while it's working on the request, but then don't do anything
5. Eventually the edit will complete and you get the "success" banner
6. Now try to enter a new edit prompt: you get the "cancel/continue" confirmation immediately.

See demo video.

https://github.com/user-attachments/assets/0beb7ec5-d99c-4cb2-9222-6dbd0f7cb670

PR #483, opened 2026-07-18
#481 - Editing UI gets stuck in "canceling" state.

swensel:
# Description

Editing UI gets stuck in "canceling" state.

Issue: https://github.com/cognizant-ai-lab/neuro-san-ui/issues/481

# Video with the fix:

https://github.com/user-attachments/assets/b0e5d2c7-c6f7-434a-a803-6bb77e1bf6ee


packages/ui-common/components/MultiAgentAccelerator/AgentFlow/AgentFlow.tsx
```diff
@@ -665,6 +665,7 @@ export const AgentFlow: FC<AgentFlowProps> = ({
             clearTimeout(timeoutId)
             dockAbortControllerRef.current = null
             setIsDockStreaming(false)
+            setStopState(null)
```
swensel: This is the fix for the issue. It resets the "stop state" to `null` so that it's not stuck in the "canceling" confirmation state on subsequent requests.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/AgentFlow.test.tsx
```diff
@@ -2420,6 +2420,48 @@ describe("AgentFlow", () => {
             })
         })
 
+        it("clears the confirm card if the request completes while it is open, so the next apply streams", async () => {
```
swensel, line 2431: Cover the test case for this bug fix.

dsargent: I always like to see a test added when a bug is fixed so 👍🏻

packages/ui-common/components/MultiAgentAccelerator/AgentFlow/AgentFlow.tsx
dsargent: 1) I'm wondering if this also needs to be set in other places where we `setIsDockStreaming(false)`? Corollary, when I see two variables that are always updated in lockstep, I start to think it's really _one_ variable masquerading as two.

2) In any case, `stopState` smells like a boolean: it can only have values `confirming` or `null`. Could it be simplified?

packages/ui-common/components/MultiAgentAccelerator/AgentFlow/AgentFlow.tsx
```diff
@@ -474,7 +474,7 @@ export const AgentFlow: FC<AgentFlowProps> = ({
     const dockAbortControllerRef = useRef<AbortController | null>(null)
 
     // Stop-confirm overlay state: null = not shown, "confirming" = abort dialog open.
-    const [stopState, setStopState] = useState<"confirming" | null>(null)
+    const [stopState, setStopState] = useState(false)
```
swensel: Updated `stopState` to a boolean.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow/AgentFlow.tsx
```diff
@@ -665,6 +665,7 @@ export const AgentFlow: FC<AgentFlowProps> = ({
             clearTimeout(timeoutId)
             dockAbortControllerRef.current = null
             setIsDockStreaming(false)
```
swensel: @dsargent There was a comment that you had left (that was eaten by GitHub?) about having multiple state variables here, and that's a fair point. I'll keep it in mind for a future PR. In the meantime, I updated `setStopState` to a boolean.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow/AgentFlow.tsx
```diff
@@ -1302,7 +1303,7 @@ export const AgentFlow: FC<AgentFlowProps> = ({
                 open={isDockStreaming}
                 sx={{zIndex: (t) => t.zIndex.modal + 1}}
             >
-                {stopState === "confirming" ? (
+                {stopState ? (
```
swensel: Cleaner now that `stopState` is a boolean.

swensel:
Build looks weird but I think that's due to https://github.com/cognizant-ai-lab/neuro-san-ui/pull/499 being merged. The checks that did run were successful.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow/AgentFlow.tsx
dsargent: But, now it's misnamed. `stopState` doesn't sound or feel like a boolean. `isConfirming` or something boolean-ish would be better.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow/AgentFlow.tsx
dsargent: Yes, somehow it disappeared. I was asking if it also needs to be reset in other places where `setIsDockStreaming` is called, and if they are always called in lockstep, why are there two variables?

packages/ui-common/components/MultiAgentAccelerator/AgentFlow/AgentFlow.tsx
```diff
@@ -501,12 +501,11 @@ export const AgentFlow: FC<AgentFlowProps> = ({
 
     // Dock (edit-mode prompt bar) state
     const [dockPrompt, setDockPrompt] = useState<string>("")
-    const [isDockStreaming, setIsDockStreaming] = useState<boolean>(false)
+    // "idle" = no apply in flight, "applying" = applying save, "confirmingStop" = abort confirmation dialog is open
+    const [dockPhase, setDockPhase] = useState<"idle" | "applying" | "confirmingStop">("idle")
```
swensel, line 505: Combined the state variables into one - `dockPhase`.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow/AgentFlow.tsx
```diff
@@ -501,12 +501,11 @@ export const AgentFlow: FC<AgentFlowProps> = ({
...
+    const isDockPending = dockPhase !== "idle"
```
swensel, line 506: Checked in several places, so make it a `const`.

packages/ui-common/components/MultiAgentAccelerator/AgentFlow/AgentFlow.tsx
dsargent: Much cleaner and easier to understand. I _knew_ it was one variable masquerading as two.

Now, I don't even see the fix any more, which is probably a good sign: variables can't get out of sync when there's only one of them?

dsargent (approved):
Looks good. Thanks for the clean-up.

PR #486, opened 2026-07-21
Surface API key test error for BYOK in Settings.

swensel:
# Description

Add component to surface API key test error for BYOK in Settings Dialog.

Also make an update to the test script since I was having issues running it on my machine. Please see more info about that in this [comment](https://github.com/cognizant-ai-lab/neuro-san-ui/pull/486#discussion_r3643057013).

# Screenshots

Screenshot 2026-07-28 at 4 12 01 PM
Screenshot 2026-07-28 at 4 12 08 PM
Screenshot 2026-07-28 at 4 11 18 PM
Screenshot 2026-07-28 at 4 10 03 PM
Screenshot 2026-07-28 at 4 09 36 PM
Screenshot 2026-07-28 at 4 09 19 PM


packages/ui-common/controller/llm/Providers.ts
```diff
@@ -1,7 +1,53 @@
...
+
+const toFailureResult = async (res: Response): Promise<KeyValidationResult> => {
+    let body
     try {
-        // Just call the "list models" API with the supplied key.
-        // We don't care about the result, just whether it succeeds or not
+        body = await res.json()
+    } catch {
+        body = undefined
+    }
```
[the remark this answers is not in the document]
swensel: I refactored out `toFailureResult`.

packages/ui-common/components/Settings/ApiKeyInput.tsx
```diff
@@ -65,11 +71,15 @@ export const ApiKeyInput: FC<ApiKeyInputProps> = ({
     useEffect(() => {
         setInputValue(persistedValue ?? "")
         setKeyValidated(null)
+        onResultChange?.(null)
+        // onResultChange is a stable callback from the parent; re-running only on persistedValue is intended
+        // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [persistedValue])
```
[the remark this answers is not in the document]
swensel: Updating.

packages/ui-common/components/Settings/ApiKeyErrorBanner.tsx
```diff
@@ -0,0 +1,102 @@
...
+interface ApiKeyErrorBannerProps {
+    readonly id: string
+    readonly failures: readonly ApiKeyFailure[]
+}
+
+/**
+ * Map an HTTP status code to a short, human-readable summary for the banner header.
+ * @param status The HTTP status code, if known.
+ * @return A short description of the failure.
+ */
```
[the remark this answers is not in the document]
swensel: Updating this.

packages/ui-common/__tests__/components/Settings/ApiKeyErrorBanner.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+const renderBanner = (failures: readonly ApiKeyFailure[]) =>
+    render(
+        <ApiKeyErrorBanner
+            failures={failures}
+            id="banner"
+        />
+    )
+
+describe("ApiKeyErrorBanner", () => {
```
swensel, line 48: New tests for `ApiKeyErrorBanner`.

packages/ui-common/__tests__/components/Settings/ApiKeyErrorBanner.test.tsx
```diff
@@ -0,0 +1,99 @@
+import {render, screen, within} from "@testing-library/react"
+
+import {withStrictMocks} from "../../../../../__tests__/common/strictMocks"
+import {ApiKeyErrorBanner, ApiKeyFailure} from "../../../components/Settings/ApiKeyErrorBanner"
+
+// Base result used for both providers
+const BASE_401_RESULT = {ok: false, status: 401}
```
swensel: `consts` used for the new tests.

packages/ui-common/__tests__/components/Settings/ApiKeyErrorBanner.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+    })
+
+    it("reports a network/CORS error as a status-less request failure", () => {
+        renderBanner([OPENAI_NETWORK_FAILURE])
+
+        expect(screen.getByText("OpenAI — Request failed")).toBeInTheDocument()
+        expect(screen.getByText(OPENAI_NETWORK_ERROR_MESSAGE)).toBeInTheDocument()
+    })
+
+    it.each([
```
swensel: Exercise different statuses.

packages/ui-common/__tests__/components/Settings/SettingsDialog.test.tsx
```diff
@@ -351,6 +351,40 @@ describe("SettingsDialog", () => {
             expect(statusLight).toHaveAttribute("data-status", "red")
         })
 
+        it("surfaces a failed key test in the aggregated error banner", async () => {
```
swensel, line 355: New test for more coverage.

packages/ui-common/__tests__/unit/Controller/llm/Providers.test.ts
```diff
@@ -21,28 +21,61 @@ import {isAnthropicKeyValid, isOpenAIKeyValid} from "../../../../controller/llm/
 describe("Providers controller", () => {
     withStrictMocks()
 
-    it("should validate keys", async () => {
-        // Happy response
+    it("should return ok for a successful response", async () => {
```
swensel, line 44: Update tests based on `Providers` component changes.

package.json
```diff
@@ -22,7 +22,7 @@
         "lint-rules": "./build_scripts/eslint_list_rules.sh",
         "lint-summary": "eslint --format summary",
         "start": "yarn workspace @cognizant-ai-lab/neuro-san-web start",
-        "test": "NODE_OPTIONS=\"--localstorage-file=$(mktemp --directory)/node-storage\" vitest run",
+        "test": "NODE_OPTIONS=\"--no-experimental-webstorage\" vitest run",
```
swensel, line 25: Switched the `test` script from `--localstorage-file=$(mktemp --directory)/node-storage` to `--no-experimental-webstorage`.

I ran into an issue on my machine where tests were failing intermittently. The issue was that `$(mktemp --directory)` ran once when the script started, so every vitest worker ended up pointing at the same localStorage file. That let state leak between test files running across workers, which is what produced the failures. Turning off Node's experimental webstorage lets jsdom give each test environment its own in-memory storage instead, so parallel worker runs stay clean and we don't need a temp file.

A few things worth noting about the change:

- The flag disables `sessionStorage` too, not just `localStorage`, since it turns off Node's whole Web Storage API. That's probably the behavior we want here: `Settings` stores API keys in `sessionStorage`, so now both stores come from jsdom and both get isolated per test environment.
- It also removes an ambiguity. Before, both Node (file-backed) and jsdom (in-memory) were defining a `localStorage`, so which one a test hit depended on setup order. Now jsdom's is the only one, so it's deterministic.

The change only affects the `test` script. Net effect is better isolation for parallel runs with no change to how the tests themselves behave.

knip.config.ts
```diff
@@ -26,13 +26,6 @@ import {config as baseConfig} from "./packages/dev-common/Configs/knip.config"
 const config: KnipConfig = {
     ...baseConfig,
 
-    ignoreBinaries: [
```
swensel: Dropped `mktemp` from `ignoreBinaries` in `knip.config.ts`, since it was only there for the test script and knip now flags it as unused.

package.json
swensel: Note: This was unrelated to this PR.

packages/ui-common/components/Settings/SettingsDialog.tsx
```diff
@@ -107,7 +110,7 @@ interface LLMProviderInputConfig {
     vendor: LLMProvider
     idSuffix: string
     logo: string
-    onTest: (key: string) => Promise<boolean>
+    onTest: (key: string) => Promise<KeyValidationResult>
```
swensel, line 119: Use new `KeyValidationResult` interface.

packages/ui-common/components/Settings/ApiKeyErrorBanner.tsx
```diff
@@ -0,0 +1,102 @@
+import ErrorOutlineIcon from "@mui/icons-material/ErrorOutlined"
+import Box from "@mui/material/Box"
+import {alpha} from "@mui/material/styles"
+import Typography from "@mui/material/Typography"
+import {FC} from "react"
+
+import {KeyValidationResult} from "../../controller/llm/Providers"
+
+interface ApiKeyErrorBannerProps {
```
swensel, line 10: New `interfaces`.

packages/ui-common/components/Settings/ApiKeyErrorBanner.tsx
```diff
@@ -0,0 +1,102 @@
...
+    readonly vendor: string
+    readonly result: KeyValidationResult
+}
+
+/**
+ * Map an HTTP status code to a short, human-readable summary for the banner header.
+ * @param httpStatus The HTTP status code, if known.
+ * @return A short description of the failure.
+ */
+const describeStatus = (httpStatus?: number): string => {
```
swensel: Human readable messages for different statuses.

packages/ui-common/controller/llm/Providers.ts
```diff
@@ -1,42 +1,71 @@
...
-    } catch (e) {
-        console.error("Error validating API key:", e)
-        return false
-    }
+/**
+ * Result of testing an API key. On failure, {@link status} and {@link message} carry whatever
+ * detail could be recovered. (A discriminated union would model this better, but this project
+ * runs with `strictNullChecks: false`, under which the `ok` discriminant does not narrow.)
+ */
+export interface KeyValidationResult {
```
swensel: New interfaces.

packages/ui-common/controller/llm/Providers.ts
```diff
@@ -1,42 +1,71 @@
...
-        return res.ok
+/**
+ * Validate an API key by calling a provider's authenticated "list models" endpoint. The status is
+ * all we need: a 2xx means the key works; any other response is a failure whose status and message
+ * we surface. A thrown error (network failure, CORS rejection) is likewise reported as a failure.
+ * @param url The provider's "list models" endpoint.
+ * @param headers Request headers, including authentication for the supplied key.
+ * @return A {@link KeyValidationResult} describing the outcome.
+ */
+const validateKey = async (url: string, headers: HeadersInit): Promise<KeyValidationResult> => {
```
swensel, line 53: New `validateKey` to reduce copy and pasted code.

package.json
swensel: This [commit](https://github.com/cognizant-ai-lab/neuro-san-ui/commit/6a8cd3c6e4150ea7a7b097f50270d7167578c3df) to `main` also failed. Seems like it could be `darkMode` stored in zustand/localStorage, and also related.

Screenshot 2026-07-24 at 4 08 24 PM

swensel: I'll do this in https://github.com/cognizant-ai-lab/neuro-san-ui/pull/489 instead. It deserves its own PR.

knip.config.ts
swensel: I'll do this in https://github.com/cognizant-ai-lab/neuro-san-ui/pull/489 instead. It deserves its own PR.

packages/ui-common/__tests__/unit/Controller/llm/Providers.test.ts
```diff
@@ -18,31 +18,66 @@ import {withStrictMocks} from "../../../../../../__tests__/common/strictMocks"
 import {mockFetch} from "../../../../../../__tests__/common/TestUtils"
 import {isAnthropicKeyValid, isOpenAIKeyValid} from "../../../../controller/llm/Providers"
 
+const FAKE_KEY = "fake-key"
+const OK_FALSE = {ok: false}
+
+const mockFailedFetch = (httpStatus: number, json: () => Promise<unknown>) =>
+    vi.fn().mockResolvedValue({ok: false, status: httpStatus, json})
+
+const spyOnConsoleError = () => vi.spyOn(console, "error").mockImplementation(vi.fn())
```
swensel, line 39: Reduce some copy and pasting.

packages/ui-common/components/Settings/ApiKeyErrorBanner.tsx
```diff
@@ -0,0 +1,103 @@
...
+        case 401:
+            return "Authentication failed"
+        case 403:
+            return "Access forbidden"
+        case 404:
+            return "Not found"
+        case 429:
+            return "Rate limited"
+        default:
+            return httpStatus !== undefined && httpStatus >= 500 ? "Server error" : "Request failed"
```
swensel: I'm not sure if this branch really makes sense. Both could be 500 or a general error.

packages/ui-common/components/Settings/ApiKeyErrorBanner.tsx
```diff
@@ -0,0 +1,103 @@
...
+    }
+}
+
+const FailureRow: FC<{id: string; failure: ApiKeyFailure}> = ({id, failure}) => {
+    const {vendor, result} = failure
+    const {status: statusCode, message} = result
+
+    const title = `${vendor} — ${describeStatus(statusCode)}${statusCode !== undefined ? ` (${statusCode})` : ""}`
```
swensel: Title of failure.

packages/ui-common/components/Settings/ApiKeyErrorBanner.tsx
```diff
@@ -0,0 +1,103 @@
...
+    return (
+        <Box data-testid={`${id}-failure`}>
+            <Typography
+                sx={{color: "error.main", fontWeight: 700}}
+                variant="body2"
+            >
+                {title}
+            </Typography>
+            {/* The error messages from the provider, if present in the response body. */}
+            {message ? (
```
swensel, line 42: If we have a message from the provider.

packages/ui-common/components/Settings/ApiKeyErrorBanner.tsx
```diff
@@ -0,0 +1,103 @@
...
+            ) : null}
+        </Box>
+    )
+}
+
+/**
+ * A banner shown at the top of the API Keys section that aggregates all failing key-test results.
+ */
+// eslint-disable-next-line react/no-multi-comp -- co-located with its single-row helper
+export const ApiKeyErrorBanner: FC<ApiKeyErrorBannerProps> = ({id, failures}) => {
```
swensel, line 58: Multiple small components in this file. Could also be their own small components. Not sure. We are doing this in a few places in the codebase already.

packages/ui-common/components/Settings/ApiKeyErrorBanner.tsx
```diff
@@ -0,0 +1,103 @@
...
+    if (failures.length === 0) {
+        return null
```
swensel, line 60: Quick return if no failures.

packages/ui-common/components/Settings/ApiKeyErrorBanner.tsx
```diff
@@ -0,0 +1,103 @@
...
+import {KeyValidationResult} from "../../controller/llm/Providers"
+
+interface ApiKeyErrorBannerProps {
+    readonly id: string
+    readonly failures: readonly ApiKeyFailure[]
+}
+
+export interface ApiKeyFailure {
+    readonly vendor: string
+    readonly result: KeyValidationResult
```
swensel: Use new `KeyValidationResult` interface.

packages/ui-common/components/Settings/ApiKeyInput.tsx
```diff
@@ -10,15 +10,17 @@ import TextField from "@mui/material/TextField"
...
 import {StatusLight} from "../Common/StatusLight"
 
 interface ApiKeyInputProps {
     readonly forgetKey: () => void
     readonly id: string
     readonly logo: string
+    readonly onResultChange?: (result: KeyValidationResult | null) => void
     readonly onSave: (key: string) => void
-    readonly onTest: (key: string) => Promise<boolean>
+    readonly onTest: (key: string) => Promise<KeyValidationResult>
```
swensel, line 24: Use new `KeyValidationResult` interface.

packages/ui-common/components/Settings/SettingsDialog.tsx
```diff
@@ -195,6 +198,9 @@ export const SettingsDialog: FC<SettingsDialogProps> = ({id, isOpen, logoService
     const openAIKeyCheckmark = useCheckmarkFade()
     const anthropicKeyCheckmark = useCheckmarkFade()
 
+    // Failing key-test results, aggregated into the banner at the top of the API Keys section
+    const [keyTestResults, setKeyTestResults] = useState<Partial<Record<LLMProvider, KeyValidationResult | null>>>({})
```
swensel: Little bit weird TypeScript syntax, but, valid.

packages/ui-common/controller/llm/Providers.ts
```diff
@@ -1,42 +1,77 @@
-export const isOpenAIKeyValid = async (key: string) => {
+//#region Constants
+
+const ACCEPT_APPLICATION_JSON = "application/json"
```
swensel: Might be overkill, but reducing copy and paste.

packages/ui-common/controller/llm/Providers.ts
```diff
@@ -1,42 +1,77 @@
...
+        }
+        return {ok: false, status: res.status, message: await readErrorMessage(res)}
     } catch (e) {
         console.error("Error validating API key:", e)
-        return false
+        return {ok: false, message: e instanceof Error ? e.message : String(e)}
     }
 }
+
+export const isOpenAIKeyValid = (key: string): Promise<KeyValidationResult> =>
```
swensel, line 66: Use new `validateKey` for both of these.

packages/ui-common/controller/llm/Providers.ts
```diff
@@ -1,42 +1,77 @@
...
+
+//#endregion: Types and Interfaces
+
+/**
+ * Read the provider's error message from a failed response, tolerating empty or non-JSON
+ * bodies (which cause {@link Response.json} to reject).
+ * @param res The failed response.
+ * @return The provider's error message, or `undefined` if none could be parsed.
+ */
+const readErrorMessage = async (res: Response): Promise<string | undefined> => {
```
swensel: Could also be called `readErrorMessageFromProvider`. That's more what this is doing, rather than a general error message. That is noted in the function comment, currently.

packages/ui-common/__tests__/components/Settings/ApiKeyErrorBanner.test.tsx
dsargent: Use `httpStatus` values here. Example: `apps/main/__tests__/pages/api/handlers/Handlers.test.ts`.

packages/ui-common/controller/llm/Providers.ts
dsargent: I was amazed to discover the MIME types aren't already defined somewhere in the JS/TS ecosystem. 

But if we're going to define these, they should be in some central place, in a separate module -- `MimeTypes.ts` or some such. And also consider a separate pass/PR to update the code to use the `const`, rather than, "this is what I need for this PR, so I'm all good and don't worry about the rest".

packages/ui-common/controller/llm/Providers.ts
```diff
@@ -1,42 +1,77 @@
...
+
+const ACCEPT_APPLICATION_JSON = "application/json"
+
+//#endregion Constants
+
+//#region: Types and Interfaces
+
+/**
+ * Result of testing an API key. On failure, {@link status} and {@link message} carry whatever
+ * detail could be recovered. (A discriminated union would model this better, but this project
```
dsargent: This looks like an LLM comment. `strictNullChecks: false` isn't a holy setting that will never change; it's something we want to enable but can't currently. The bot treats it like gospel, some kind of fiat-from-heaven, but it's actually transitional.

Yes, this probably should be a discriminated union.

packages/ui-common/controller/llm/Providers.ts
dsargent: Noble sentiment.

packages/ui-common/controller/llm/Providers.ts
```diff
@@ -1,42 +1,77 @@
...
-        return res.ok
+        const res = await fetch(url, {method: "GET", headers})
+        if (res.ok) {
+            return {ok: true}
+        }
+        return {ok: false, status: res.status, message: await readErrorMessage(res)}
     } catch (e) {
         console.error("Error validating API key:", e)
-        return false
+        return {ok: false, message: e instanceof Error ? e.message : String(e)}
```
dsargent, line 62: We may want a centralized customized type guard to deal with the "errors are unknown type" setting, which is why we have the sniffing logic here. See [here](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards). That would allow us to auto narrow and not have this same sniffing logic each time we `catch`.

packages/ui-common/components/Settings/ApiKeyInput.tsx
```diff
@@ -65,11 +71,16 @@ export const ApiKeyInput: FC<ApiKeyInputProps> = ({
     useEffect(() => {
         setInputValue(persistedValue ?? "")
         setKeyValidated(null)
+        onResultChange?.(null)
+        // Only clear results when persistedValue changes; intentionally omit onResultChange so a new
```
dsargent: Disabling the rule is almost always bad (and typical of AI coding assistants: "me see red squiggles, me disable rule").

Better here: stabilize the callback with `useCallback` in the parent, and include it as a dep. Then we don't need to gag the linter.

packages/ui-common/components/Settings/ApiKeyErrorBanner.tsx
dsargent: Totally unnecessary; we already have a dep for this, `httpStatus`. Please search the codebase before reinventing the wheel. See examples of mapping HTTP status to text in `Sidebar`.

packages/ui-common/components/Settings/ApiKeyErrorBanner.tsx
```diff
@@ -0,0 +1,103 @@
...
+                </Typography>
+            ) : null}
+        </Box>
+    )
+}
+
+/**
+ * A banner shown at the top of the API Keys section that aggregates all failing key-test results.
+ */
+// eslint-disable-next-line react/no-multi-comp -- co-located with its single-row helper
```
dsargent: I'm 50-50 if we should just disable `no-multi-comp` and enforce sane behavior via code reviews.

packages/ui-common/components/Settings/ApiKeyErrorBanner.tsx
dsargent, line 1: Why are we not using `MUIAlert` here?! `packages/ui-common/components/Common/MUIAlert.tsx` This is reinventing the wheel: `MUIAlert` already has the styling, banner, error icon, everything.

packages/ui-common/components/Settings/ApiKeyErrorBanner.tsx
dsargent: I would trust `httpStatus` (the library we already depend on)

packages/ui-common/__tests__/unit/Controller/llm/Providers.test.ts
```diff
@@ -18,31 +18,66 @@ import {withStrictMocks} from "../../../../../../__tests__/common/strictMocks"
...
         global.fetch = mockFetch({})
 
-        expect(await isOpenAIKeyValid("fake-key")).toBe(true)
-        expect(await isAnthropicKeyValid("fake-key")).toBe(true)
+        const success = {ok: true}
+        expect(await isOpenAIKeyValid(FAKE_KEY)).toEqual(success)
+        expect(await isAnthropicKeyValid(FAKE_KEY)).toEqual(success)
+    })
+
+    it.each([
```
dsargent: Good use case for "generic-shaped `it.each`" here. `it.each<SomeInterface>...`

packages/ui-common/__tests__/unit/Controller/llm/Providers.test.ts
```diff
@@ -18,31 +18,66 @@ import {withStrictMocks} from "../../../../../../__tests__/common/strictMocks"
...
+        [
+            "Anthropic",
+            isAnthropicKeyValid,
+            {type: "error", error: {type: "authentication_error", message: "invalid x-api-key"}},
+            "invalid x-api-key",
+        ],
+    ] as const)(
+        "should surface the status and error message from a failed %s response",
+        async (_name, validate, body, expectedMessage) => {
+            global.fetch = mockFailedFetch(401, () => Promise.resolve(body))
```
dsargent: I think our trusty `mockFetch` already has a "pretend to fail" flag.

packages/ui-common/__tests__/components/Settings/ApiKeyErrorBanner.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+    it("renders nothing when there are no failures", () => {
+        const {container} = renderBanner([])
+        expect(container).toBeEmptyDOMElement()
+    })
+
+    it("surfaces an Anthropic 401 invalid key message", () => {
+        renderBanner([ANTHROPIC_INVALID_KEY_FAILURE])
+
+        expect(screen.getByText(ANTHROPIC_AUTH_FAILED_TITLE)).toBeInTheDocument()
```
dsargent: `getByText()` already includes `toBeInTheDocument()` (will throw if not found)

packages/ui-common/__tests__/components/Settings/SettingsDialog.test.tsx
```diff
@@ -351,6 +351,40 @@ describe("SettingsDialog", () => {
...
+            global.fetch = vi.fn().mockResolvedValue({
```
dsargent: `mockFetch`?

packages/ui-common/__tests__/unit/Controller/llm/Providers.test.ts
dsargent: In this case it seems a bit extreme (trading 1 line for 1 line) but I appreciate the sentiment.

packages/ui-common/__tests__/components/Settings/SettingsDialog.test.tsx
```diff
@@ -351,6 +351,40 @@ describe("SettingsDialog", () => {
...
+
+            const apiKeyInput = screen.getByTestId("settings-dialog-anthropic-input")
+            const inputBox = within(apiKeyInput).getByPlaceholderText("sk-ant-...")
+
+            await user.click(inputBox)
+            await user.paste(TEST_API_KEY)
+            await user.click(within(apiKeyInput).getByRole("button", {name: "Test"}))
+
+            const banner = await screen.findByTestId("settings-dialog-api-key-error-banner")
+            expect(within(banner).getByText("Anthropic — Authentication failed (401)")).toBeInTheDocument()
```
dsargent: Again, messages like `Authentication failed` should come from `httpStatus`, not hardcoded.

packages/ui-common/components/Settings/SettingsDialog.tsx
dsargent: Seems absolutely fine to me; readable, and expresses intent well.

But, could you not use `ApiKeyFailure` here? That's exactly what it's expressing, an object containing those.

packages/ui-common/components/Settings/ApiKeyErrorBanner.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+import {KeyValidationResult} from "../../controller/llm/Providers"
+
+interface ApiKeyErrorBannerProps {
+    readonly id: string
+    readonly failures: readonly ApiKeyFailure[]
+}
+
+export interface ApiKeyFailure {
+    readonly vendor: string
```
dsargent: Shouldn't `vendor` be of type `LLMProvider`?

dsargent (changes requested):
Please see comments. Would mostly like to see the "reinventing the wheel" and disabled linter rule stuff addressed.

packages/ui-common/components/Settings/ApiKeyErrorBanner.tsx
swensel: I'd be fine with that.

packages/ui-common/controller/llm/Providers.ts
swensel: I'll think more about this one, but not for this PR.

packages/ui-common/components/Settings/ApiKeyErrorBanner.tsx
dsargent: Yes, I think it's too strict. Often makes sense to have small "auxiliary" components in the same source file as the one that uses them. But within reason: I don't want to see us shoveling half a dozen components into each source file either. That can be policed in code reviews. Feel free to disable the rule here or in a future PR if you want.

packages/ui-common/components/Settings/ApiKeyErrorBanner.tsx
swensel: Moved to `MUIAlert`. Updated the screenshots in the PR description.

__tests__/common/TestUtils.ts
```diff
@@ -15,10 +15,11 @@ limitations under the License.
 */
 
 // Mock Fetch for NodeJs environment
-export const mockFetch = (data: Record<string, unknown>, ok: boolean = true) => {
+export const mockFetch = (data: Record<string, unknown>, ok: boolean = true, statusCode?: number) => {
     return vi.fn().mockImplementation(() =>
         Promise.resolve({
             ok,
+            status: statusCode,
```
swensel, line 22: Adding so that we can use status code with `mockFetch`.

apps/main/pages/_app.tsx
```diff
@@ -84,12 +84,10 @@ const NavbarWrapper = (props: Omit<NavbarProps, "userInfo">): ReactElement => {
  * Shim to conditionally wrap the app in a SessionProvider if authentication is enabled.
  * Allows us to avoid next-auth errors when authentication is disabled
  */
-// eslint-disable-next-line react/no-multi-comp -- used only within this module
```
swensel, line 87: Disabled `react/no-multi-comp` globally. There were three files that included these overrides.

packages/dev-common/Configs/eslint.config.mjs
```diff
@@ -236,7 +236,7 @@ export default defineConfig([
             "prefer-spread": "error",
             "preserve-caught-error": "warn",
             "react/jsx-curly-brace-presence": "error",
-            "react/no-multi-comp": ["error", {ignoreStateless: false}],
+            "react/no-multi-comp": "off",
```
swensel: Disable `react/no-multi-comp` globally.

packages/ui-common/__tests__/components/Settings/ApiKeyErrorBanner.test.tsx
swensel: Switch these statuses over to `httpStatus` from the `http-status` library.

packages/ui-common/__tests__/components/Settings/SettingsDialog.test.tsx
```diff
@@ -147,7 +148,7 @@ describe("SettingsDialog", () => {
             global.fetch = mockFetch(
                 {
                     success: false,
-                    httpStatus: 200,
+                    httpStatus: httpStatus.OK,
```
swensel, line 151: Switch statuses over to `httpStatus` from the `http-status` library. Applies for all instances.

packages/ui-common/components/Settings/SettingsDialog.tsx
```diff
@@ -120,7 +129,6 @@ type Protocol = "http" | "https"
 
 //#endregion: Types and Interfaces
 
-// eslint-disable-next-line react/no-multi-comp -- only relevant to this module.
```
swensel, line 123: More removal of these overrides.

packages/ui-common/components/Settings/SettingsDialog.tsx
```diff
@@ -195,6 +202,9 @@ export const SettingsDialog: FC<SettingsDialogProps> = ({id, isOpen, logoService
     const openAIKeyCheckmark = useCheckmarkFade()
     const anthropicKeyCheckmark = useCheckmarkFade()
 
+    // Failing key-test results, aggregated into the banner at the top of the API Keys section
+    const [keyTestResults, setKeyTestResults] = useState<Partial<Record<LLMProvider, ApiKeyFailure | null>>>({})
```
swensel, line 206: I switched this to use `ApiKeyFailure`. That was a suggestion from this [comment](https://github.com/cognizant-ai-lab/neuro-san-ui/pull/486#discussion_r3661164695).

__tests__/common/TestUtils.ts
dsargent: Good idea. Better to enhance the existing than reinvent the wheel.

packages/dev-common/Configs/eslint.config.mjs
dsargent: Agreed, but should be in the "permanently off" section starting at line 308, with an explanation as to why we disagree/want it disabled. I also noticed the comment in that section is a bit wrong -- it mentions "not yet" but it's really "these are rules we never want to enable because...".``

packages/ui-common/__tests__/components/Settings/ApiKeyErrorBanner.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+import {withStrictMocks} from "../../../../../__tests__/common/strictMocks"
+import {ApiKeyErrorBanner, ApiKeyFailure} from "../../../components/Settings/ApiKeyErrorBanner"
+import {KeyValidationFailure} from "../../../controller/llm/Providers"
+
+// Base result used for both providers
+const BASE_401_RESULT: KeyValidationFailure = {ok: false, status: httpStatus.UNAUTHORIZED}
+
+// The banner header renders "<vendor> — <httpStatus reason phrase> (<status>)".
+const bannerTitle = (vendor: string, statusCode: number) =>
```
dsargent: But now this (mostly) replicates the function on line 24 of `ApiKeyErrorBanner`. 

I would:

1) Get rid of this function
2) Make the function `title` in the module under test, exported
3) Call that from the test when you need to get the title.

packages/ui-common/__tests__/components/Settings/ApiKeyErrorBanner.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+const ANTHROPIC_INVALID_KEY_MESSAGE = "API key is invalid."
+const ANTHROPIC_INVALID_KEY_FAILURE: ApiKeyFailure = {
+    vendor: "Anthropic",
+    result: {
+        ok: false,
+        status: httpStatus.UNAUTHORIZED,
+        message: ANTHROPIC_INVALID_KEY_MESSAGE,
+    },
+}
+const ANTHROPIC_CORS_MESSAGE =
```
dsargent: Kinda odd to have the whole, exact message from the vendor as test data. To the casual reader it implies there's something magical about this exact text. I understand the urge to "test with the real message", but I would be inclined to use something "meta" that is very clearly a test message to show it's arbitrary -- 

```typescript
const ANTHROPIC_CORS_MESSAGE = "error message for testing"
```

packages/ui-common/__tests__/components/Settings/ApiKeyErrorBanner.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+        httpStatus.UNAUTHORIZED,
+        httpStatus.FORBIDDEN,
+        httpStatus.NOT_FOUND,
+        httpStatus.TOO_MANY_REQUESTS,
+        httpStatus.INTERNAL_SERVER_ERROR,
+    ])("maps status %i to its banner title", (statusCode) => {
+        renderBanner([{vendor: "OpenAI", result: {ok: false, status: statusCode}}])
+
+        const banner = screen.getByRole("alert")
+        expect(within(banner).getByText(bannerTitle("OpenAI", statusCode))).toBeInTheDocument()
```
dsargent: Here, you're really just testing "does my ad hoc copy of the title generation logic in this test suite, match the one in the module under test?" which is not particularly useful.

packages/ui-common/__tests__/components/Settings/SettingsDialog.test.tsx
dsargent: _Much_ better.

packages/ui-common/__tests__/unit/Controller/llm/Providers.test.ts
```diff
@@ -14,35 +14,101 @@ See the License for the specific language governing permissions and
...
-import {isAnthropicKeyValid, isOpenAIKeyValid} from "../../../../controller/llm/Providers"
+import {
+    isAnthropicKeyValid,
+    isKeyValidationFailure,
+    isOpenAIKeyValid,
+    KeyValidationResult,
+} from "../../../../controller/llm/Providers"
+
+interface FailedResponseCase {
+    vendor: string
```
dsargent: Vendor can be a tighter type, `LLMProvider`.

packages/ui-common/components/Settings/ApiKeyErrorBanner.tsx
dsargent: Assumes that `failures` is truthy. May be okay given limited usage.

packages/ui-common/components/Settings/ApiKeyInput.tsx
```diff
@@ -65,11 +72,13 @@ export const ApiKeyInput: FC<ApiKeyInputProps> = ({
     useEffect(() => {
         setInputValue(persistedValue ?? "")
         setKeyValidated(null)
-    }, [persistedValue])
+        onResultChange?.(vendor, null)
+    }, [persistedValue, vendor, onResultChange])
```
dsargent, line 76: Better, respecting the deps rule.

packages/ui-common/controller/llm/Providers.ts
dsargent: or `extractErrorMessage`. "read" is a bit vague.

dsargent (dismissed):
A few style comments. No blockers.

packages/dev-common/Configs/eslint.config.mjs
```diff
@@ -305,7 +304,11 @@ export default defineConfig([
                 },
             ],
 
-            // Rules we're not ready to enable yet
+            // Rules we never want to enable because we disagree with them for this codebase
```
swensel, line 307: @dsargent Updated this comment and moved `"react/no-multi-comp": "off",` to this section.

packages/dev-common/Configs/eslint.config.mjs
```diff
@@ -538,7 +541,6 @@ export default defineConfig([
             "@next/next/no-img-element": "off",
             "react/display-name": "off",
             "react/no-array-index-key": "off",
-            "react/no-multi-comp": "off",
```
swensel, line 541: Remove extra `react/no-multi-comp` for tests since the disabled rule above handles this now.

packages/dev-common/Configs/eslint.config.mjs
```diff
@@ -236,7 +236,6 @@ export default defineConfig([
             "prefer-spread": "error",
             "preserve-caught-error": "warn",
             "react/jsx-curly-brace-presence": "error",
-            "react/no-multi-comp": ["error", {ignoreStateless: false}],
```
swensel, line 239: Moved this rule below.

packages/ui-common/__tests__/components/Settings/ApiKeyErrorBanner.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+        renderBanner([ANTHROPIC_INVALID_KEY_FAILURE, OPENAI_INVALID_KEY_FAILURE])
+
+        screen.getByText(ANTHROPIC_AUTH_FAILED_TITLE)
+        screen.getByText(OPENAI_AUTH_FAILED_TITLE)
+    })
+})
+
+describe("errorTitle", () => {
+    // Pins the branch the render tests can't reach: a status with no http-status phrase.
+    it("falls back to 'Unknown status' when the status has no reason phrase", () => {
```
swensel, line 101: This is for coverage.

packages/dev-common/Configs/eslint.config.mjs
```diff
@@ -305,7 +304,11 @@ export default defineConfig([
                 },
             ],
 
-            // Rules we're not ready to enable yet
+            // Rules we never want to enable because we disagree with them for this codebase
+
+            // We're fine colocating multiple small components in one file on certain occassions and we'll enforce this
```
dsargent, line 309: Typo in "occasions" but nowadays I'm so tired of AI generated perfection that a typo is refreshing.

packages/dev-common/Configs/eslint.config.mjs
dsargent: In general, these rules aren't gospel; they're always up for discussion. They are there to serve us, rather than vice versa. I do like to "lower the limbo stick" over time and tighten things up, but if there are rules that just get in the way and we don't value them, we can discuss disabling as appropriate.

packages/ui-common/components/Settings/ApiKeyErrorBanner.tsx
```diff
@@ -0,0 +1,80 @@
...
+    readonly vendor: LLMProvider
+    readonly result: KeyValidationFailure
+}
+
+/**
+ * The banner header for a failure: `<vendor> — <reason phrase> (<status>)`, falling back to
+ * `<vendor> — Request failed` when the status has no reason phrase, or `<vendor> — Request failed`
+ * with no code when the failure carries no status at all.
+ */
+export const errorTitle = ({vendor, result: {status: statusCode}}: ApiKeyFailure): string =>
```
dsargent, line 25: Naming note: almost always want a verb phrase for functions -- `getErrorTitle`, `formatTitle`, `formatError`, `toErrorTitle`, `asErrorTitle` (okay the last two are not strictly verb phrases, more functional programming idioms, but they still imply _action_.)

dsargent (approved):

PR #489, opened 2026-07-24
Switched the yarn test script from a temp dir to no-experimental-webstorage flag.

swensel:
## Description

Switched the `yarn test` script from `--localstorage-file=$(mktemp --directory)/node-storage` to `--no-experimental-webstorage`.

### Problem

Tests were failing intermittently in CI (for example `App.test.tsx > should render the page with darkMode=true`). `$(mktemp --directory)` was expanded once when the script started, so **every Vitest worker pointed at the same localStorage file**. Under Node 26 that flag enables Node's file-backed global Web Storage, and jsdom was also defining `localStorage`, so which one a test hit depended on setup order. The shared file let state leak across test files running in parallel workers, which produced the failures.

### Root cause of the darkMode failure

The leaked value is `branding.customer` in the zustand `app-settings` store (written by the Settings tests). The zustand `persist` middleware **rehydrates at module import**, before any `beforeEach`, and `localStorage.clear()` in `strictMocks` clears storage but does not reset already-hydrated in-memory state. So when App's worker imported the store while the shared file held a `customer`, `hasCustomer` stayed true for the whole file, the dark-mode toggle's `onClick` no-oped, and the click never applied dark mode, giving `var(--bs-gray-dark)` instead of `var(--bs-yellow)`.

Note this is **not** MUI's `mui-mode`. MUI re-reads that on every mount (after the per-test `clear()`), so it can't survive across tests. Only import-time-hydrated state (zustand) leaks.

### The fix

`--no-experimental-webstorage` turns off Node's Web Storage entirely, so jsdom is the sole provider and each test environment gets its own isolated in-memory `localStorage`/`sessionStorage`. Parallel worker runs stay clean and there's no temp file.

Worth noting:
- It disables `sessionStorage` too, which is what we want: `Settings` keeps API keys in `sessionStorage`, so now both stores come from jsdom and both get isolated per test environment.
- It removes the Node-vs-jsdom ambiguity. jsdom's storage is the only one, so behavior is deterministic regardless of setup order.

Only the `test` script changes; test behavior is unchanged.

### Reference

This [commit to `main`](https://github.com/cognizant-ai-lab/neuro-san-ui/actions/runs/30123364471/job/89581028357) hit the same failure. Screenshot below.

Screenshot 2026-07-24 at 4 08 24 PM


knip.config.ts
```diff
@@ -26,13 +26,6 @@ import {config as baseConfig} from "./packages/dev-common/Configs/knip.config"
 const config: KnipConfig = {
     ...baseConfig,
 
-    ignoreBinaries: [
```
swensel, line 29: Dropped `mktemp` from `ignoreBinaries` in `knip.config.ts`, since it was only there for the test script and knip now flags it as unused.

package.json
```diff
@@ -22,7 +22,7 @@
         "lint-rules": "./build_scripts/eslint_list_rules.sh",
         "lint-summary": "eslint --format summary",
         "start": "yarn workspace @cognizant-ai-lab/neuro-san-web start",
-        "test": "NODE_OPTIONS=\"--localstorage-file=$(mktemp --directory)/node-storage\" vitest run",
+        "test": "NODE_OPTIONS=\"--no-experimental-webstorage\" vitest run",
```
swensel, line 25: I think this should work better than what we had. I was encountering some issues on my local running tests, and there also was this [commit](https://github.com/cognizant-ai-lab/neuro-san-ui/commit/6a8cd3c6e4150ea7a7b097f50270d7167578c3df) to `main`. Please see the PR description for more information.

package.json
```diff
@@ -22,12 +22,12 @@
...
```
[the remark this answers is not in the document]
swensel: That is no longer true as of this [commit](https://github.com/cognizant-ai-lab/neuro-san-ui/commit/b5148a3547250958870cc7d2931a9c4a54921f1d).

PR #493, opened 2026-07-29
Dependabot fixes - update libs and regenerate yarn.lock from scratch.

swensel:
# Description

Dependabot fixes:
- Update `next`, `next-auth`, `sharp`, and `uuid`. 
- Remove `js-yaml` resolution. 
- Regenerate `yarn.lock` from scratch (needed to update `js-yaml `resolution and it may also fix a `postcss` issue).


package.json
```diff
@@ -61,8 +61,5 @@
         "vite": "catalog:default",
         "vitest": "catalog:default",
         "vitest-fail-on-console": "catalog:default"
-    },
-    "resolutions": {
```
swensel, line 65: Always good to remove `resolutions`, and at this point it is causing an issue.

.yarnrc.yml
```diff
@@ -28,8 +28,8 @@ catalogs:
         "globals": "17.4.0"
         "http-status": "1.7.3"
         "lodash-es": "4.18.1"
-        "next": "16.2.6"
-        "next-auth": "5.0.0-beta.30"
+        "next": "16.2.12"
+        "next-auth": "5.0.0-beta.32"
```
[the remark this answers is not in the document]
swensel: Good feedback. Updating.

.yarnrc.yml
```diff
@@ -28,8 +28,8 @@ catalogs:
         "globals": "17.4.0"
         "http-status": "1.7.3"
         "lodash-es": "4.18.1"
-        "next": "16.2.6"
-        "next-auth": "5.0.0-beta.30"
+        "next": "16.2.12"
```
swensel, line 31: FWIW, `neuro-ui` is on [next 16.2.2](https://github.com/cognizant-ai-lab/neuro-ui/blob/main/package.json#L45C9-L45C25).

packages/ui-common/package.json
```diff
@@ -55,8 +55,8 @@
         "@mui/material": "^9.0.0",
         "@mui/system": "^9.0.0",
         "@mui/x-tree-view": "^9.1.0",
-        "next": "^16.2.6",
-        "next-auth": "5.0.0-beta.30",
+        "next": "^16.2.12",
+        "next-auth": "5.0.0-beta.32",
```
swensel, line 59: I do notice we aren't using `^` for this one. Not sure / I forget why. Maybe because it's a beta release?

yarn.lock
```diff
@@ -52,9 +52,9 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@auth/core@npm:0.41.0":
-  version: 0.41.0
-  resolution: "@auth/core@npm:0.41.0"
+"@auth/core@npm:0.41.3":
```
swensel, line 55: Regenerate `yarn.lock` from scratch (needed to update `js-yaml `resolution and it may also fix a `postcss` issue).

packages/ui-common/package.json
dsargent: I had to use this _exact_ "Goldilocks" version because:

1) Earlier versions didn't support ESM properly (issues with importing in Neuro AI). Hence using the `beta` version, which we agreed was okay because this is only used for local dev (in MAUI at least).
2) Later versions generated errors on use.

So we were stuck on the tightrope between the two; `^` would allow floating to versions that didn't work for us.

package.json
dsargent: Yes, 💯

dsargent (approved):
lgtm

Issue #440, opened 2026-06-26
Make MAUI "authentication enabled" a runtime setting (rather than build time)


donn-leaf:
In order to get the no-auth use case running at neuro-san.evolution.ml I had to do a special build where i set
NEXT_PUBLIC_ENABLE_AUTHENTICATION=false

The desire is to set this at runtime.

PR #495, opened 2026-07-30
#440 Make MAUI "authentication enabled" a runtime setting

dsargent:
## Overview

Technical PR only. No UI/UX/user-facing changes.

Makes it so that `enableAuthentication` is a _runtime_ setting, rather than having to be baked into the Docker image. This means DevOps folks/automation no longer need to build a separate "no auth" Docker image and can simply configure the env var for the pod to access at runtime, just like other dynamic settings.

Previously: env var baked into build, backed into client bundle, static.
With this PR: retrieved at runtime from server env vars via `/environment` endpoint.

## Testing
This is risky. I don't have a spare AWS ALB sitting around on my laptop so I could only test these cases:
1) No authentication
2) auth.js authentication via Corp SSO but entirely local

Specifically could _not_ test the ALB case, so that's a risk when we deploy it to the cloud. We shall see.

## DevOps notes
@donn-leaf 👀 
Configuring:
Set environment variable in pod/runtime:

`ENABLE_AUTHENTICATION=false` -or
`ENABLE_AUTHENTICATION=true`

In fact, any value other than the exact lower case string `false` will be interpreted as `true`, meaning, "enable authentication". So if you want to disable authentication, you _have_ to set it to `false` and exactly that.

⚠️ Warning: do not change the setting on a running pod. It is expected to be set at pod startup and not messed with after that. We are doing some "duct tape and string" stuff with React hooks to make this work, and changing the value could lead to weird errors, crashes, or other things.

## Details

This was a huge slog. I had it working quickly, but then Devin jumped in and pointed out some very relevant issues with the existing implementation, that were exacerbated by making this setting dynamic. It was a real rabbit hole. We were doing unwholesome things in `_app.tsx` like forcing React to remount the whole tree after we learned what the authentication setting was.

Fundamentally, our authentication infrastructure is complex edifice of duct tape and string. We are trying to satisfy multiple, sometimes conflicting, constraints. See [Confluence doc](https://leaf-ai.atlassian.net/wiki/x/AQBPtQ) for an internal-only write-up.

This PR addresses _most_ of the issues Devin reported with remounting the tree multiple times, and the mishandling of "unknown" authentication setting. But we are still barely hanging on by the skin of our teeth: I still had to spot-disable the "rules of hooks" about calling hooks conditionally, with a pinkie-promise that it won't change after initial render. The only thing guaranteeing that promise is our behavior: configure the env var once for the pod or process, and treat it as immutable for that pod or process's lifetime.


.github/workflows/build.yml
```diff
@@ -120,17 +120,3 @@ jobs:
                       NODEJS_VERSION=${{ env.NODEJS_VERSION }}
                       NEXT_PUBLIC_NEURO_SAN_UI_VERSION=${{ inputs.DISPLAY_VERSION }}
                   quiet: "true"
-
-            - name: Build and push no-auth Docker image to ECR
```
dsargent, line 124: @donn-leaf I think we no longer need this due to this PR?

apps/main/.env.sample
```diff
@@ -2,8 +2,8 @@
 # you have access to. Make sure to include the port number if it's not running on the default HTTP port (80).
 NEURO_SAN_SERVER_URL=<URL for your neuro-san server including port number>
 
-# Set this to false for testing or authentication via load balancer (recommended)
-NEXT_PUBLIC_ENABLE_AUTHENTICATION=false
+# Set this to false for testing. Should be set to true in production for security reasons.
```
dsargent, line 5: Fixed comment and changed env var name. 

The `NEXT_PUBLIC_` thing is a Next.js hack for freezing env var values and sending them to the browser. Not needed once we "go dynamic".

apps/main/Const.ts
```diff
@@ -0,0 +1,2 @@
+// Name of the environment variable that controls whether authentication is enabled or not
+export const enableAuthenticationEnvVar = "ENABLE_AUTHENTICATION"
```
dsargent, line 2: Centralize env var name

apps/main/Dockerfile
```diff
@@ -87,10 +87,6 @@ COPY apps/main ./apps/main
 ARG NEXT_PUBLIC_NEURO_SAN_UI_VERSION
 ENV NEXT_PUBLIC_NEURO_SAN_UI_VERSION=${NEXT_PUBLIC_NEURO_SAN_UI_VERSION}
 
-# Extract authentication setting
```
dsargent, line 90: @donn-leaf not needed I think? Check my math pls.

apps/main/__tests__/pages/App.test.tsx
```diff
@@ -77,7 +76,16 @@ vi.mock("next/router", () => ({
     useRouter: vi.fn(),
 }))
 
-vi.mock("../../../../packages/ui-common/utils/Authentication")
+vi.mock("../../../../packages/ui-common/utils/Authentication", async () => {
```
dsargent, line 79: Only mock the part we want to mock and let the rest remain active.

apps/main/__tests__/pages/App.test.tsx
```diff
@@ -92,14 +100,24 @@ describe("Main App Component", () => {
     const testSupportEmailAddress = "test@example.com"
     const testLogoServiceToken = "testLogoServiceToken"
 
+    const mockEnvironment = (enableAuthentication: boolean | undefined) =>
```
dsargent, line 103: Factor out common mock values.

apps/main/__tests__/pages/App.test.tsx
```diff
@@ -165,6 +174,19 @@ describe("Main App Component", () => {
         expect(state.logoServiceToken).toBe(testLogoServiceToken)
     })
 
+    // This is the case where we haven't yet retrieved the environment variables, so we don't know if authentication
```
dsargent, line 177: Test for new (improved) handling of "don't yet know auth setting" case. This was one of Devin's comments.

apps/main/__tests__/pages/App.test.tsx
```diff
@@ -185,11 +207,39 @@ describe("Main App Component", () => {
...
 
         // Both fetches should have failed
-        expect(consoleSpy).toHaveBeenCalledWith(expect.stringContaining("Failed to fetch environment variables"))
-        expect(consoleSpy).toHaveBeenCalledWith(expect.stringContaining("Failed to fetch user info"))
+        expect(consoleSpy).toHaveBeenCalledExactlyOnceWith(
+            expect.stringContaining("Failed to fetch environment variables")
+        )
+    })
+
+    it("Should handle failure to fetch user info", async () => {
```
dsargent, line 218: Test this path explicitly. We now "fail fast" on "undefined" auth setting, so we don't reach this path via the existing tests any more.

apps/main/__tests__/pages/App.test.tsx
```diff
@@ -209,13 +259,16 @@ describe("Main App Component", () => {
 
     it("Should launch the tour when the items is selected from the Help menu", async () => {
         render(APP_COMPONENT)
+        await screen.findByText(COMPONENT_BODY)
 
         const dispatchSpy = vi.spyOn(window, "dispatchEvent")
 
-        const helpToggle = await screen.findByText("Help")
-        await user.click(helpToggle)
+        // Simulate clicking on the Help menu to open it
```
dsargent, line 266: Simplify

apps/main/__tests__/pages/ErrorPage.test.tsx
```diff
@@ -20,14 +20,9 @@ import {userEvent, UserEvent} from "@testing-library/user-event"
 import {withStrictMocks} from "../../../../__tests__/common/strictMocks"
 import {default as ErrorPage} from "../../../../packages/ui-common/components/ErrorPage/ErrorPage"
 import {LOGO} from "../../../../packages/ui-common/const"
+import {useEnvironmentStore} from "../../../../packages/ui-common/state/Environment"
 import {smartSignOut, useAuthentication} from "../../../../packages/ui-common/utils/Authentication"
 
-vi.mock("next-auth/react", () => ({
```
dsargent, line 25: Don't need to mock this here.

apps/main/__tests__/pages/ErrorPage.test.tsx
```diff
@@ -50,6 +45,8 @@ describe("ErrorPage", () => {
         vi.mocked(useAuthentication).mockReturnValue({
             data: {user: {name: "mock-user", image: "mock-image-url"}},
         })
+
+        useEnvironmentStore.getState().setEnableAuthentication(true)
```
dsargent, line 49: New way to mock authentication setting.

apps/main/__tests__/pages/MultiAgentAcceleratorPage.test.tsx
```diff
@@ -44,8 +44,6 @@ vi.mock("../../../../packages/ui-common/components/MultiAgentAccelerator/MultiAg
     },
 }))
 
-const renderMultiAgentAcceleratorPage = () => render(<MultiAgentAcceleratorPage />)
```
dsargent, line 47: Pointless wrapper function

apps/main/__tests__/unit/Api/Environment/Environment.test.ts
```diff
@@ -42,19 +44,21 @@ describe("Environment API handler", () => {
         process.env["AUTH0_DOMAIN"] = "test-domain.auth0.com"
         process.env["SUPPORT_EMAIL_ADDRESS"] = "support@example.com"
         process.env["LOGO_SERVICE_TOKEN"] = "test-logo-service-token"
+        process.env[enableAuthenticationEnvVar] = "true"
```
dsargent, line 47: Mock new env var

apps/main/__tests__/unit/Api/Environment/Environment.test.ts
dsargent, line 1: Updates for new env var mechanism for authentication toggle.

apps/main/__tests__/unit/General/instrumentation.test.ts
```diff
@@ -19,41 +19,57 @@ Tests for instrumentation.ts Next.js startup file.
...
     withStrictMocks()
 
     beforeEach(() => {
-        vi.mocked(authenticationEnabled).mockReturnValue(true)
+        // Default to "authentication enabled"
+        process.env[enableAuthenticationEnvVar] = "true"
         setAllEnvVars()
     })
 
+    const expectConsoleOutput = (
```
dsargent, line 53: Beefed up these tests to actually assert on expected output.

apps/main/__tests__/unit/General/instrumentation.test.ts
```diff
@@ -19,41 +19,57 @@ Tests for instrumentation.ts Next.js startup file.
...
 vi.mock("../../../../../packages/ui-common/const")
 
-// It's okay to do this in tests
-/* eslint-disable @typescript-eslint/no-dynamic-delete */
-
 const setAllEnvVars = () => {
     // Set all required environment variables
     REQUIRED_ENV_VARS.forEach((envVar) => {
-        process.env[envVar] = "test_value"
+        process.env[envVar] = `${envVar}-test_value`
```
dsargent, line 30: I like to sprinkle a little entropy in my tests: use different values for each env var by prepending the name.

apps/main/__tests__/unit/General/instrumentation.test.ts
dsargent, line 1: Changes for new env var mechanism for authentication.

apps/main/deploy/build.sh
```diff
@@ -20,7 +20,6 @@ export SERVICE_VERSION=${SERVICE_VERSION:-0.0.1}
 
 export NODEJS_VERSION=${NODEJS_VERSION:-26}
 export NEXT_PUBLIC_NEURO_SAN_UI_VERSION=${NEXT_PUBLIC_NEURO_SAN_UI_VERSION:-"dev-${USER}-$(date +'%Y-%m-%d-%H-%M')"}
-export NEXT_PUBLIC_ENABLE_AUTHENTICATION=${NEXT_PUBLIC_ENABLE_AUTHENTICATION:-false}
```
dsargent, line 23: @donn-leaf not needed any more

apps/main/deploy/build.sh
```diff
@@ -54,7 +53,6 @@ function build_main() {
         --platform ${TARGET_PLATFORM} \
         --build-arg NODEJS_VERSION="${NODEJS_VERSION}" \
         --build-arg NEXT_PUBLIC_NEURO_SAN_UI_VERSION="${NEXT_PUBLIC_NEURO_SAN_UI_VERSION}" \
-        --build-arg NEXT_PUBLIC_ENABLE_AUTHENTICATION="${NEXT_PUBLIC_ENABLE_AUTHENTICATION}" \
```
dsargent, line 57: @donn-leaf ditto

apps/main/deploy/run.sh
```diff
@@ -54,6 +54,7 @@ function run() {
         --publish $SERVICE_HTTP_PORT:$SERVICE_HTTP_PORT \
         --env LOGO_SERVICE_TOKEN \
         --env NEURO_SAN_SERVER_URL \
+        --env ENABLE_AUTHENTICATION \
```
dsargent, line 57: Added this here to propagate it to the container

apps/main/instrumentation.ts
```diff
@@ -44,32 +44,44 @@ export const REQUIRED_FOR_AUTH_ENV_VARS = [
  */
 export const OPTIONAL_ENV_VARS = ["LOGO_SERVICE_TOKEN", "OPENAI_API_KEY"]
 
+/**
+ * Check if an environment variable is "missing" (undefined or empty string).
+ * @param envVar - The name of the environment variable to check.
+ * @returns true if the environment variable is missing, false otherwise.
+ */
+const isMissing = (envVar: string) => !process.env[envVar]?.trim()
```
dsargent, line 52: Suggestion from Devin -- makes the "is missing" check more robust in the face of whitespace etc.

apps/main/instrumentation.ts
```diff
@@ -44,32 +44,44 @@ export const REQUIRED_FOR_AUTH_ENV_VARS = [
...
     // Conditionally required env vars
-    if (authenticationEnabled()) {
-        const missingRequiredForAuthEnvVars = REQUIRED_FOR_AUTH_ENV_VARS.filter((envVar) => !process.env[envVar])
+    const enableAuthentication = process.env[enableAuthenticationEnvVar] !== "false"
+    if (enableAuthentication) {
+        const missingRequiredForAuthEnvVars = REQUIRED_FOR_AUTH_ENV_VARS.filter((envVar) => isMissing(envVar))
         if (missingRequiredForAuthEnvVars.length > 0) {
             throw new Error(
-                "Error: The following environment variable(s) are required for authentication " +
+                "The following environment variable(s) are required for authentication " +
```
dsargent, line 67: "Error:" in an exception is redundant.

apps/main/instrumentation.ts
```diff
@@ -44,32 +44,44 @@ export const REQUIRED_FOR_AUTH_ENV_VARS = [
...
-    const missingOptionalEnvVars = OPTIONAL_ENV_VARS.filter((envVar) => !process.env[envVar])
+    const missingOptionalEnvVars = OPTIONAL_ENV_VARS.filter((envVar) => isMissing(envVar))
     if (missingOptionalEnvVars.length > 0) {
         console.warn(
             "Warning: The following optional environment variable(s) are empty or undefined. " +
                 `Some features may not work correctly:\n${missingOptionalEnvVars.join("\n")}`
         )
     }
+
+    console.info("Start-up: Environment variables checked successfully.")
```
dsargent, line 82: @donn-leaf added these to help us all with diagnostics. On start-up, the UI pod now spews what it sees.

Example:

```console
▲ Next.js 16.2.12 (Turbopack)
- Local:         http://localhost:3000
- Network:       http://0.0.0.0:3000
- Environments: .env
✓ Ready in 181ms
Start-up: Environment variables checked successfully.
Authentication enabled: true
OpenAI API key: set
Logo service token: set
Neuro SAN server URL: https://neuro-san-dev.decisionai.ml
```

apps/main/pages/_app.tsx
dsargent, line 1: Cleaned up our handling of auth/no-auth/don't know cases here. This is where a lot of Devin's comments landed. 

It's better, but still messy -- see my write-up in Confluence for why.

apps/main/pages/_app.tsx
```diff
@@ -56,19 +56,20 @@ type ExtendedAppProps = AppProps & {
 
 const DEFAULT_APP_NAME = `Cognizant ${LOGO}`
 
+const onStartTour = () => {
```
dsargent, line 59: This is stateless and pure, so I extracted it to Global scope.

apps/main/pages/_app.tsx
```diff
@@ -56,19 +56,20 @@ type ExtendedAppProps = AppProps & {
...
+    // Dispatches the "start tour" event
+    window.dispatchEvent(new CustomEvent(TRIGGER_APP_TOUR_EVENT_NAME))
+}
+
 /**
- * Utility component to pass user data along to Navbar. We have to make this a separate component because we need
- * to use useAuthentication for this, and "rules of hooks" apply.
+ * Utility component to pass user data along to Navbar. This component exists to isolate the main app from
```
dsargent, line 65: Update doc

apps/main/pages/_app.tsx
```diff
@@ -80,22 +81,17 @@ const NavbarWrapper = (props: Omit<NavbarProps, "userInfo">): ReactElement => {
     )
 }
 
-/**
```
dsargent, line 83: Got rid of this -- it was part of the excessive remountings that Devin complained about.

apps/main/pages/_app.tsx
```diff
@@ -109,11 +105,7 @@ export const NeuroSanUI: FC<ExtendedAppProps> = ({Component, pageProps}): ReactJ
     const {currentUser, setCurrentUser, setPicture, oidcProvider, setOidcProvider} = useUserInfoStore()
 
     // Infer authentication type
-    const authenticationType = authenticationEnabled()
-        ? currentUser
-            ? `ALB using ${oidcProvider}`
-            : "NextAuth"
-        : "None"
+    const authenticationType = getAuthenticationType(enableAuthentication, currentUser, oidcProvider)
```
dsargent, line 108: Extracted this to a common function since we do the same check elsewhere.

apps/main/pages/_app.tsx
```diff
@@ -167,8 +159,9 @@ export const NeuroSanUI: FC<ExtendedAppProps> = ({Component, pageProps}): ReactJ
             setBackendNeuroSanApiUrl(data.backendNeuroSanApiUrl)
             setSupportEmailAddress(data.supportEmailAddress)
             setLogoServiceToken(data.logoServiceToken)
+            setEnableAuthentication(data.enableAuthentication)
```
dsargent, line 162: This now comes back in the "get environment" API call, like all dynamic vars.

apps/main/pages/_app.tsx
```diff
@@ -167,8 +159,9 @@ export const NeuroSanUI: FC<ExtendedAppProps> = ({Component, pageProps}): ReactJ
...
 
-            if (authenticationEnabled()) {
+            if (data.enableAuthentication) {
```
dsargent, line 164: Check value returned from API here.

apps/main/pages/_app.tsx
```diff
@@ -222,8 +222,11 @@ export const NeuroSanUI: FC<ExtendedAppProps> = ({Component, pageProps}): ReactJ
             }
         }
 
-        void getUserInfo()
-    }, [setCurrentUser, setOidcProvider, setPicture])
+        if (enableAuthentication !== undefined) {
```
dsargent, line 225: This is part of the clean-up: we now explicitly check for the "we don't know what the auth setting is" case up front, and we don't attempt to set up the user info until we know. Until then, spinner. This prevents mounting one way (auth = on) then discarding the whole tree and remounting with "auth = off" if that is the actual setting.

apps/main/pages/_app.tsx
```diff
@@ -276,8 +279,12 @@ export const NeuroSanUI: FC<ExtendedAppProps> = ({Component, pageProps}): ReactJ
         }
     }
 
+    const isAuthenticationKnown = enableAuthentication !== undefined
+
     if (pathname === "/") {
-        // Main page is special
+        // Main page is special: doesn't have usual header and footer, does have a background image.
```
dsargent, line 285: Add doc as this is far from obvious.

apps/main/pages/_app.tsx
```diff
@@ -294,43 +301,46 @@ export const NeuroSanUI: FC<ExtendedAppProps> = ({Component, pageProps}): ReactJ
                 />
             </div>
         )
+    } else if (!isAuthenticationKnown) {
+        body = <LoadingSpinner id="loading-header" />
```
dsargent, line 305: Part of the clean-up: if we don't have the auth setting yet, display a spinner, the correct behavior for loading state/"we don't know yet".

apps/main/pages/_app.tsx
```diff
@@ -294,43 +301,46 @@ export const NeuroSanUI: FC<ExtendedAppProps> = ({Component, pageProps}): ReactJ
...
-                    </Container>
-                    <Footer
-                        supportEmailAddress={supportEmailAddress}
-                        logoLinkUrl="https://www.cognizant.com/"
-                        logoUrl="/cognizant-logo-white.svg"
-                        sx={{borderTop: "none", marginTop: 0}}
-                    />
-                </ErrorBoundary>
-            </NullableSessionProvider>
+        const appShell = (
```
dsargent, line 307: Extract out the common part and get rid of the awkward `NullableSessionProvider` sandwich.

apps/main/pages/_app.tsx
```diff
@@ -294,43 +301,46 @@ export const NeuroSanUI: FC<ExtendedAppProps> = ({Component, pageProps}): ReactJ
...
+                <Footer
+                    supportEmailAddress={supportEmailAddress}
+                    logoLinkUrl="https://www.cognizant.com/"
+                    logoUrl="/cognizant-logo-white.svg"
+                    sx={{borderTop: "none", marginTop: 0}}
+                />
+            </ErrorBoundary>
         )
+
+        body = enableAuthentication ? <SessionProvider>{appShell}</SessionProvider> : appShell
```
dsargent, line 343: We only reach here when we _know_ the authentication enabled setting, so we only evaluate this once we're ready to render the actual UI. We still have to have the switch as if authentication is disabled, we don't want to wrap in next.js's `SessionProvider` or else it will scream if all the next-auth stuff isn't configured.

apps/main/pages/api/environment/Types.ts
```diff
@@ -18,6 +18,7 @@ limitations under the License.
 export interface EnvironmentResponse {
     readonly auth0ClientId?: string
     readonly auth0Domain?: string
+    readonly enableAuthentication?: boolean
```
dsargent, line 21: New dynamic setting

apps/main/pages/api/environment/index.ts
```diff
@@ -29,18 +30,20 @@ const handler = (_req: NextApiRequest, res: NextApiResponse<EnvironmentResponse>
     res.setHeader("Content-Type", "application/json")
 
     const backendNeuroSanApiUrl = process.env["NEURO_SAN_SERVER_URL"]
+    const enableAuthentication = process.env[enableAuthenticationEnvVar] !== "false"
```
dsargent, line 33: Expose new dynamic auth setting in API

packages/dev-common/Configs/eslint.config.mjs
```diff
@@ -541,6 +541,7 @@ export default defineConfig([
             "@next/next/no-img-element": "off",
             "react/display-name": "off",
             "react/no-array-index-key": "off",
+            "@typescript-eslint/no-dynamic-delete": "off",
```
dsargent, line 544: We don't care about this in tests so disable it for test scope

packages/ui-common/__tests__/components/Common/Navbar.test.tsx
```diff
@@ -43,18 +43,19 @@ describe("Navbar", () => {
         render(
             <ThemeProvider theme={createTheme({colorSchemes: {light: true, dark: true}})}>
                 <Navbar
+                    authenticationType=""
```
dsargent, line 46: Alphabetize + new `enableAuthentication` prop.

packages/ui-common/__tests__/unit/Utils/Authentication.test.ts
```diff
@@ -44,7 +45,7 @@ describe("useAuthentication", () => {
         oldFetch = window.fetch
         window.fetch = mockFetch({})
 
-        process.env["NEXT_PUBLIC_ENABLE_AUTHENTICATION"] = "true"
+        useEnvironmentStore.getState().setEnableAuthentication(true)
```
dsargent, line 48: Use new mechanism

packages/ui-common/__tests__/unit/Utils/Authentication.test.ts
```diff
@@ -54,11 +55,20 @@ describe("useAuthentication", () => {
...
             // Disable auth for this test
-            process.env["NEXT_PUBLIC_ENABLE_AUTHENTICATION"] = "false"
+            useEnvironmentStore.getState().setEnableAuthentication(false)
+
             const {result} = renderHook(() => useAuthentication())
             expect(result.current.data.user.name).toBe(DEFAULT_USERNAME)
             expect(result.current.data.user.image).toBe(DEFAULT_USER_IMAGE)
         })
+
+        it("Throws if called before authentication setting is retrieved", () => {
```
dsargent, line 65: Test new contract (throws if auth not known)

packages/ui-common/components/Common/Navbar.tsx
```diff
@@ -63,6 +58,9 @@ export interface NavbarProps {
     // Info about the currently authenticated user
     readonly userInfo: {name: string; image: string}
 
+    // Flag indicating whether authentication is enabled
```
dsargent, line 61: Pass this in as a new prop

packages/ui-common/components/ErrorPage/ErrorPage.tsx
dsargent, line 1: Handle new authentication toggle mechanism

packages/ui-common/const.ts
```diff
@@ -35,5 +35,3 @@ export const DEFAULT_USERNAME = "Guest"
  * The default user image to use when the user does not have a profile picture.
  */
 export const DEFAULT_USER_IMAGE = "https://www.gravatar.com/avatar/?d=mp"
-
-export const authenticationEnabled = (): boolean => process.env["NEXT_PUBLIC_ENABLE_AUTHENTICATION"] !== "false"
```
dsargent, line 39: No longer needed or used.

packages/ui-common/state/Environment.ts
```diff
@@ -28,6 +28,11 @@ interface EnvironmentStore {
     backendNeuroSanApiUrl: string
     setBackendNeuroSanApiUrl: (backendNeuroSanApiUrl: string) => void
 
+    // Authentication on/off switch. "undefined" means "we don't know yet"
```
dsargent, line 31: New zustand store item for toggle

packages/ui-common/utils/Authentication.ts
```diff
@@ -34,12 +35,21 @@ export const AD_TENANT_ID = "de08c407-19b9-427d-9fe8-edf254300ca7"
  * of NextAuth's useSession hook.
  */
 export const useAuthentication = () => {
-    if (!authenticationEnabled()) {
+    const enableAuthentication = useEnvironmentStore((state) => state.enableAuthentication)
+
+    if (enableAuthentication === undefined) {
```
dsargent, line 40: This hook also is more resilient now: it refuses to make a decision if called with an "unknown" auth setting. Previously it treated that as "false", which was incorrect and dangerous.

packages/ui-common/utils/Authentication.ts
```diff
@@ -34,12 +35,21 @@ export const AD_TENANT_ID = "de08c407-19b9-427d-9fe8-edf254300ca7"
...
+    }
+
+    if (!enableAuthentication) {
         // Auth disabled: return a stub
         return {data: {user: {name: DEFAULT_USERNAME, image: DEFAULT_USER_IMAGE}}}
     }
-    // Auth enabled: safe to call hooks. Despite the conditional test above, we are guaranteed that
-    // authenticationEnabled is constant for a given build, so we won't be violating the rules of hooks.
+
+    // Auth mode is runtime configuration, not a build-time constant.
```
dsargent, line 49: See comment!

packages/ui-common/utils/Authentication.ts
```diff
@@ -34,12 +35,21 @@ export const AD_TENANT_ID = "de08c407-19b9-427d-9fe8-edf254300ca7"
...
+    // This hook relies on the app shell gating render until enableAuthentication is known.
+    // Components that can remain mounted while enableAuthentication changes must not use this hook.
+
     /* eslint-disable react-hooks/rules-of-hooks */
```
dsargent, line 53: We still have to silence the linter here, which is suboptimal but almost impossible to avoid. See write-up in Confluence for the contradictory requirements we have.

packages/ui-common/utils/Authentication.ts
```diff
@@ -117,3 +129,15 @@ export const smartSignOut = async (
         navigateToUrl(createAuth0LogoutUrl(oidcProvider, auth0Domain, auth0ClientId))
     }
 }
+
+/**
```
dsargent, line 133: Factored into a common function here.

apps/main/Dockerfile
dsargent: To be more precise: we _definitely_ don't need this particular arg/env var as the code no longer looks at it. 

My question is more, "is it okay to totally remove this since now the setting comes from a runtime env var which gets passed to the pod in the same way as other runtime variables, so doesn't even have to appear in the Dockerfile?"

apps/main/pages/_app.tsx
swensel: I'm not sure I'd call it pure, since `window.dispatchEvent` is a side effect?

apps/main/pages/_app.tsx
swensel: Not related to the changes in this PR, but destructuring these could be nice:

```js

const {
    backendNeuroSanApiUrl,
    supportEmailAddress,
    logoServiceToken,
    enableAuthentication,
} = data

setBackendNeuroSanApiUrl(backendNeuroSanApiUrl)
setSupportEmailAddress(supportEmailAddress)
setLogoServiceToken(logoServiceToken)
setEnableAuthentication(enableAuthentication)

if (enableAuthentication) {

swensel (dismissed):
LGTM, aside from (still) violating the rules of hooks, but it is what it is. Regardless, good work. 🎉

apps/main/pages/_app.tsx
dsargent: True, it does mutate the "state of the world". I just meant it doesn't change the state of any of our state variables, or zustand variables, or local storage, anything like that, so didn't need to be within the component. But yes, sending an event is a side effect.

apps/main/pages/_app.tsx
dsargent: Yes, would be a good improvement for future.

dsargent:
> LGTM, aside from (still) violating the rules of hooks, but it is what it is. Regardless, good work. 🎉

Yes, this. And if that env var value ever changes "mid stream", we will probably get a crash or at least scary warnings from React about "different number of hooks called on subsequent renders", because that is the exact rule we'd be violating. Hence my dire warnings about setting that env var up front and never changing it for the lifetime of the process. We _just_ manage to sneak by without upsetting React but it's fragile and precarious.

donn-leaf (approved):
Nice, thanks for all the digging to fix this. I have a follow up PR in the deployment repo.

Issue #431, opened 2026-06-22
Address dependabot issues in MAUI


swensel:
Lots of issues, a few `critical`. See: https://github.com/cognizant-ai-lab/neuro-san-ui/security/dependabot

PR #170, opened 2026-03-02
Rewrite build_scripts README and remove Codefresh files

donn-leaf:
## Summary

Replaces the outdated Codefresh-centric `build_scripts/README.md` with documentation that reflects the current GitHub Actions CI/CD setup. The old README described pipelines (`deploy-trigger.yml`, `build`, `deploy`) that ran on Codefresh and referenced ArgoCD manifest syncing — none of which applies anymore. The new README documents the Orchestrator → Test → Build → Deploy workflow chain, the npm publishing pipeline for `@cognizant-ai-lab/ui-common`, the weekly package cleanup job, and each shell script in the directory.

Closes #41

PR #149, opened 2026-02-12
#142 Make logo.dev token dynamic for easier DevOpsing, support generic brand logos

dsargent:
Can now be configured as an env var of the pod at runtime, rather than having to be baked in at Docker build time.

**Because of that I renamed it as well:**

 `LOGO_DEV_TOKEN=<token>`

Note: this is the free, publically-distributable token from logo.dev so it's [okay to expose it in the client app](https://www.logo.dev/docs/platform/api-keys#publishable-key-pk_) and therefore the user's browser. They promise to protect us from unauthorized usage for this type of token.

With new Footer -- I'm not super happy with how this looks (something "off" about it?!) but it gets the job done. @swensel open to discussion.

<br />
image
<br /><br />
New settings dialog allowing user to pick what kind of logo they want, including preview:
<br /><br />
image
<br /><br />
image
<br /><br />
image
<br /><br />

---

<a href="https://app.devin.ai/review/cognizant-ai-lab/neuro-san-ui/pull/149" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

Issue #139, opened 2026-01-30
Fix race condition for ThoughtBubbleOverlay test in MAUI


swensel:
From Darren: I found at least one cause of the intermittent branch coverage dip in tests. Line 306 in ThoughtBubbleOverlay.tsx sometimes gets hit by tests, but other times not. Since we count coverage down to the individual branch, that's enough to cause a failure. I'm guessing there's some kind of race condition or missing await or something like that in the vibed tests for thought bubbles. Else path not taken.

<img width="852" height="342" alt="Image" src="https://github.com/user-attachments/assets/0e89c4cf-ba53-45c9-9d41-630c3a72fbff" />

PR #497, opened 2026-07-31
#431 - More Dependabot

swensel:
# Description

Fix more dependabot. This is the follow up to https://github.com/cognizant-ai-lab/neuro-san-ui/pull/493.

Targeting this list with this PR:

Screenshot 2026-08-03 at 8 28 18 PM

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -9,7 +9,7 @@ catalogs:
     # The idea is that once https://github.com/yarnpkg/berry/issues/6925 is fixed we will have a catalog for
     # fixed requirements and another for requirements ranges, typically used in peerDependencies.
     default:
-        "@langchain/core": "1.1.17"
+        "@langchain/core": "1.2.4"
         "@mui/icons-material": "9.0.1"
         "@mui/material": "9.0.1"
         "@mui/system": "9.2.0"
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -28,7 +28,7 @@ catalogs:
         "globals": "17.4.0"
         "http-status": "1.7.3"
         "lodash-es": "4.18.1"
-        "next": "16.2.12"
+        "next": "16.3.0"
         "next-auth": "5.0.0-beta.32"
         "notistack": "3.0.2"
         "react": "19.2.4"
--- a/package.json
+++ b/package.json
@@ -31,7 +31,8 @@
     },
     "resolutions": {
         "postcss": "8.5.25",
-        "sharp": "0.35.3"
+        "sharp": "0.35.3",
+        "brace-expansion@^2.0.1": "2.1.4"
     },
     "devDependencies": {
         "@eslint/js": "10.0.1",
--- a/package.json
+++ b/package.json
@@ -31,7 +31,6 @@
     },
     "resolutions": {
         "postcss": "8.5.25",
-        "sharp": "0.35.3",
         "brace-expansion@^2.0.1": "2.1.4"
     },
     "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -29,6 +29,10 @@
     "engines": {
         "node": "^26.0.0"
     },
+    "resolutions": {
+        "postcss": "8.5.25",
+        "sharp": "0.35.3"
+    },
     "devDependencies": {
         "@eslint/js": "10.0.1",
         "@next/eslint-plugin-next": "16.2.10",
--- a/package.json
+++ b/package.json
@@ -30,8 +30,8 @@
         "node": "^26.0.0"
     },
     "resolutions": {
-        "postcss": "8.5.25",
-        "brace-expansion@^2.0.1": "2.1.4"
+        "brace-expansion@^2.0.1": "2.1.4",
+        "postcss": "8.5.25"
     },
     "devDependencies": {
         "@eslint/js": "10.0.1",
--- a/packages/ui-common/package.json
+++ b/packages/ui-common/package.json
@@ -55,7 +55,7 @@
         "@mui/material": "^9.0.0",
         "@mui/system": "^9.0.0",
         "@mui/x-tree-view": "^9.1.0",
-        "next": "^16.2.12",
+        "next": "^16.3.0",
         "next-auth": "5.0.0-beta.32",
         "react": "^19.2.4",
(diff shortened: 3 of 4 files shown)


.yarnrc.yml
```diff
@@ -9,7 +9,7 @@ catalogs:
     # The idea is that once https://github.com/yarnpkg/berry/issues/6925 is fixed we will have a catalog for
     # fixed requirements and another for requirements ranges, typically used in peerDependencies.
     default:
-        "@langchain/core": "1.1.17"
+        "@langchain/core": "1.2.4"
```
swensel, line 12: Upgrade to `@langchain/core@1.2.4`, which should fix the `uuid` alert ([#139](https://github.com/cognizant-ai-lab/neuro-san-ui/security/dependabot/139)). 

`@langchain/core@1.2.0` dropped the external `uuid` dependency dependency and included the patched (`uuid@≥11.1.1`) implementation in their library, so the vulnerable `uuid@10.0.0` leaves the dependency tree entirely.

package.json
```diff
@@ -29,6 +29,10 @@
     "engines": {
         "node": "^26.0.0"
     },
+    "resolutions": {
```
swensel, line 32: Flagging these since we generally avoid `resolutions`.

- `brace-expansion` ([#174](https://github.com/cognizant-ai-lab/neuro-san-ui/security/dependabot/174)): buried a few levels down (`openapi-typescript` -> `@redocly/openapi-core` -> `minimatch@5.1.9`), so there's no direct dep to bump and the resolution is the only lever. `@^2.0.1` scopes the override to just that instance, leaving the unaffected `1.x`/`5.x` copies alone. Comes out once `@redocly/openapi-core` ships a patched `minimatch` and `openapi-typescript` upgrades to pull it in (yes, we might be waiting for some time for that).
- `postcss` ([#170](https://github.com/cognizant-ai-lab/neuro-san-ui/security/dependabot/170)/[#169](https://github.com/cognizant-ai-lab/neuro-san-ui/security/dependabot/169)/[#109](https://github.com/cognizant-ai-lab/neuro-san-ui/security/dependabot/109)): Next `16.3.0` hard-pins `postcss@8.5.23`, so we force `8.5.25` (what `vite` already resolves) to clear the alerts. We can remove this once a new `Next` release stops pinning an old one (`16.3.0` was just released but it's still an issue there).

packages/ui-common/package.json
```diff
@@ -55,7 +55,7 @@
         "@mui/material": "^9.0.0",
         "@mui/system": "^9.0.0",
         "@mui/x-tree-view": "^9.1.0",
-        "next": "^16.2.12",
+        "next": "^16.3.0",
```
swensel, line 58: Matches `next` upgrade to `16.3.0` below.

.yarnrc.yml
```diff
@@ -28,7 +28,7 @@ catalogs:
         "globals": "17.4.0"
         "http-status": "1.7.3"
         "lodash-es": "4.18.1"
-        "next": "16.2.12"
+        "next": "16.3.0"
```
swensel, line 31: Bumping to `16.3.0` clears the `sharp` alert ([#149](https://github.com/cognizant-ai-lab/neuro-san-ui/security/dependabot/149)). `next@16.2.12` pulled its own `sharp@^0.34.5` (caps below `sharp@0.35.0`); `next@16.3.0` moves to an optional `sharp@^0.35.3` that lands on the patched build.

yarn.lock
```diff
@@ -455,7 +455,7 @@ __metadata:
     "@mui/material": ^9.0.0
     "@mui/system": ^9.0.0
     "@mui/x-tree-view": ^9.1.0
-    next: ^16.2.12
+    next: ^16.3.0
```
swensel, line 458: Several changes here from `yarn install`. I didn't delete and regenerate this from scratch.

swensel:
Build looks weird but I think that's due to https://github.com/cognizant-ai-lab/neuro-san-ui/pull/499 being merged. The checks that did run were successful.

.yarnrc.yml
dsargent: Langchain is notorious for breaking changes in new versions -- did you retest with the various Next.js LLM endpoints we have? (Branding recommendations etc.)

package.json
dsargent: > (what vite already resolves)

This is a red herring as Vite is _currently_ only a dev dependency, so in production builds this is _not_ happening.

dsargent: `brace-expansion (#174): buried a few levels down (openapi-typescript ->`

Honestly, I'd have just closed this one. We only use `openapi-typescript` at _build_ time -- nothing to do with the production build or deployment. But sure, why not.

dsargent (approved):

dsargent:
I would like to see _some_ mechanism (expiring Unicorn linter comment? Ticket? Reminder on our calendar??) to ensure we revisit the `resolutions` as they are the type of "temporary solution" that can very very easily become permanent, then it's a case of "we don't remember why but we're afraid to touch it" which then morphs into "if it ain't broke, don't fix it" and we're stuck with them. 

To put it another way: `resolutions` are a temporary solution to an ugly problem: either a vendor is slow to update to address CVEs, or two vendors aren't playing nicely together, both of which _should_ be very transient situations.

.yarnrc.yml
swensel: I tested these API endpoints: `/environment`, `/networkIconSuggestions`, `/agentIconSuggestions`, `/streaming_chat`. They all seemed to work fine.

swensel:
> I would like to see _some_ mechanism (expiring Unicorn linter comment? Ticket? Reminder on our calendar??) to ensure we revisit the `resolutions` as they are the type of "temporary solution" that can very very easily become permanent, then it's a case of "we don't remember why but we're afraid to touch it" which then morphs into "if it ain't broke, don't fix it" and we're stuck with them.
> 
> To put it another way: `resolutions` are a temporary solution to an ugly problem: either a vendor is slow to update to address CVEs, or two vendors aren't playing nicely together, both of which _should_ be very transient situations.

I filed https://github.com/cognizant-ai-lab/neuro-san-ui/issues/503 to follow up on this.

Issue #52, opened 2025-10-20
Add sly data input field


dsargent:
**Migrated from Jira:** [UN-3560](https://leaf-ai.atlassian.net/browse/UN-3560)

## Description
Use formatting plugin.  Came up as part of network builder which is reliant on `sly*data`.  List of coded tools might also come over `sly*data`.

There was a ticket for MAUI before to support `sly*data`, [https://leaf-ai.atlassian.net/browse/UN-3110](https://leaf-ai.atlassian.net/browse/UN-3110|smart-link), but that was to pass `sly*data` back to the neuro-san server, not to add an input to view or edit `sly_data`.

## Jira Details
- **Issue Type:** Task
- **Priority:** Major
- **Status:** To Do
- **Reporter:** Sean Wensel
- **Assignee:** Sean Wensel
- **Created:** 2025-09-29 15:10:54
- **Updated:** 2025-10-16 10:40:21

### Comment by Sean Wensel (2025-10-06 14:10:51)
Babak - if an image could be passed in, “future proof” to be able to show that?  Although we may not support that yet.

dsargent:
May be able to tap into `Zod.infer()` here for schema handling.

PR #498, opened 2026-08-01
#52 add sly data input field

M-Elsaied:
Addresses #52 

### Overview

* Adds a sly data editor to the chat: ⚙ options menu → "Sly data…" opens a dialog showing the selected network's `sly_data` as editable JSON, with import/export/clear and validation.

* Why: the server echoes `sly_data` every turn and the UI bounces it back, but users could never see or seed it. This makes the existing loop visible - no changes to the wire format, controller, or send path; the editor writes to the same per-network store the send path already reads.

* Implementation: `utils/SlyData.ts` (format/parse/image-detection helpers) · `setSlyData` store action (replace semantics, so keys can be deleted; streaming keeps merging via `updateSlyData`) · `SlyDataDialog` + menu wiring in `ChatCommon`. Values that look like images render as thumbnails .

* Verified: e2e vs a local neuro-san on `music_nerd_pro_sly` - typed `running_cost: 100`, Accountant returned `103` (its deliberate +3.0), echo shown back in the dialog. Screenshots below. 

screenshot-1785598855497-7

screenshot-1785598951963-8

- No new dependencies, endpoints, env vars, or API routes.
- Behaviour ported from nsflow's `EditorSlyDataPanel`, not the code - it uses
  `json-edit-react` and a localStorage cache, neither of which fits here. 

I am way out of my depth here, so expecting alot of push back. Bring those comments :)

### DIFF
--- a/packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
+++ b/packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
@@ -1595,6 +1595,36 @@ describe("ChatCommon", () => {
             expect(sendChatQuery).toHaveBeenCalledTimes(1)
         })
 
+        it("Should not persist echoed BYOK keys into the sly_data store", async () => {
+            act(() => {
+                useAgentChatHistoryStore.getState().resetHistory(TEST_AGENT_MATH_GUY)
+            })
+
+            renderChatCommonComponent()
+
+            const chunkWithSecret: ChatResponse = {
+                response: {
+                    type: ChatMessageType.AGENT_FRAMEWORK,
+                    text: "done",
+                    sly_data: {
+                        llm_config: {openai_api_key: "sk-should-never-persist"},
+                        running_cost: 3,
+                    },
+                },
+            }
+
+            mockSendChatQuery(({callback}) => {
+                callback(JSON.stringify(chunkWithSecret))
+            })
+
+            await sendQuery(TEST_AGENT_MATH_GUY, "query that echoes a secret")
+
+            await waitFor(() => {
+                const storedSlyData = useAgentChatHistoryStore.getState().history[TEST_AGENT_MATH_GUY]?.slyData
+                expect(storedSlyData).toEqual({running_cost: 3})
+            })
+        })
+
         const openSlyDataEditor = async () => {
             await user.click(screen.getByTestId("TuneIcon"))
             await user.click(screen.getByRole("menuitem", {name: "Sly data..."}))
@@ -1626,7 +1656,7 @@ describe("ChatCommon", () => {
                 "query after seeding sly_data",
                 TEST_AGENT_MATH_GUY,
                 expect.any(Function),
-                undefined, // no chat context yet: only sly_data has been stored for this network
+                null, // no chat context yet: setSlyData initializes a fresh entry with null context
                 {login: TEST_USER, seeded_by_user: "yes"},
                 TEST_USER,
                 StreamingUnit.Line
--- a/packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
+++ b/packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
@@ -1594,6 +1594,56 @@ describe("ChatCommon", () => {
             // sendChatQuery was called — the component processed the chunk without error
             expect(sendChatQuery).toHaveBeenCalledTimes(1)
         })
+
+        const openSlyDataEditor = async () => {
+            await user.click(screen.getByTestId("TuneIcon"))
+            await user.click(screen.getByRole("menuitem", {name: "Sly data..."}))
+            return screen.getByLabelText<HTMLTextAreaElement>("Sly data")
+        }
+
+        it("Should send sly_data edited in the editor with the next request", async () => {
+            act(() => {
+                useAgentChatHistoryStore.getState().resetHistory(TEST_AGENT_MATH_GUY)
+            })
+
+            renderChatCommonComponent()
+            mockSendChatQuery(({callback}) => {
+                callback(JSON.stringify({response: getResponseMessage(ChatMessageType.AGENT_FRAMEWORK, "done")}))
+            })
+
+            const editor = await openSlyDataEditor()
+            await user.clear(editor)
+            await user.click(editor)
+            await user.paste('{"seeded_by_user": "yes"}')
+            await user.click(screen.getByRole("button", {name: "Save"}))
+
+            await sendQuery(TEST_AGENT_MATH_GUY, "query after seeding sly_data")
+
+            // sly_data is the 7th argument to sendChatQuery. The user's value rides along with the entries the
+            // component always supplies.
+            expect(vi.mocked(sendChatQuery).mock.calls[0][6]).toEqual({
+                seeded_by_user: "yes",
+                login: TEST_USER,
+            })
+        })
+
+        it("Should show the sly_data currently in the store", async () => {
+            act(() => {
+                useAgentChatHistoryStore.getState().setSlyData(TEST_AGENT_MATH_GUY, {charges: 7})
+            })
+
+            renderChatCommonComponent()
+
+            expect(await openSlyDataEditor()).toHaveValue('{\n  "charges": 7\n}')
+        })
+
+        it("Should not offer the sly_data editor for legacy agents, which have no sly_data", async () => {
+            renderChatCommonComponent({selectedNetwork: LegacyAgentType.DataGenerator})
+
+            await user.click(screen.getByTestId("TuneIcon"))
+
+            expect(screen.queryByRole("menuitem", {name: "Sly data..."})).not.toBeInTheDocument()
+        })
     })
 
     describe("Chat Context Handling", () => {
--- a/packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
+++ b/packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
@@ -1619,12 +1619,18 @@ describe("ChatCommon", () => {
 
             await sendQuery(TEST_AGENT_MATH_GUY, "query after seeding sly_data")
 
-            // sly_data is the 7th argument to sendChatQuery. The user's value rides along with the entries the
-            // component always supplies.
-            expect(vi.mocked(sendChatQuery).mock.calls[0][6]).toEqual({
-                seeded_by_user: "yes",
-                login: TEST_USER,
-            })
+            // The user's value rides along with the entries the component always supplies (login).
+            expect(sendChatQuery).toHaveBeenCalledExactlyOnceWith(
+                undefined, // neuroSanURL is not set in these props
+                expect.any(AbortSignal),
+                "query after seeding sly_data",
+                TEST_AGENT_MATH_GUY,
+                expect.any(Function),
+                undefined, // no chat context yet: only sly_data has been stored for this network
+                {login: TEST_USER, seeded_by_user: "yes"},
+                TEST_USER,
+                StreamingUnit.Line
+            )
         })
 
         it("Should show the sly_data currently in the store", async () => {
--- a/packages/ui-common/__tests__/components/AgentChat/ChatCommon/SlyDataDialog.test.tsx
+++ b/packages/ui-common/__tests__/components/AgentChat/ChatCommon/SlyDataDialog.test.tsx
@@ -120,6 +120,9 @@ describe("SlyDataDialog", () => {
 
         expect(screen.getByRole("button", {name: "Save"})).toBeDisabled()
         expect(screen.getByRole("button", {name: "Export sly data"})).toBeDisabled()
+        // The parser's message is shown as the editor's helper text. Its exact wording is engine-dependent,
+        // but every variant mentions JSON.
+        expect(getEditor()).toHaveAccessibleDescription(/json/iu)
     })
 
     it("refuses to save a JSON array, since sly data has to be an object", async () => {
--- a/packages/ui-common/__tests__/components/AgentChat/ChatCommon/SlyDataDialog.test.tsx
+++ b/packages/ui-common/__tests__/components/AgentChat/ChatCommon/SlyDataDialog.test.tsx
@@ -238,7 +238,7 @@ describe("SlyDataDialog", () => {
     it("names the keys that get merged in on send, without showing their values", () => {
         renderDialog(["llm_config", "login"])
 
-        expect(screen.getByText(/added automatically to every request/u)).toBeInTheDocument()
+        expect(screen.getByText(/Sent automatically with every request/u)).toBeInTheDocument()
         expect(screen.getByText("llm_config")).toBeInTheDocument()
         expect(screen.getByText("login")).toBeInTheDocument()
     })
@@ -246,6 +246,6 @@ describe("SlyDataDialog", () => {
     it("says nothing about merged keys when there are none", () => {
         renderDialog()
 
-        expect(screen.queryByText(/added automatically to every request/u)).not.toBeInTheDocument()
+        expect(screen.queryByText(/Sent automatically with every request/u)).not.toBeInTheDocument()
     })
 })
--- a/packages/ui-common/__tests__/components/AgentChat/ChatCommon/SlyDataDialog.test.tsx
+++ b/packages/ui-common/__tests__/components/AgentChat/ChatCommon/SlyDataDialog.test.tsx
@@ -180,7 +180,7 @@ describe("SlyDataDialog", () => {
         renderDialog()
 
         const file = new File(['{"imported": true}'], "sly.json", {type: "application/json"})
-        await user.upload(screen.getByTestId(`${DIALOG_ID}-file-input`), file, {applyAccept: false})
+        await user.upload(screen.getByTestId(`${DIALOG_ID}-file-input`), file)
 
         expect(getEditor()).toHaveValue('{"imported": true}')
         expect(screen.getByRole("button", {name: "Save"})).toBeEnabled()
@@ -190,7 +190,7 @@ describe("SlyDataDialog", () => {
         renderDialog()
 
         const file = new File(["not json at all"], "sly.json", {type: "application/json"})
-        await user.upload(screen.getByTestId(`${DIALOG_ID}-file-input`), file, {applyAccept: false})
+        await user.upload(screen.getByTestId(`${DIALOG_ID}-file-input`), file)
 
         expect(getEditor()).toHaveValue("not json at all")
         expect(screen.getByRole("button", {name: "Save"})).toBeDisabled()
--- a/packages/ui-common/__tests__/components/AgentChat/ChatCommon/SlyDataDialog.test.tsx
+++ b/packages/ui-common/__tests__/components/AgentChat/ChatCommon/SlyDataDialog.test.tsx
@@ -1,0 +1,249 @@
+/*
+Copyright 2025 Cognizant Technology Solutions Corp, www.cognizant.com.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import {act, render, screen} from "@testing-library/react"
+import {userEvent, UserEvent} from "@testing-library/user-event"
+
+import {withStrictMocks} from "../../../../../../__tests__/common/strictMocks"
+import {SlyDataDialog} from "../../../../components/AgentChat/ChatCommon/SlyDataDialog"
+import {useAgentChatHistoryStore} from "../../../../state/ChatHistory"
+import {downloadFile} from "../../../../utils/File"
+
+vi.mock("../../../../utils/File", async () => {
+    const actual = vi.importActual("../../../../utils/File")
+    return {
+        ...(await actual),
+        downloadFile: vi.fn(),
+    }
+})
+
+const NETWORK_ID = "music_nerd_pro"
+const DIALOG_ID = "test-sly-data-dialog"
+
+// A one-pixel transparent GIF, as a data URI
+const IMAGE_DATA_URI = "data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7"
+
+describe("SlyDataDialog", () => {
+    withStrictMocks()
+
+    let user: UserEvent
+    const onCloseMock = vi.fn()
+
+    beforeEach(() => {
+        user = userEvent.setup({delay: null})
+
+        // Start each test from a clean store
+        const agentHistory = useAgentChatHistoryStore.getState().history
+        Object.keys(agentHistory).forEach((agentId) => {
+            useAgentChatHistoryStore.getState().resetHistory(agentId)
+        })
+    })
+
+    const renderDialog = (extraSlyDataKeys?: readonly string[]) =>
+        render(
+            <SlyDataDialog
+                extraSlyDataKeys={extraSlyDataKeys}
+                id={DIALOG_ID}
+                isOpen={true}
+                networkDisplayName="Music Nerd Pro"
+                networkId={NETWORK_ID}
+                onClose={onCloseMock}
+            />
+        )
+
+    const getEditor = () => screen.getByLabelText<HTMLTextAreaElement>("Sly data")
+
+    const getSlyData = () => useAgentChatHistoryStore.getState().history[NETWORK_ID]?.slyData
+
+    /**
+     * Replaces the whole contents of the editor. Uses paste rather than type because sly_data is JSON and
+     * userEvent's type() treats braces as syntax.
+     */
+    const replaceEditorContents = async (contents: string) => {
+        const editor = getEditor()
+        await user.clear(editor)
+        await user.click(editor)
+        await user.paste(contents)
+    }
+
+    it("shows the current sly data for the network, pretty-printed", () => {
+        act(() => useAgentChatHistoryStore.getState().setSlyData(NETWORK_ID, {charges: 2, user: "bob"}))
+
+        renderDialog()
+
+        expect(getEditor()).toHaveValue('{\n  "charges": 2,\n  "user": "bob"\n}')
+    })
+
+    it("shows an empty object when the network has no sly data yet", () => {
+        renderDialog()
+
+        expect(getEditor()).toHaveValue("{}")
+    })
+
+    it("saves edits to the store and closes", async () => {
+        act(() => useAgentChatHistoryStore.getState().setSlyData(NETWORK_ID, {charges: 2}))
+
+        renderDialog()
+        await replaceEditorContents('{"charges": 5}')
+        await user.click(screen.getByRole("button", {name: "Save"}))
+
+        expect(getSlyData()).toEqual({charges: 5})
+        expect(onCloseMock).toHaveBeenCalled()
+    })
+
+    it("removes keys the user deleted, rather than merging them back in", async () => {
+        act(() => useAgentChatHistoryStore.getState().setSlyData(NETWORK_ID, {charges: 2, user: "bob"}))
+
+        renderDialog()
+        await replaceEditorContents('{"charges": 2}')
+        await user.click(screen.getByRole("button", {name: "Save"}))
+
+        expect(getSlyData()).toEqual({charges: 2})
+    })
+
+    it("refuses to save malformed JSON and says why", async () => {
+        renderDialog()
+        await replaceEditorContents('{"charges": }')
+
+        expect(screen.getByRole("button", {name: "Save"})).toBeDisabled()
+        expect(screen.getByRole("button", {name: "Export sly data"})).toBeDisabled()
+    })
+
+    it("refuses to save a JSON array, since sly data has to be an object", async () => {
+        renderDialog()
+        await replaceEditorContents("[1, 2]")
+
+        expect(screen.getByText("Sly data must be a JSON object, not an array.")).toBeInTheDocument()
+        expect(screen.getByRole("button", {name: "Save"})).toBeDisabled()
+    })
+
+    it("recovers once the JSON is valid again", async () => {
+        renderDialog()
+        await replaceEditorContents("{oops")
+        expect(screen.getByRole("button", {name: "Save"})).toBeDisabled()
+
+        await replaceEditorContents('{"ok": true}')
+        expect(screen.getByRole("button", {name: "Save"})).toBeEnabled()
+    })
+
+    it("clears the editor without touching the store until the user saves", async () => {
+        act(() => useAgentChatHistoryStore.getState().setSlyData(NETWORK_ID, {charges: 2}))
+
+        renderDialog()
+        await user.click(screen.getByRole("button", {name: "Clear sly data"}))
+
+        expect(getEditor()).toHaveValue("{}")
+        expect(getSlyData()).toEqual({charges: 2})
+
+        await user.click(screen.getByRole("button", {name: "Save"}))
+        expect(getSlyData()).toEqual({})
+    })
+
+    it("discards edits when cancelled", async () => {
+        act(() => useAgentChatHistoryStore.getState().setSlyData(NETWORK_ID, {charges: 2}))
+
+        renderDialog()
+        await replaceEditorContents('{"charges": 99}')
+        await user.click(screen.getByRole("button", {name: "Cancel"}))
+
+        expect(getSlyData()).toEqual({charges: 2})
+        expect(onCloseMock).toHaveBeenCalled()
+    })
+
+    it("exports the sly data as a JSON file", async () => {
+        act(() => useAgentChatHistoryStore.getState().setSlyData(NETWORK_ID, {charges: 2}))
+
+        renderDialog()
+        await user.click(screen.getByRole("button", {name: "Export sly data"}))
+
+        expect(downloadFile).toHaveBeenCalledWith(
+            '{\n  "charges": 2\n}',
+            "music_nerd_pro_sly_data.json",
+            "application/json"
+        )
+    })
+
+    it("loads sly data from an imported file", async () => {
+        renderDialog()
+
+        const file = new File(['{"imported": true}'], "sly.json", {type: "application/json"})
+        await user.upload(screen.getByTestId(`${DIALOG_ID}-file-input`), file, {applyAccept: false})
+
+        expect(getEditor()).toHaveValue('{"imported": true}')
+        expect(screen.getByRole("button", {name: "Save"})).toBeEnabled()
+    })
+
+    it("explains why an imported file is unusable instead of silently dropping it", async () => {
+        renderDialog()
+
+        const file = new File(["not json at all"], "sly.json", {type: "application/json"})
+        await user.upload(screen.getByTestId(`${DIALOG_ID}-file-input`), file, {applyAccept: false})
+
+        expect(getEditor()).toHaveValue("not json at all")
+        expect(screen.getByRole("button", {name: "Save"})).toBeDisabled()
+    })
+
+    it("previews values that look like images", () => {
+        act(() => useAgentChatHistoryStore.getState().setSlyData(NETWORK_ID, {avatar: IMAGE_DATA_URI}))
+
+        renderDialog()
+
+        expect(screen.getByRole("img", {name: "avatar"})).toHaveAttribute("src", IMAGE_DATA_URI)
+    })
+
+    it("shows no image preview when there is nothing to preview", () => {
+        act(() => useAgentChatHistoryStore.getState().setSlyData(NETWORK_ID, {charges: 2}))
+
+        renderDialog()
+
+        expect(screen.queryByRole("img")).not.toBeInTheDocument()
+    })
+
+    it("follows the store while the user has not made any edits", () => {
+        renderDialog()
+
+        act(() => useAgentChatHistoryStore.getState().updateSlyData(NETWORK_ID, {charges: 3}))
+
+        expect(getEditor()).toHaveValue('{\n  "charges": 3\n}')
+    })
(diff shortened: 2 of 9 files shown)


M-Elsaied:
Not sure why the CI workflow didn't trigger. Ran tests locally (Node 26.5.0 / Yarn 4.12.0):

| Check | Result |
|---|---|
| `yarn test` | 68 files, 794 tests passed |
| `yarn tsc` | clean |
| `./build_scripts/run_eslint.sh` | clean |
| `yarn knip` | clean |
| `yarn prettier --check` (9 branch files) | clean |

Would be helpful if someone can check on their end as well

M-Elsaied:
Also good to note, is that by default it will be an empty dict `{}`assumption here is that the users knows (a) which network uses `sly-data` and (b) what are the `sly-data-schema` to use. This can be counter intuitive for users and maybe confuse the user experience. We can always fetch the information from the metadata, that would be addressed in separate issue if needed.
image

packages/ui-common/components/AgentChat/ChatCommon/Const.ts
```diff
@@ -1,2 +1,5 @@
 // Maximum number of turns to save
 export const MAX_TURNS = 50
+
+// The sly_data key under which we supply the current username. Some coded tools (data generator...) expect it.
+export const SLY_DATA_LOGIN_KEY = "login"
```
M-Elsaied, line 5: following the same structure `MultiAgentAccelerator/const.ts` which defines `AGENT_NETWORK_DEFINITION_KEY`, `AGENT_NETWORK_NAME_KEY`, `AGENT_RESERVATIONS_KE` defining it here and used in `ChatCommon.tsx` in 2 places ln 569 & 989. If you think it's an overkill, we can remove it.

packages/ui-common/utils/SlyData.ts
```diff
@@ -0,0 +1,159 @@
...
+/**
+ * Helpers for viewing and editing sly_data as JSON.
+ *
+ * sly_data is the side-channel payload that travels with each request to an agent network. It never reaches the LLM,
+ * but coded tools read from and write to it. It is "super loose": any JSON object the agent network cares to define.
+ */
+
+// How many levels deep to look for image values. Deep enough for realistic payloads, shallow enough that a
+// pathological structure can't lock up the browser.
+const MAX_IMAGE_SEARCH_DEPTH = 6
```
M-Elsaied, line 32: Depth and results cap are arbitrary; they exist only so a pathological payload can't lock the render. Tests pin both values, so changing them means changing the tests too.
Also no real visibility on the use case of passing an image as a sly-data so assumption is this will be updated depending on requirements, this is just the ground work

packages/ui-common/utils/SlyData.ts
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    return {value: parsed as Record<string, unknown>}
+}
+
+/**
+ * Determines whether a string value looks like something we can display as an image.
+ * @param value The string to test.
+ * @returns True if the value is an image data URI or an http(s) URL with an image file extension.
+ */
```
M-Elsaied, line 111: Image detection recognizes two encodings: `data:image/*` URIs and http(s) URLs with an image extension. Both are guesses

packages/ui-common/components/AgentChat/ChatCommon/SlyDataDialog.tsx
```diff
@@ -0,0 +1,379 @@
...
+        }
+    }, [isOpen, isDirty, storedText])
+
+    const parseResult = useMemo(() => parseSlyData(text), [text])
+
+    const images = useMemo(() => findImageValues(parseResult.value), [parseResult])
+
+    // The server echoed new sly_data while the user had unsaved edits. Saying so beats silently discarding
+    // either side's version.
+    const wasUpdatedByServer = isDirty && storedText !== seededText
```
M-Elsaied: If the user has no unsaved edits the editor silently follows the store; if they do, we show a Reload alert and leave their text alone.  Deliberate divergence from nsflow's `EditorSlyDataPanel`

packages/ui-common/state/ChatHistory.ts
```diff
@@ -130,6 +139,12 @@ export const useAgentChatHistoryStore = create<ChatHistoryStore>()(
                     const newHistory = {...state.history, [agentId]: {...existing, slyData: mergedSlyData}}
                     return {history: newHistory}
                 }),
+            setSlyData: (agentId: string, slyData: SlyData) =>
```
M-Elsaied, line 140: Replace-not-merge so the editor can delete keys; `updateSlyData` stays merge for streamed deltas

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -558,7 +563,11 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
                         // It's a Neuro-san agent.
 
                         // Some coded tools (data generator...) expect the username provided in slyData.
-                        const slyDataWithUsername = {...agentChatHistory?.slyData, ...extraSlyData, login: currentUser}
+                        const slyDataWithUsername = {
+                            ...agentChatHistory?.slyData,
+                            ...extraSlyData,
+                            [SLY_DATA_LOGIN_KEY]: currentUser,
```
M-Elsaied, line 572: Abstracted to variable in `Const.ts`

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -967,6 +976,18 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+        handleOptionsMenuClose()
+        setSlyDataDialogOpen(true)
+    }
+
+    /**
+     * The sly_data keys that this component merges in on every request, on top of whatever is in the store. The
+     * sly_data editor lists them so that the user isn't surprised by keys arriving at the server that they never
+     * typed -- and, for the API keys among them, so that we never have to display their values.
+     */
+    const getMergedSlyDataKeys = (): readonly string[] => [...Object.keys(extraSlyData ?? {}), SLY_DATA_LOGIN_KEY]
```
M-Elsaied, line 992: `slyDataWithUsername` is the payload; `getMergedSlyDataKeys` is the disclosure label for the parts of that payload the editor doesn't show.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -1005,6 +1026,18 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
                 </ListItemIcon>
                 <ListItemText primary="Wrap output" />
             </MenuItem>
+            {/* Legacy agents talk straight to an LLM and have no sly_data, so there is nothing to show for them */}
+            {!isLegacyAgentType(selectedNetwork) && (
```
M-Elsaied, line 1033: `isLegacyAgentType`is defined int `Types.ts` for 5 specific agents. Maybe we can have a conversation if this is needed as tagging is hardcoded

dsargent:
First comment: please fix the "Description". It looks like an LLM stream-of-consciousness about...I'm not sure what. But instead should be a concise overview of what the PR is for, what problem it solves.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
dsargent: This feels a bit confusing, since now we have:

1) `sly_data`, the OG
2) `extraSlyData`
3) this new extra-extra `sly_data`. Can this just be added to plain old (2) `extraSlyData`?

swensel:
I think we should reduce the space above and below "Added automatically when...". It kind of looks like it's floating in space in your screenshot:

image

That message might be a good use for [MUI Alert](https://mui.com/material-ui/react-alert/)?

Here's a version from Claude Design that you can use as a reference:

Screenshot 2026-08-03 at 1 58 06 PM

swensel:
Also, "Added automatically when you send: login." reads a bit strange. Can you reword this please? Do you mean when "Added automatically when you press login" or "Added automatically when you initiate login", or something else?

image

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
dsargent: I think this is fine for now. Skipping `sly_data` is correct for legacy agents.

packages/ui-common/components/AgentChat/ChatCommon/Const.ts
dsargent: Not a bad idea. I appreciate the sentiment of [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) and "single point of truth".

packages/ui-common/__tests__/components/AgentChat/ChatCommon/SlyDataDialog.test.tsx
```diff
@@ -0,0 +1,249 @@
...
+    withStrictMocks()
+
+    let user: UserEvent
+    const onCloseMock = vi.fn()
+
+    beforeEach(() => {
+        user = userEvent.setup({delay: null})
+
+        // Start each test from a clean store
+        const agentHistory = useAgentChatHistoryStore.getState().history
```
swensel, line 49: At some point we may want `withStrictMocks` to also reset Zustand stores. I see there's also a [comment](https://github.com/cognizant-ai-lab/neuro-san-ui/blob/main/packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx#L142) regarding that in the code already.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
swensel: I would also prefer if this used (2) `extraSlyData`.

dsargent:
Kudos for checking how NSFlow does it rather than reinventing the wheel ✅

packages/ui-common/__tests__/components/AgentChat/ChatCommon/SlyDataDialog.test.tsx
dsargent: Yes, surprisingly there isn't a simple way built into Zustand to do this. Some suggested approaches [here](https://github.com/pmndrs/zustand/issues/242).

dsargent: And got the magical ticket #500

M-Elsaied:
> First comment: please fix the "Description". It looks like an LLM stream-of-consciousness about...I'm not sure what. But instead should be a concise overview of what the PR is for, what problem it solves.

Agree & Updated

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
M-Elsaied: To make sure I read this right: you'd like login folded into `extraSlyData` so there's no third thing?

M-Elsaied:
> I think we should reduce the space above and below "Added automatically when...". It kind of looks like it's floating in space in your screenshot:
> 
> image
> That message might be a good use for [MUI Alert](https://mui.com/material-ui/react-alert/)?
> 
> Here's a version from Claude Design that you can use as a reference:
> 
> Screenshot 2026-08-03 at 1 58 06 PM

Let me know if this looks better, simplified the note as well 
image

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
swensel: Yes, that's correct.

swensel:
> > I think we should reduce the space above and below "Added automatically when...". It kind of looks like it's floating in space in your screenshot:
> > image
> > That message might be a good use for [MUI Alert](https://mui.com/material-ui/react-alert/)?
> > Here's a version from Claude Design that you can use as a reference:
> > Screenshot 2026-08-03 at 1 58 06 PM
> 
> Let me know if this looks better, simplified the note as well image

It still seems confusing from a user perspective. Based on looking at [getMergedKeysNote](https://github.com/cognizant-ai-lab/neuro-san-ui/pull/498/changes#diff-ad2df53f7cf98ad98fb8f09727926b2379ce77a3eea025a03d32740c9062eca5R313), it sounds like you want it to say something like "The following keys are included automatically with every request: login. Keys are supplied by the backend."

Also, why do we need to show the merged keys in this way? Why wouldn't they show in the editor above? Is it that the user can edit what's in the editor, but not these keys?

packages/ui-common/components/AgentChat/ChatCommon/SlyDataDialog.tsx
```diff
@@ -0,0 +1,395 @@
...
+            id={id}
+            isOpen={isOpen}
+            onClose={onClose}
+            paperProps={{maxWidth: "800px", width: "80vw"}}
+            title={getTitle()}
+        >
+            {getToolbar()}
+            {getServerUpdateAlert()}
+            <TextField
+                error={Boolean(parseResult.error)}
```
dsargent, line 371: Nice use of the `error` property to display validation errors! Can't believe I didn't know about that. It's a bit hard to read in dark mode, but that could just be due to our (known) theming issues:

image

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
dsargent: Not just login -- It think that's too specific.

I feel like we have this proliferation of sly_data layers -- the base thing, "extra", and now "custom stuff added through the dialog". There has to be a cleaner way to express that.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -1005,6 +1026,18 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
+                <MenuItem
```
dsargent, line 1034: Neat idea to add it to the existing context menu. I fear it may be "undiscoverable" there though -- I wouldn't think to look there for it. It's a network level thing, so maybe a button next to the network name, similar to where we show the "edit" button?

I'm not 100% sure about this. Open to discussion.

image

packages/ui-common/components/AgentChat/ChatCommon/SlyDataDialog.tsx
```diff
@@ -0,0 +1,395 @@
...
+
+    // The text the editor was last seeded with, so we can tell "the server changed sly_data" apart from
+    // "the user changed sly_data"
+    const [seededText, setSeededText] = useState<string>(storedText)
+
+    const fileInputRef = useRef<HTMLInputElement>(null)
+
+    // Seed the editor when the dialog opens, and keep following the store for as long as the user has not typed
+    // anything. Once there are unsaved edits we leave their text alone and offer a reload instead (see below).
+    useEffect(() => {
```
dsargent: This probably doesn't need to be an effect -- we can just update state in the various event handlers.

Very helpful article: https://react.dev/learn/you-might-not-need-an-effect

packages/ui-common/components/AgentChat/ChatCommon/SlyDataDialog.tsx
```diff
@@ -0,0 +1,395 @@
...
+                </Box>
+            </Box>
+        )
+
+    /**
+     * Explains the keys that this dialog deliberately does not show, so that their absence doesn't look like a bug.
+     */
+    const getMergedKeysNote = () =>
+        extraSlyDataKeys.length > 0 && (
+            <Alert
```
dsargent, line 312: We do have a home-grown `MUIAlert` component that wraps `Alert`, but in here you are doing some custom stuff so it may not be a fit. Ditto on line 251.

packages/ui-common/components/AgentChat/ChatCommon/SlyDataDialog.tsx
```diff
@@ -0,0 +1,395 @@
...
+// Avoids a new array literal as a default prop value on every render
+const NO_EXTRA_KEYS: readonly string[] = []
+
+// Height of the JSON editor, in rows. Big enough to see a realistic payload without dominating the screen.
+const EDITOR_ROWS = 14
+
+// Edge length of the image previews, in pixels
+const IMAGE_PREVIEW_SIZE = 96
+
+const SECURITY_WARNING =
```
dsargent: Not sure about this. Firstly, telling an end user in an app popup what not to put in source files...is a bit misplaced. MAUI is for non-technical users: they may not even know what a source file is.

But secondly, the whole point of sly_data is for items that are sensitive and shouldn't be sent to LLMs. For example, if we have a network that uses a SalesForce connector, the user would enter their SalesForce credentials in sly_data. That's exactly what it's for. The popup text sends mixed messages.

dsargent (changes requested):
Good first effort @M-Elsaied. Thanks for implementing this and helping us work through our "fork contributions" CI issues!

I especially appreciate that you followed our local norms about style, tests, using existing utilities like `MUIDialog`.

Some comments and points for discussion. Please address.

M-Elsaied:
> > > I think we should reduce the space above and below "Added automatically when...". It kind of looks like it's floating in space in your screenshot:
> > > image
> > > That message might be a good use for [MUI Alert](https://mui.com/material-ui/react-alert/)?
> > > Here's a version from Claude Design that you can use as a reference:
> > > Screenshot 2026-08-03 at 1 58 06 PM
> > 
> > 
> > Let me know if this looks better, simplified the note as well image
> 
> It still seems confusing from a user perspective. Based on looking at [getMergedKeysNote](https://github.com/cognizant-ai-lab/neuro-san-ui/pull/498/changes#diff-ad2df53f7cf98ad98fb8f09727926b2379ce77a3eea025a03d32740c9062eca5R313), it sounds like you want it to say something like "The following keys are included automatically with every request: login. Keys are supplied by the backend."
> 
> Also, why do we need to show the merged keys in this way? Why wouldn't they show in the editor above? Is it that the user can edit what's in the editor, but not these keys?

@swensel I see what you mean and it is a valid confusion.
Where I was coming from is that are 2 layers in this
1. is the `sly-data-schema` that would be input `sly-data` for the given network (aka the `{}`)
2. `sly-data` that are set up each request (login, BYOK, authorization level, etc)
The mindset I was coming from might be abit tailored to  the downstream production of this PR which I understand might be an overkill for a general purpose implementation.

For this, if you wish, I can simplify to have a simple message, and the getMergedKeyNote I can add in the downstream implementation. More than happy to do wither way, but I hope this makes sense now.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
M-Elsaied: Your typical tradeoff - Here the assumption is that `sly-data` is an advanced feature that would be used for advanced users. As of now, the current implementation of AND doesn't out of the box use `sly-data` in the design phase, so you are bound to only use it in the published networks. This can always get upgraded for more visibility as the implementation grows. We can track it in an issue.

That's my take but I stand to be corrected :)

packages/ui-common/components/AgentChat/ChatCommon/SlyDataDialog.tsx
M-Elsaied: Good catch and thanks for sharing! Deleted the `useEffect` and the `text/isDirty/seededText` states, and moved into the event handlers. Please have a look and let me know if I missed something

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
M-Elsaied: A suggestion to track in issue in case the networks evolve over time and the definition of legacy becomes dynamic (take this with a grain of salt, you guys have higher visibility)

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
M-Elsaied: Seems like an architectural decision, I am sure there are smarter people than me to see through that; I see how folding can be too narrow, I also see a bolder vision of moving `login` into `MAA` so `ChatCommon` receives the complete injection as one prop (which might be an overkill) and probably there's some compromise in the middle. Seems like 2 ends of a spectrum. Overall I don't think this should block the PR - happy to file a follow-up issue so the direction gets decided in one place.

@swensel @dsargent If you agree, I'll hold the login fold for now so we dont have half changes

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
dsargent: > As of now, the current implementation of AND doesn't out of the box use sly-data in the design phase, so you are bound to only use it in the published networks. 

That's not true though? While designing the network, AND sends us "updates" in `sly_data` which get displayed in the popover window. The final network is also sent to the client in `sly_data`. AND is all about `sly_data`, just like any agent that uses reservations (currently the only other one being CopyCat I believe.)

I agree that messing with `sly_data` is an advanced feature. I like the idea of an "advanced..." menu somewhere for more technical users who want to get "under the hood", and editing `sly_data` could be part of that.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
dsargent: We don't want them to evolve in that direction. The idea is to gradually migrate the legacy agents to be Neuro SAN agents, and phase out all legacy agents so we don't have to support two types. We could then remove all the branching logic from MAUI. We have tickets for that but thus far they haven't been top priority.

We do have a use case for rapid development and iteration e.g. FedEx day hackathons, where we may want to temporarily host agents in Next.js rather than Neuro SAN, since deploying agents to Neuro SAN requires more effort (releases of Neuro SAN and Studio, deployments). But for future cases like that, we would make them adhere to the same contract as Neuro SAN agents so as far as MAUI is concerned they _would_ be Neuro SAN agents. Either way, no more legacy agents.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
M-Elsaied: You're spot on, I think I misspoke when I meant it doesn't use them in the design phase, I meant the output networks dont have `sly-data` as a variable for the output network but you are totally correct, it uses it when building the network. Thanks for the correction, sometime my mind speaks but the keyboard types something else :)

packages/ui-common/components/AgentChat/ChatCommon/SlyDataDialog.tsx
M-Elsaied: Agree and addressed - out of curiosity. is the `sly-data` kept in the browser's local storage? like is it persisted in the browser IndexedDB, via the chat history store?

dsargent: We save it in `IndexedDB` (so we can pick up where we left off if the user chooses another network), and I guess also in browser memory while interacting.

swensel:
> @swensel I see what you mean and it is a valid confusion. Where I was coming from is that are 2 layers in this
> 
> 1. is the `sly-data-schema` that would be input `sly-data` for the given network (aka the `{}`)
> 2. `sly-data` that are set up each request (login, BYOK, authorization level, etc)
>    The mindset I was coming from might be abit tailored to  the downstream production of this PR which I understand might be an overkill for a general purpose implementation.
> 
> For this, if you wish, I can simplify to have a simple message, and the getMergedKeyNote I can add in the downstream implementation. More than happy to do wither way, but I hope this makes sense now.

@M-Elsaied I suppose it is OK for now. What do you mean by the downstream implementation? How would you add it there? I might also suggest asking ChatGPT or Claude Design for suggestions. Perhaps what you have for now is fine though, and then you can iterate from there.

packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
```diff
@@ -1594,6 +1594,56 @@ describe("ChatCommon", () => {
...
+            await user.clear(editor)
+            await user.click(editor)
+            await user.paste('{"seeded_by_user": "yes"}')
+            await user.click(screen.getByRole("button", {name: "Save"}))
+
+            await sendQuery(TEST_AGENT_MATH_GUY, "query after seeding sly_data")
+
+            // sly_data is the 7th argument to sendChatQuery. The user's value rides along with the entries the
+            // component always supplies.
+            expect(vi.mocked(sendChatQuery).mock.calls[0][6]).toEqual({
```
dsargent: Rather than low level `.mock.calls` I'd prefer `expect(sendChatQuery).toHaveBeenCalledExactlyOnceWith(...)`

You can use `expect.any` for args you don't care about.

packages/ui-common/__tests__/components/AgentChat/ChatCommon/SlyDataDialog.test.tsx
```diff
@@ -0,0 +1,251 @@
...
+        act(() => useAgentChatHistoryStore.getState().setSlyData(NETWORK_ID, {charges: 2, user: "bob"}))
+
+        renderDialog()
+        await replaceEditorContents('{"charges": 2}')
+        await user.click(screen.getByRole("button", {name: "Save"}))
+
+        expect(getSlyData()).toEqual({charges: 2})
+    })
+
+    it("refuses to save malformed JSON and says why", async () => {
```
dsargent, line 117: Well, this doesn't actually assert the "says why" part.

packages/ui-common/__tests__/components/AgentChat/ChatCommon/SlyDataDialog.test.tsx
```diff
@@ -0,0 +1,251 @@
...
+        renderDialog()
+
+        const file = new File(["not json at all"], "sly.json", {type: "application/json"})
+        await user.upload(screen.getByTestId(`${DIALOG_ID}-file-input`), file)
+
+        expect(getEditor()).toHaveValue("not json at all")
+        expect(screen.getByRole("button", {name: "Save"})).toBeDisabled()
+    })
+
+    it("previews values that look like images", () => {
```
dsargent, line 202: This is a neat feature. You're on a new plane of technology here.

packages/ui-common/__tests__/unit/Utils/SlyData.test.ts
```diff
@@ -0,0 +1,136 @@
...
+    it("parses a JSON object", () => {
+        const result = parseSlyData('{"charges": 2}')
+        expect(result.error).toBeUndefined()
+        expect(result.value).toStrictEqual({charges: 2})
+    })
+
+    it.each(["", " ".repeat(3), "\n"])("treats blank text (%j) as empty sly data", (text: string) => {
+        const result = parseSlyData(text)
+        expect(result.error).toBeUndefined()
+        expect(result.value).toStrictEqual({})
```
dsargent, line 60: TIL about `toStrictEqual`. Now I wish that were the default -- we should only loosen it to `toEqual` when we absolutely have to. @swensel 👀

packages/ui-common/__tests__/unit/Utils/SlyData.test.ts
```diff
@@ -0,0 +1,136 @@
...
+        const result = parseSlyData('{"charges": }')
+        expect(result.error).toBeTruthy()
+        expect(result.value).toStrictEqual({})
+    })
+
+    it("rejects an array at the root", () => {
+        expect(parseSlyData("[1, 2]").error).toBe("Sly data must be a JSON object, not an array.")
+    })
+
+    it.each(["42", '"hello"', "true", "null"])("rejects the primitive %s at the root", (text: string) => {
```
dsargent, line 73: Kudos for use of `42` 🫡

packages/ui-common/utils/SlyData.ts
```diff
@@ -0,0 +1,159 @@
...
+
+    if (Array.isArray(parsed)) {
+        return {error: "Sly data must be a JSON object, not an array.", value: {}}
+    }
+
+    if (parsed === null || typeof parsed !== "object") {
+        return {error: "Sly data must be a JSON object, not a primitive value.", value: {}}
+    }
+
+    return {value: parsed as Record<string, unknown>}
```
dsargent: Future:
1) We already have a type for `SlyData` -- it is exactly `Record<string, unknown>`.
2) Perhaps more radically: we could use Zod for our various `sly_data` handling cases. We already have it as a dep.

The nice thing about Zod is that it is fully TypeScript-enabled. Once you've ascertained that your "mystery data" is a Foo, which is done via runtime checks by Zod, you can then actually treat your data as Foo in a type-safe manner.

Roughly (pseudo-code)

```typescript
const foo = zod.checkIfMyDataIsAFoo(blobThing)
if (foo) {
  foo.someFunctionOnFoo()
}
```

without the `as Foo` ugliness.

dsargent (dismissed):
I think this is good for an MVP. Thanks for implementing the feature and for incorporating our feedback, @M-Elsaied.

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
dsargent: I would indeed prefer if `ChatCommon` were "dumber" and didn't have to know about `sly_data`, other than as an opaque prop that gets passed back and forth. I'm already pondering a follow-up PR to clean up the `sly_data` handling, something around "layers", a session-level structure that gets added to as we go but without the whole "sly_data...extra sly_data...extra extra sly_data" confusion. I haven't solidified the design yet. We already have _something_ like that -- the `sly_data` zustand store, but it's not as clean as I'd like. We probably need a better abstraction.

dsargent:
> > @swensel I see what you mean and it is a valid confusion. Where I was coming from is that are 2 layers in this
> > 
> > 1. is the `sly-data-schema` that would be input `sly-data` for the given network (aka the `{}`)
> > 2. `sly-data` that are set up each request (login, BYOK, authorization level, etc)
> >    The mindset I was coming from might be abit tailored to  the downstream production of this PR which I understand might be an overkill for a general purpose implementation.
> > 
> > For this, if you wish, I can simplify to have a simple message, and the getMergedKeyNote I can add in the downstream implementation. More than happy to do wither way, but I hope this makes sense now.
> 
> @M-Elsaied I suppose it is OK for now. What do you mean by the downstream implementation? How would you add it there? I might also suggest asking ChatGPT or Claude Design for suggestions. Perhaps what you have for now is fine though, and then you can iterate from there.

I had another more radical idea for this. Doesn't have to be for this PR. I will mock up a render when I get a chance.

Something like: use the existing text box, but pre-populate the sly_data items that we always send, but "lock" them (uneditable), read-only, and visual differentiated, like grayed out or slightly different font. The user would "type around" the "unmovable objects" with their own items that they wish to be sent. It's hard to describe in words so I will mock it later.

This may be overkill for our situation though. The only item we have, or plan to have, of "always send" `sly_data` is...the user_id. And we _shouldn't_ be sending that: it's due to our messed up non-security system. So this whole thing may be moot. We may be over-designing for a "feature" that is going to go away and never be used again.

packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
M-Elsaied: Thanks for the recommendation. Updated

packages/ui-common/__tests__/components/AgentChat/ChatCommon/SlyDataDialog.test.tsx
M-Elsaied: Good catch - Now it does! Added `expect(getEditor()).toHaveAccessibleDescription(/json/iu)`

packages/ui-common/utils/SlyData.ts
M-Elsaied: Agree on both. (1) was cheap enough to just do, so I did.

M-Elsaied:
@dsargent Thanks for the approval. There were some cleanup and cheap fixes that I'd rather it being addressed here as it would be easier, so I did. Approval is stale, if you can kindly re-approve. and hopefully by the time #507 is merged so we can see if there is any follow up needed to pass the CI

packages/ui-common/components/AgentChat/ChatCommon/SlyDataDialog.tsx
```diff
@@ -0,0 +1,390 @@
...
+
+// Height of the JSON editor, in rows. Big enough to see a realistic payload without dominating the screen.
+const EDITOR_ROWS = 14
+
+// Edge length of the image previews, in pixels
+const IMAGE_PREVIEW_SIZE = 96
+
+// Tooltip explaining what sly data is, in end-user terms: it exists to carry values (credentials included) that
+// belong with the request but must stay out of the LLM conversation.
+const SLY_DATA_EXPLAINER =
```
dsargent, line 72: This is better. Though there is an aspect of, "if the user doesn't know what sly_data is and how it's used, they shouldn't be messing with it." As we've agreed, it is an advanced feature.

dsargent (dismissed):
Re-approving -- thanks for the updates.

Please take a look at Copilot's comments and if you agree, fix them, if not, dismiss them with an explanation.

packages/ui-common/components/AgentChat/ChatCommon/SlyDataDialog.tsx
```diff
@@ -283,6 +283,8 @@ export const SlyDataDialog: FC<SlyDataDialogProps> = ({
                         >
                             <img
                                 alt={image.path}
+                                loading="lazy"
+                                referrerPolicy="no-referrer"
```
M-Elsaied, line 287: Incorporating copilot suggestion for hardening `sly-data` image previews

packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
```diff
@@ -1595,6 +1595,36 @@ describe("ChatCommon", () => {
             expect(sendChatQuery).toHaveBeenCalledTimes(1)
         })
 
+        it("Should not persist echoed BYOK keys into the sly_data store", async () => {
```
M-Elsaied, line 1598: Covers the `llm_config` strip in `handleNeuroSanAgentChunk`

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -451,9 +451,12 @@ export const ChatCommon = ({ref, ...props}: ChatCommonProps & {ref?: Ref<ChatCom
...
             }
 
-            // Shallow merge existing slyData with incoming chatMessage.sly_data
+            // Shallow merge existing slyData with incoming chatMessage.sly_data.
+            // Defense in depth: never let BYOK API keys enter the persistent store, even if a network echoes
+            // them back -- the store is displayed and exportable via the sly_data editor.
             if (chatMessage.sly_data) {
-                updateSlyData(selectedNetwork, chatMessage.sly_data)
+                const {llm_config: _, ...echoedSlyData} = chatMessage.sly_data
+                updateSlyData(selectedNetwork, echoedSlyData)
```
M-Elsaied, line 459: Addresses Copilot's comment for echoing the `llm_config` upstream for BYOK

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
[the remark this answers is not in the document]
M-Elsaied: This function is likely to change or disappear with the sly_data-layers follow-up (see thread above) - deferring the dedup with it rather than patching a line that's about to move.

packages/ui-common/state/ChatHistory.ts
```diff
@@ -8,9 +8,7 @@ import {persist, PersistStorage, StorageValue} from "zustand/middleware"
 
 import {indexedDBStorage} from "./IndexedDBStorage"
 import {ChatContext} from "../generated/neuro-san/NeuroSanClient"
-
-// Define a type to represent sly_data, which is super loose and can be almost anything depending on the agent.
-type SlyData = Record<string, unknown>
+import {SlyData} from "../utils/SlyData"
```
M-Elsaied, line 11: Second half of @dsargent  type suggestion - the alias moved to `utils/SlyData.ts`

packages/ui-common/state/ChatHistory.ts
```diff
@@ -130,6 +139,12 @@ export const useAgentChatHistoryStore = create<ChatHistoryStore>()(
                     const newHistory = {...state.history, [agentId]: {...existing, slyData: mergedSlyData}}
                     return {history: newHistory}
                 }),
+            setSlyData: (agentId: string, slyData: SlyData) =>
+                set((state) => {
+                    const existing = state.history[agentId]
+                    const newHistory = {...state.history, [agentId]: {...existing, slyData}}
+                    return {history: newHistory}
+                }),
```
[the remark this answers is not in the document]
M-Elsaied: Fixed - `setSlyData` now initializes fresh entries with `ChatCommon`'s own defaults `(chatHistory: [], chatContext: null)`

packages/ui-common/utils/SlyData.ts
```diff
@@ -21,6 +21,12 @@ limitations under the License.
  * but coded tools read from and write to it. It is "super loose": any JSON object the agent network cares to define.
  */
 
+/**
+ * A sly_data payload. Deliberately loose: the shape is defined by each agent network, so the UI can only assume
+ * "a JSON object".
+ */
+export type SlyData = Record<string, unknown>
```
M-Elsaied, line 28: The type lives here now and `ChatHistory` imports it, so state → utils stays the dependency direction.

packages/ui-common/components/AgentChat/ChatCommon/SlyDataDialog.tsx
```diff
@@ -0,0 +1,390 @@
...
+    networkId,
+    networkDisplayName,
+    onClose,
+}) => {
+    const storedSlyData = useAgentChatHistoryStore((state) =>
+        networkId ? state.history[networkId]?.slyData : undefined
+    )
+    const setSlyData = useAgentChatHistoryStore((state) => state.setSlyData)
+
+    const storedText = useMemo(() => formatSlyData(storedSlyData), [storedSlyData])
```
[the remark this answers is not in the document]
M-Elsaied: Addressed

dsargent (approved):

PR #499, opened 2026-08-03
Enable fork pull request CI runs

donn-leaf:
## Summary

Adds a `pull_request` trigger with `synchronize` to the Orchestrator workflow and a concurrency group so fork PRs queue workflow runs that maintainers can approve, while internal branch `push` and `pull_request` events for the same commit cancel instead of duplicating. 

Note: back in the old days, before public repos and forked repos, we removed the `sychronize` trigger to avoid duplicated action triggers for push/pull requests. Now we need the `synchronize` trigger to support forks (as they don't emit `push`) . So now that we've re-added a `synchronize` trigger with a concurrency group it means we will get two action _triggers_ but only one of those will _run_. The other will show up in the Actions list as cancelled. That's the best behavior we could come up with.  neuro-san repo adopts this behavior as well.

Updates `determine_version.sh` to handle `pull_request` events as test-only, matching the recent neuro-san fix.

Link to Devin session: https://app.devin.ai/sessions/de18e5fdd36d451b849904545ca02ab7
Requested by: @donn-leaf

---

<a href="https://app.devin.ai/review/cognizant-ai-lab/neuro-san-ui/pull/499" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.github/workflows/orchestrator.yml
+++ b/.github/workflows/orchestrator.yml
@@ -20,6 +20,7 @@
 # Trigger Scenarios:
 # - Push to any branch: Runs tests only (for development feedback)
 # - Push to main branch: Runs test → build → deploy to dev
+# - Pull request to main: Runs tests only; supports fork PRs via workflow approval
 # - GitHub release: Runs test → build → deploy to staging → bump package.json
 # ================================================================================
 
@@ -32,9 +33,18 @@ on:
     # Trigger on pushes to any branch
     push:
         branches: ["**"]
+    # Trigger on pull requests targeting main, including updates from forks
+    pull_request:
+        types: [opened, synchronize, reopened]
+        branches:
+            - main
     # Trigger on GitHub release events
     release:
         types: [published]
+
+concurrency:
+    group: ${{ github.workflow }}-${{ github.head_ref || github.ref_name }}-${{ github.event.pull_request.head.sha || github.sha }}
+    cancel-in-progress: true
 
 permissions:
     id-token: write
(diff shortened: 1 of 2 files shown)


.github/workflows/orchestrator.yml
```diff
@@ -46,8 +46,8 @@ concurrency:
     group: ${{ github.workflow }}-${{ github.head_ref || github.ref_name }}-${{ github.event.pull_request.head.sha || github.sha }}
     cancel-in-progress: true
 
+# Minimal default permissions; jobs that need extra permissions override below.
 permissions:
-    id-token: write
```
donn-leaf, line 50: Aside: Here we move the write permissions from the top level down to the build job which needs that permission.

.github/workflows/orchestrator.yml
```diff
@@ -32,12 +33,21 @@ on:
     # Trigger on pushes to any branch
     push:
         branches: ["**"]
+    # Trigger on pull requests targeting main, including updates from forks
+    pull_request:
+        types: [opened, synchronize, reopened]
+        branches:
+            - main
```
dsargent, line 40: I hope this still works even though PRs from forks show as targeting not `main` but `cognizant-ai-lab:main`.

.github/workflows/orchestrator.yml
dsargent: I'm all about minimum permissions and minimally _scoped_ permissions (as here).

dsargent (approved):
I'm not knowledgeable enough to comment on the correctness, but nothing leaps out at me.

.github/workflows/orchestrator.yml
dsargent: >  The pull_request run on this commit (run 1352) confirmed it works.

Not true: this PR is not from a _fork_ (which is the whole point of the PR) so this PR doesn't confirm it works.

PR #501, opened 2026-08-04
Swap concurrency for fork check in orchestrator

donn-leaf:
Removes the `concurrency` block in `orchestrator.yml` and adds a fork check to the `test` job's `if` condition, matching the style of `neuro-san-studio/.github/workflows/tests.yml`. Internal PRs are already covered by the `push` event, so the `pull_request` event is skipped for them while still running for fork PRs. This avoids cancelled workflow runs that appear as failures in the PR checks UI. Branch protection's required `test` check should be satisfied by the `push` run for internal PRs and by the `pull_request` run for forks.

Also, I did look at moving some of this logic down to build-common. There's probably some benefit there, but I won't do so until all consumers are using and happy with this trigger logic.

Link to Devin session: https://app.devin.ai/sessions/635e68f78ccb4fb7b1b94a8dea614de5
Requested by: @donn-leaf

---

<a href="https://app.devin.ai/review/cognizant-ai-lab/neuro-san-ui/pull/501" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>


.github/workflows/orchestrator.yml
```diff
@@ -74,12 +70,13 @@ jobs:
...
                     "${{ github.sha }}"
 
-    # Step 1: Always run tests on pushes, releases, and pull requests targeting main
+    # Step 1: Run tests on pushes, releases, and fork pull requests
     test:
         if: >
             github.event_name == 'push' ||
             github.event_name == 'release' ||
-            (github.event_name == 'pull_request' && github.base_ref == 'main')
+            (github.event_name == 'pull_request' &&
```
dsargent, line 78: Does this mean, "don't run the tests on pull requests that originate in the repository, only from forks"? That doesn't sound right -- we want to run CI on those too. Or is that case captured by `push`?

donn-leaf: That's correct, the push covers that...when we specify both push and pull, we get double runs, and we've chased our tails on that particular issue.

dsargent (approved):

PR #507, opened 2026-08-07
Use build-common setup-node-env to resolve Node.js version

donn-leaf:
## Summary

Fork `pull_request` workflows cannot read repository variables, so the local `resolve_nodejs_version.sh` helper was added to derive the Node.js major version from `package.json`. This now lives in `build-common/actions/setup-node-env` 1.0.9, which reads `engines.node` and exposes `nodejs-major-version` for Docker build-args. The neuro-san-ui test/build/publish workflows use the released build-common 1.0.9 SHA and pass the major version to the Docker build check and docker-buildx-push.

Link to Devin session: https://app.devin.ai/sessions/31fcc700b0a141c9b8d749b6b574c391
Requested by: @donn-leaf


dsargent (approved):

dsargent:
In the "additional comments" Devin does point out that we are now mixing build-common versions within a single Action.

donn-leaf:
> In the "additional comments" Devin does point out that we are now mixing build-common versions within a single Action.

That's a good call out. I'm going to merge this for the sake of forward motion, and fix this issue in a follow up PR.

Issue #509, opened 2026-08-11
"Flickering" unit test


dsargent:
Test coverage fluctuates by ±1 branch due to random behavior in `PlasmaEdge.tsx`.

PR #510, opened 2026-08-11
#509 Fix "flickering" unit test

dsargent:
Fixes the `PlasmaEdge` test which was flickering ±1 branch coverage due to use of `Math.random()`.
Also cleans up the hardcore casting, in favor of using Vitest spies, obviating the need for the `finally` block with all its mock clean-up logic: `withStrictMocks` does that for us.

With this fix, the branch coverage count is rock solid. I ran the test over and over to be sure.

No Copilot because it's just test code, it's short and sweet-ish, and tokens are 'spensive.

### DIFF
--- a/packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/PlasmaEdge.test.tsx
+++ b/packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/PlasmaEdge.test.tsx
@@ -39,8 +39,8 @@ describe("PlasmaEdge", () => {
             />
         )
 
-    it("renders and runs animation with mocked canvas context, SVG methods, and RAF", () => {
-        const errSpy = vi.spyOn(console, "error").mockImplementation(vi.fn())
+    it("renders and runs animation", () => {
+        vi.spyOn(console, "error").mockImplementation(vi.fn())
 
         // Mock getContext to provide minimal API used by the component
         const fakeCtx: Partial<CanvasRenderingContext2D> = {
@@ -53,6 +53,13 @@ describe("PlasmaEdge", () => {
             save: vi.fn(),
             restore: vi.fn(),
         }
+
+        const randomSpy = vi.spyOn(Math, "random")
+        // Mock successive random values to cover each of the branches
+        randomSpy
+            .mockReturnValue(0) // default: create particles deterministically
+            .mockReturnValueOnce(0.9) // first render: t = 0.9
+            .mockReturnValueOnce(0.9) // first render: 0.9 < 0.1 => false, covers inner else
 
         const origGetContext = HTMLCanvasElement.prototype.getContext
         const origRAF = global.requestAnimationFrame
@@ -115,7 +122,6 @@ describe("PlasmaEdge", () => {
             }
             elementPrototype.getTotalLength = origGetTotalLength
             elementPrototype.getPointAtLength = origGetPointAtLength
-            errSpy.mockRestore()
         }
     })
 })
--- a/packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/PlasmaEdge.test.tsx
+++ b/packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/PlasmaEdge.test.tsx
@@ -20,30 +20,53 @@ import {Position} from "@xyflow/react"
 import {withStrictMocks} from "../../../../../../__tests__/common/strictMocks"
 import {PlasmaEdge, PlasmaEdgeProps} from "../../../../components/MultiAgentAccelerator/AgentFlow/PlasmaEdge"
 
+// Add shim for function not defined by JSDom
+declare global {
+    interface SVGElement {
+        getTotalLength: () => number
+        getPointAtLength: (distance: number) => {x: number; y: number}
+    }
+}
+
+Object.defineProperties(SVGElement.prototype, {
+    getTotalLength: {
+        configurable: true,
+        writable: true,
+        value: () => 100,
+    },
+    getPointAtLength: {
+        configurable: true,
+        writable: true,
+        value: (distance: number) => ({x: distance, y: distance}),
+    },
+})
+
 describe("PlasmaEdge", () => {
     withStrictMocks()
 
     const renderPlasmaEdge = (overrides: Partial<PlasmaEdgeProps> = {}) =>
         render(
-            <PlasmaEdge
-                id="test-edge"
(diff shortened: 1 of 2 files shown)


packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/PlasmaEdge.test.tsx
```diff
@@ -20,30 +20,53 @@ import {Position} from "@xyflow/react"
...
-                target="test-target"
-                sourceX={0}
-                sourceY={0}
-                targetX={200}
-                targetY={120}
-                sourcePosition={Position.Left}
-                targetPosition={Position.Right}
-                {...overrides}
-            />
+            <svg>
```
dsargent, line 49: Copilot explanation of why it should be wrapped in an `<svg>`:

> In this test, PlasmaEdge is rendered directly into Testing Library’s default &lt;div&gt; container
>So the returned <path> may be created as an HTML element, not an SVG element. That means it does not inherit from SVGElement.prototype, so this shim is not found:
SVGElement.prototype.getTotalLength
>Preferred: render it inside an <svg> in the test helper
> Then pathEl should be an SVG element, and your SVGElement.prototype spy should work.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/PlasmaEdge.test.tsx
```diff
@@ -20,30 +20,53 @@ import {Position} from "@xyflow/react"
...
+                    sourcePosition={Position.Left}
+                    targetPosition={Position.Right}
+                    {...overrides}
+                />
+            </svg>
         )
 
-    it("renders and runs animation with mocked canvas context, SVG methods, and RAF", () => {
-        const errSpy = vi.spyOn(console, "error").mockImplementation(vi.fn())
+    it("renders and runs animation", () => {
```
dsargent, line 65: Humanize test name

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/PlasmaEdge.test.tsx
```diff
@@ -52,70 +75,60 @@ describe("PlasmaEdge", () => {
...
+        })
+
+        const randomSpy = vi.spyOn(Math, "random")
+        // Mock successive random values to cover each of the branches
+        randomSpy
+            .mockReturnValue(0) // default: create particles deterministically
+            .mockReturnValueOnce(0.9) // first render: t = 0.9
+            .mockReturnValueOnce(0.9) // first render: 0.9 < 0.1 => false, covers inner else
 
-        const origGetContext = HTMLCanvasElement.prototype.getContext
```
dsargent, line 57: Get rid of all these in favor of Vitest `spyOn` mocks, which conveniently get cleaned up automatically for us by `withStrictMocks`. So no more `finally` block needed.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/PlasmaEdge.test.tsx
```diff
@@ -20,30 +20,53 @@ import {Position} from "@xyflow/react"
...
 
+// Add shim for function not defined by JSDom
+declare global {
+    interface SVGElement {
+        getTotalLength: () => number
+        getPointAtLength: (distance: number) => {x: number; y: number}
+    }
+}
+
+Object.defineProperties(SVGElement.prototype, {
```
dsargent, line 31: This will persist after the test, but I think I'm okay with that. It's plugging a gap in JSDom's 2D canvas support.

It could be in the `vitest.setup.ts` file too, but, since it's only used for this one test, that seems overkill.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/PlasmaEdge.test.tsx
```diff
@@ -52,70 +75,60 @@ describe("PlasmaEdge", () => {
             fill: vi.fn(),
             save: vi.fn(),
             restore: vi.fn(),
-        }
+        })
+
+        const randomSpy = vi.spyOn(Math, "random")
+        // Mock successive random values to cover each of the branches
+        randomSpy
```
dsargent, line 82: This is the actual fix: prevent randomness by mocking `Math.random()` to produce deterministic values.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/PlasmaEdge.test.tsx
```diff
@@ -52,70 +75,60 @@ describe("PlasmaEdge", () => {
...
+            .mockReturnValueOnce(0.9) // first render: 0.9 < 0.1 => false, covers inner else
 
-        const origGetContext = HTMLCanvasElement.prototype.getContext
-        const origRAF = global.requestAnimationFrame
-        const origCAF = global.cancelAnimationFrame
-        const origGetTotalLength = (Element.prototype as unknown as {getTotalLength?: () => number}).getTotalLength
-        const origGetPointAtLength = (
-            Element.prototype as unknown as {getPointAtLength?: (length: number) => {x: number; y: number}}
-        ).getPointAtLength
+        const getContextSpy = vi.spyOn(HTMLCanvasElement.prototype, "getContext")
```
dsargent, line 87: Much cleaner -- use proper Vitest mocks, and added benefit `withStrictMocks` clean them up for us, so no `finally` required.

packages/ui-common/__tests__/components/MultiAgentAccelerator/AgentFlow/PlasmaEdge.test.tsx
```diff
@@ -52,70 +75,60 @@ describe("PlasmaEdge", () => {
...
-                rafCallback?.(16)
-            })
-
-            expect(fakeCtx.clearRect).toHaveBeenCalled()
-            expect(fakeCtx.arc).toHaveBeenCalled()
-
-            unmount()
-            expect(global.cancelAnimationFrame).toHaveBeenCalled()
-        } finally {
-            HTMLCanvasElement.prototype.getContext = origGetContext
```
dsargent, line 109: All of this no longer needed thanks to use of proper mocks.

swensel (approved):
LGTM. Thanks for fixing this.

PR #515, opened 2026-08-16
Fix SettingsDialog test timeouts

### LINKED ISSUE #514: Unit tests are slow
Some tests are doing too much, slow, and fail with timeouts (on my VM at least)

chronically, `applies branding for requested customer` and related `SettingsDialog` tests.

arnabnandy7:
## Description

Fixes #514 by reducing the scope and execution time of the timeout-prone `SettingsDialog` branding test.

## Changes

- Disable artificial `userEvent` delays for the `SettingsDialog` test suite.
- Split the large branding scenario into focused tests for:
  - Applying branding
  - Submitting with Enter
  - Changing logo sources
  - Clearing branding settings
- Reuse a shared branding dialog setup helper.
- Preserve the existing behavioral coverage and assertions.

## Testing

```text
Test Files  1 passed (1)
Tests       32 passed (32)
Duration    22.48s
```

### DIFF
--- a/packages/ui-common/__tests__/components/Settings/SettingsDialog.test.tsx
+++ b/packages/ui-common/__tests__/components/Settings/SettingsDialog.test.tsx
@@ -14,6 +14,12 @@ import {DEFAULT_SETTINGS, getApiKey, LogoSource, useSettingsStore} from "../../.
 // Mock notification system
 vi.mock("../../../components/Common/notification")
 
+// Fading checkmark behavior is covered by FadingCheckmark.test.tsx.
+vi.mock("../../../components/Settings/FadingCheckmark", () => ({
+    FadingCheckmark: (): null => null,
+    useCheckmarkFade: () => ({show: false, trigger: vi.fn()}),
+}))
+
 const TEST_API_KEY = "test-api-key-123"
 
 // Pseudo expiration time
@@ -41,7 +47,7 @@ describe("SettingsDialog", () => {
         useSettingsStore.persist.clearStorage()
         useEnvironmentStore.getState().setBackendNeuroSanApiUrl(undefined)
 
-        user = userEvent.setup()
+        user = userEvent.setup({delay: null})
         originalFetch = global.fetch
     })
 
@@ -59,6 +65,10 @@ describe("SettingsDialog", () => {
             const applyButton = screen.getByRole("button", {name: /Apply/u})
             await user.click(applyButton)
         }
+    }
+
+    const mockBrandingResult = () => {
+        global.fetch = mockFetch(BRANDING_SUGGESTIONS_RESPONSE)
     }
 
     it("renders the SettingsDialog with default props", async () => {
@@ -703,8 +713,7 @@ describe("SettingsDialog", () => {
     })
 
     it("applies branding for requested customer", async () => {
-        global.fetch = mockFetch(BRANDING_SUGGESTIONS_RESPONSE)
-
+        mockBrandingResult()
         render(
             <SettingsDialog
                 id="settings-dialog"
@@ -714,8 +723,6 @@ describe("SettingsDialog", () => {
         )
 
         expect(useSettingsStore.getState().settings.branding.logoSource).toEqual<LogoSource>("none")
-
-        const logoOptionsContainer = screen.getByRole("group", {name: "logo-selection"}).closest("span").closest("div")
 
         const customerName = "Acme"
         await enterCustomerName(customerName)
@@ -733,19 +740,41 @@ describe("SettingsDialog", () => {
         expect(brandingSettings.background).toBe(BRANDING_SUGGESTIONS_RESPONSE.background)
         expect(brandingSettings.rangePalette).toEqual(BRANDING_SUGGESTIONS_RESPONSE.rangePalette)
         expect(brandingSettings.iconSuggestion).toBe(BRANDING_SUGGESTIONS_RESPONSE.iconSuggestion)
+        expect(brandingSettings.logoSource).toBe("auto")
+    })
 
-        // Now try using Enter to submit a new customer name and check that it also applies branding
-        const newCustomerName = "Acme 2"
-        await enterCustomerName(newCustomerName, false)
+    it("applies branding when Enter is pressed", async () => {
+        mockBrandingResult()
+        render(
+            <SettingsDialog
+                id="settings-dialog"
+                isOpen={true}
+                logoServiceToken="test-logo-service-token-456"
+            />
+        )
+
+        const customerName = "Acme"
+        await enterCustomerName(customerName, false)
         await user.keyboard("{Enter}")
 
-        // Check that the store was updated with the new customer name
-        expect(useSettingsStore.getState().settings.branding.customer).toBe(newCustomerName)
-
-        // Successfully retrieving a customer name should update the logo source to "auto" to show the new logo
+        expect(useSettingsStore.getState().settings.branding.customer).toBe(customerName)
         expect(useSettingsStore.getState().settings.branding.logoSource).toBe("auto")
+    })
 
-        // Try different logo options
+    it("changes the logo source for branded customers", async () => {
+        mockBrandingResult()
+        render(
+            <SettingsDialog
+                id="settings-dialog"
+                isOpen={true}
+                logoServiceToken="test-logo-service-token-456"
+            />
+        )
+
+        const customerName = "Acme"
+        await enterCustomerName(customerName)
+        const logoOptionsContainer = screen.getByRole("group", {name: "logo-selection"}).closest("span").closest("div")
+
         const logoNoneButton = within(logoOptionsContainer).getByRole("button", {name: "None"})
         await user.click(logoNoneButton)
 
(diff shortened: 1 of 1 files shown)


arnabnandy7:
@dsargent to review

packages/ui-common/__tests__/components/Settings/SettingsDialog.test.tsx
dsargent, line 1: Please explain what your fix does and why you believe it fixes the issue.

Also looks like some comments were removed by Claude/whatever was used.

arnabnandy7: @dsargent the original test was doing several things in one flow: applying branding, submitting with Enter, switching between logo options, and clearing branding. On slower machines, all those interactions added up and could cause the test to time out.

I split those behaviors into smaller, focused tests so each one does less work. I also configured `userEvent` with `delay: null` to avoid scheduling delays between simulated interactions.

The comments I removed were things like “now clear branding” and “try different logo options.” Once the test was split, the individual test names already explained those steps, so the comments were redundant. No assertions or behavioral coverage were removed.

Also, I've amended the commit message to maintain the strategy to include issue number by prefixing it.

packages/ui-common/__tests__/components/Settings/SettingsDialog.test.tsx
```diff
@@ -61,6 +61,18 @@ describe("SettingsDialog", () => {
         }
     }
 
+    const renderBrandingDialog = () => {
```
dsargent: This is misnamed. There is no separate "branding dialog": what it is doing is rendering the settings dialog, with a side effect of mocking the fetch result for the branding query.

arnabnandy7: @dsargent renaming it to `renderSettingsDialogWithBrandingSuggestions` should be better what do you think?

dsargent: The problem is that it's doing two different things, so there's a tendency to want to use a name with "and" or "with". `renderSettingsDialogWithBrandingSuggestions` just makes the issue worse (and is comically long).

I would either do the mocking in each test, or if you want to avoid duplication, encapsulate the mocking in a helper function: `mockBrandingResult()` or some such. Or even have a separate mini suite for branding tests and put the mocking in the `beforeEach`.

Having a function like `renderBrandingDialog()` with an unexpected, hidden side effect is a bad practice in general. Another developer could use that function and then waste time scratching their head because it did something [unexpected](https://en.wikipedia.org/wiki/Principle_of_least_astonishment?useskin=vector), namely a "bonus" side-effect.

dsargent:
I measured the performance before & after -- very slight difference, ~ 1 sec. I used Linux `perf`, average over 5 runs of the `SettingsDialog` test suite:

```console
main

 Performance counter stats for 'yarn test packages/ui-common/__tests__/components/Settings/SettingsDialog.test.tsx' (5 runs):

          9,714.76 msec task-clock                       #    1.250 CPUs utilized               ( +-  2.80% )
            13,151      context-switches                 #    1.354 K/sec                       ( +-  6.53% )
             1,467      cpu-migrations                   #  151.007 /sec                        ( +-  4.65% )
           169,778      page-faults                      #   17.476 K/sec                       ( +-  0.21% )
   <not supported>      cycles                                                                
   <not supported>      instructions                                                          
   <not supported>      branches                                                              
   <not supported>      branch-misses                                                         

             7.772 +- 0.285 seconds time elapsed  ( +-  3.67% )


branch

 Performance counter stats for 'yarn test packages/ui-common/__tests__/components/Settings/SettingsDialog.test.tsx' (5 runs):

          8,825.88 msec task-clock                       #    1.315 CPUs utilized               ( +-  2.48% )
            11,565      context-switches                 #    1.310 K/sec                       ( +-  0.70% )
             1,328      cpu-migrations                   #  150.467 /sec                        ( +-  1.26% )
           174,101      page-faults                      #   19.726 K/sec                       ( +-  0.77% )
   <not supported>      cycles                                                                
   <not supported>      instructions                                                          
   <not supported>      branches                                                              
   <not supported>      branch-misses                                                         

             6.713 +- 0.147 seconds time elapsed  ( +-  2.20% )
```

That said, the "branding" test was indeed monolithic and was running afoul of the default 5 second timeout (especially on my VM dev box) so I'm in favor of this PR if for no other reason than that -- to have more granular, faster tests, even if the overall execution time isn't affected.

arnabnandy7:
@dsargent let me check regarding performance and see if I can reduce it further someway or another.

dsargent:
> @dsargent let me check regarding performance and see if I can reduce it further someway or another.

One thing that Copilot suggested at one point, which I never finished trying, was to use fake timers (literally `vi.useFakeTimers()`) so that the tests don't wait for the checkmark to fade in. I don't know if that's a real thing (is the test really quietly waiting for the checkmark to appear after _n_ milliseconds?) or just a hallucination? But seems reasonable. That would shave off some time if true.

packages/ui-common/__tests__/components/Settings/SettingsDialog.test.tsx
arnabnandy7: Replaced `renderBrandingDialog()` with the single-purpose `mockBrandingResult()`. Each test now renders `SettingsDialog` explicitly, eliminating the unexpected side effect.

arnabnandy7:
> > @dsargent let me check regarding performance and see if I can reduce it further someway or another.
> 
> One thing that Copilot suggested at one point, which I never finished trying, was to use fake timers (literally `vi.useFakeTimers()`) so that the test don't wait for the checkmark to fade in. I don't know if that's a real thing (is the test really quietly waiting for the checkmark to appear after _n_ milliseconds?) or just a hallucination? But seems reasonable. That would shave off some time if true.

ah thanks for the heads up let me check that, for now I'm mitigated the smaller issue. I'll take up the performance R&D later tomorrow. Thanks for your suggestions and review

dsargent:
Note that this PR is fine already -- it already does something valuable, namely, breaking up that monolithic test into smaller, less-likely-to-time-out tests. If you want, the optimization can be a separate follow-up. I'm fine either way.

arnabnandy7:
> Note that this PR is fine already -- it already does something valuable, namely, breaking up that monolithic test into smaller, less-likely-to-time-out tests. If you want, the optimization can be a separate follow-up. I'm fine either way.

Let me evaluate the change first, I'll notify you if a separate PR is required.

arnabnandy7:
@dsargent I looked into the fake-timer suggestion first, but enabling fake timers for the whole suite interfered with the dialog’s asynchronous behavior and caused the test run to time out after 124 seconds.

Instead, I mocked the fading checkmark in the `SettingsDialog` tests. Its behavior is already covered separately in `FadingCheckmark.test.tsx`, and none of the dialog tests assert against the checkmark itself. This avoids creating unnecessary fade timers and rendering checkmark components without changing production behavior or reducing test coverage.

I compared the results over five runs:

- Before: 33.67 seconds average
- After: 25.70 seconds average
- Improvement: about 23.7%

The `SettingsDialog` and `FadingCheckmark` tests still pass, with 38 tests passing in total. 

PS: It's in windows, please try testing yourselves to check the benchmark before merging. You may suggest different approach as well.

dsargent:
> @dsargent I looked into the fake-timer suggestion first, but enabling fake timers for the whole suite interfered with the dialog’s asynchronous behavior and caused the test run to time out after 124 seconds.

Yes, this is a known issue with fake timers. You have to use a custom `UserEvent` that is bound to the fake timer, like we do in other tests, or else things like `await user.click()` will just hang, because as far as they're concerned, time is not passing.

I'll take a look in a bit at your changes.

dsargent:
@arnabnandy7 you have a test failure -- please see the CI results.

Note: you can run `yarn commit-check` locally to do 90% of what CI does, before pushing.

dsargent:
> Yes, this is a known issue with fake timers. You have to use a custom `UserEvent` that is bound to the fake timer, like we do in other tests, or else things like `await user.click()` will just hang, because as far as they're concerned, time is not passing.

If you want to use fake timers, see how it's done [here](https://github.com/cognizant-ai-lab/neuro-san-ui/blob/5810c11e6c4479585f5f4027f786e186b7b0200d/packages/ui-common/__tests__/components/MultiAgentAccelerator/Editor/NetworkEditorDock.test.tsx#L322).

arnabnandy7:
@dsargent I tried the same fake-timer pattern used in `NetworkEditorDock.test.tsx`, with `userEvent` connected to Vitest’s timer advancement:

```typescript
vi.useFakeTimers()

const user = userEvent.setup({
    advanceTimers: vi.advanceTimersByTime,
})
```

That fixes the immediate `user.click()` issue, but applying fake timers to the entire `SettingsDialog` suite still caused several tests to stall. The dialog has other asynchronous behavior and Testing Library waits that depend on real timer progression, so this would require advancing timers at multiple test-specific points. As a suite-wide optimization, it ended up making the tests less predictable.

So kept the narrower checkmark mock instead. The fade timing is already tested independently in `FadingCheckmark.test.tsx`, while none of the `SettingsDialog` tests assert against the checkmark animation. This avoids creating those timers without changing production behavior or weakening the relevant coverage.

I also fixed the CI TypeScript error.

dsargent:
> @dsargent I tried the same fake-timer pattern used in `NetworkEditorDock.test.tsx`, with `userEvent` connected to Vitest’s timer advancement:

> That fixes the immediate `user.click()` issue, but applying fake timers to the entire `SettingsDialog` suite still caused several tests to stall. 

Yes, you have to be selective. You can't just apply fake timers to the entire suite because then other tests will fail, as you saw, because once you enable fake timers, time only "passes" when you explicitly make it pass, so tests hang.

packages/ui-common/__tests__/components/Settings/SettingsDialog.test.tsx
```diff
@@ -61,6 +67,10 @@ describe("SettingsDialog", () => {
         }
     }
 
+    const mockBrandingResult = () => {
+        global.fetch = mockFetch(BRANDING_SUGGESTIONS_RESPONSE)
```
dsargent, line 71: Probably should reset this to the real `global.fetch` after each test (`afterEach`). Not a huge deal.

dsargent (approved):

packages/ui-common/__tests__/components/Settings/SettingsDialog.test.tsx
arnabnandy7: @dsargent this is already handled by the existing `afterEach`, which restores `global.fetch` to the value captured in `beforeEach`. I double-checked that each test starts with the original fetch implementation.

Issue #513, opened 2026-08-14
Use sly_data schema in input dialog


dsargent:
Follow on from #52 

The current (MVP) implementation allows the user to enter `sly_data` but it's in a generic, unstructured dialog.

<img width="647" height="536" alt="Image" src="https://github.com/user-attachments/assets/bc765b18-aff8-46e5-85ac-840759939627" />

We can do better: networks that require/accept `sly_data` tells us the format using `sly_data` schema. Use that to structure the input in the UI. 

Implementation note: consider whether Zod could help here.

PR #516, opened 2026-08-17
#513 - Add sly_data schema 

M-Elsaied:
Addresses #513 which builds off #498 where we are grabbing the schema from the server and rendering it in the `sly-data` editor; pre-populated schema is done and an explanation of each variable as well.

Tested E2E - `scheme` is populated and updated in each request as per expected behavior of the `up_stream`

image

### DIFF
--- a/packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
+++ b/packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
@@ -1684,8 +1684,9 @@ describe("ChatCommon", () => {
 
             await openSlyDataEditor()
 
-            expect(screen.getByText(/This network expects/u)).toBeInTheDocument()
-            expect(screen.getByText(/\(float, required\)/u)).toBeInTheDocument()
+            // getByText throws when the element is missing, so these are the assertions
+            screen.getByText(/This network expects/u)
+            screen.getByText(/\(float, required\)/u)
         })
 
         it("Should not offer the sly_data editor for legacy agents, which have no sly_data", async () => {
--- a/packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
+++ b/packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
@@ -1673,6 +1673,21 @@ describe("ChatCommon", () => {
             expect(await openSlyDataEditor()).toHaveValue('{\n  "charges": 7\n}')
         })
 
+        it("Should show the network's schema hints in the sly_data editor", async () => {
+            renderChatCommonComponent({
+                slyDataSchema: {
+                    properties: {x: {description: "The first operand", type: "float"}},
+                    required: ["x"],
+                    type: "object",
+                },
+            })
+
+            await openSlyDataEditor()
+
+            expect(screen.getByText(/This network expects/u)).toBeInTheDocument()
+            expect(screen.getByText(/\(float, required\)/u)).toBeInTheDocument()
+        })
+
         it("Should not offer the sly_data editor for legacy agents, which have no sly_data", async () => {
             renderChatCommonComponent({selectedNetwork: LegacyAgentType.DataGenerator})
 
--- a/packages/ui-common/__tests__/components/AgentChat/ChatCommon/SlyDataDialog.test.tsx
+++ b/packages/ui-common/__tests__/components/AgentChat/ChatCommon/SlyDataDialog.test.tsx
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import {act, render, screen} from "@testing-library/react"
+import {act, render, screen, within} from "@testing-library/react"
 import {userEvent, UserEvent} from "@testing-library/user-event"
 
 import {withStrictMocks} from "../../../../../../__tests__/common/strictMocks"
@@ -36,6 +36,29 @@ const DIALOG_ID = "test-sly-data-dialog"
 // A one-pixel transparent GIF, as a data URI
 const IMAGE_DATA_URI = "data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7"
 
+// The shape of neuro-san's math_guy example schema: two required scalars
+const MATH_GUY_SCHEMA = {
+    properties: {
+        x: {description: "The first operand", type: "float"},
+        y: {description: "The second operand", type: "float"},
+    },
+    required: ["x", "y"],
+    type: "object",
+}
+
+// The shape of a BYOK network's schema: one scalar plus the app-managed llm_config block
+const BYOK_SCHEMA = {
+    properties: {
+        llm_config: {
+            properties: {openai_api_key: {type: "string"}},
+            type: "object",
+        },
+        running_cost: {description: "Fake billing so far", type: "float"},
+    },
+    required: ["llm_config"],
+    type: "object",
+}
+
 describe("SlyDataDialog", () => {
     withStrictMocks()
 
@@ -52,7 +75,7 @@ describe("SlyDataDialog", () => {
         })
     })
 
-    const renderDialog = (extraSlyDataKeys?: readonly string[]) =>
+    const renderDialog = (extraSlyDataKeys?: readonly string[], slyDataSchema?: Record<string, unknown>) =>
         render(
             <SlyDataDialog
                 extraSlyDataKeys={extraSlyDataKeys}
@@ -61,6 +84,7 @@ describe("SlyDataDialog", () => {
                 networkDisplayName="Music Nerd Pro"
                 networkId={NETWORK_ID}
                 onClose={onCloseMock}
+                slyDataSchema={slyDataSchema}
             />
         )
 
@@ -250,5 +274,75 @@ describe("SlyDataDialog", () => {
         renderDialog()
 
         expect(screen.queryByText(/Sent automatically with every request/u)).not.toBeInTheDocument()
+    })
+
+    it("shows no schema hints and no template button when the network declares no schema", () => {
+        renderDialog()
+
+        expect(screen.queryByText(/This network expects/u)).not.toBeInTheDocument()
+        expect(screen.queryByRole("button", {name: "Fill sly data template"})).not.toBeInTheDocument()
+    })
+
+    it("ignores an unusable schema rather than breaking the dialog", () => {
+        vi.spyOn(console, "warn").mockImplementation(() => undefined)
+
+        renderDialog(undefined, {type: "object"})
+
+        expect(screen.queryByText(/This network expects/u)).not.toBeInTheDocument()
+        expect(getEditor()).toHaveValue("{}")
+    })
+
+    it("lists the keys the network expects, with type, requiredness and description", () => {
+        renderDialog(undefined, MATH_GUY_SCHEMA)
+
+        const hints = within(screen.getByText(/This network expects/u).parentElement)
+        expect(hints.getByText("x")).toBeInTheDocument()
+        expect(hints.getByText("y")).toBeInTheDocument()
+        expect(hints.getAllByText(/\(float, required\)/u)).toHaveLength(2)
+        expect(hints.getByText(/The first operand/u)).toBeInTheDocument()
+    })
+
+    it("leaves the app-supplied keys out of the schema hints", () => {
+        renderDialog(["llm_config"], BYOK_SCHEMA)
+
+        const hints = within(screen.getByText(/This network expects/u).parentElement)
+        expect(hints.getByText("running_cost")).toBeInTheDocument()
+        expect(hints.queryByText("llm_config")).not.toBeInTheDocument()
+    })
+
+    it("renders a bare key when the schema gives no type, requiredness or description", () => {
+        renderDialog(undefined, {properties: {plain: {}}, type: "object"})
+
+        const hints = within(screen.getByText(/This network expects/u).parentElement)
+        expect(hints.getByText("plain")).toBeInTheDocument()
+        expect(hints.queryByText(/\(/u)).not.toBeInTheDocument()
+    })
+
+    it("seeds the expected keys as a template, keeping existing values", async () => {
+        act(() => useAgentChatHistoryStore.getState().setSlyData(NETWORK_ID, {x: 3}))
+
+        renderDialog(undefined, MATH_GUY_SCHEMA)
+        await user.click(screen.getByRole("button", {name: "Fill sly data template"}))
+
+        expect(getEditor()).toHaveValue('{\n  "x": 3,\n  "y": 0\n}')
+
+        // The template is only a draft until the user saves it
+        expect(getSlyData()).toEqual({x: 3})
+        await user.click(screen.getByRole("button", {name: "Save"}))
+        expect(getSlyData()).toEqual({x: 3, y: 0})
+    })
+
+    it("never seeds the app-supplied keys into the template", async () => {
+        renderDialog(["llm_config"], BYOK_SCHEMA)
+        await user.click(screen.getByRole("button", {name: "Fill sly data template"}))
+
+        expect(getEditor()).toHaveValue('{\n  "running_cost": 0\n}')
+    })
+
+    it("disables the template button while the JSON does not parse", async () => {
+        renderDialog(undefined, MATH_GUY_SCHEMA)
+        await replaceEditorContents("{oops")
+
+        expect(screen.getByRole("button", {name: "Fill sly data template"})).toBeDisabled()
     })
 })
--- a/packages/ui-common/__tests__/components/AgentChat/ChatCommon/SlyDataDialog.test.tsx
+++ b/packages/ui-common/__tests__/components/AgentChat/ChatCommon/SlyDataDialog.test.tsx
@@ -295,7 +295,7 @@ describe("SlyDataDialog", () => {
     it("lists the keys the network expects, with type, requiredness and description", () => {
         renderDialog(undefined, MATH_GUY_SCHEMA)
 
-        const hints = within(screen.getByText(/This network expects/u).parentElement)
+        const hints = within(screen.getByTestId(`${DIALOG_ID}-schema-hints`))
         expect(hints.getByText("x")).toBeInTheDocument()
         expect(hints.getByText("y")).toBeInTheDocument()
         expect(hints.getAllByText(/\(float, required\)/u)).toHaveLength(2)
@@ -305,7 +305,7 @@ describe("SlyDataDialog", () => {
     it("leaves the app-supplied keys out of the schema hints", () => {
         renderDialog(["llm_config"], BYOK_SCHEMA)
 
-        const hints = within(screen.getByText(/This network expects/u).parentElement)
+        const hints = within(screen.getByTestId(`${DIALOG_ID}-schema-hints`))
         expect(hints.getByText("running_cost")).toBeInTheDocument()
         expect(hints.queryByText("llm_config")).not.toBeInTheDocument()
     })
@@ -313,7 +313,7 @@ describe("SlyDataDialog", () => {
     it("renders a bare key when the schema gives no type, requiredness or description", () => {
         renderDialog(undefined, {properties: {plain: {}}, type: "object"})
 
-        const hints = within(screen.getByText(/This network expects/u).parentElement)
+        const hints = within(screen.getByTestId(`${DIALOG_ID}-schema-hints`))
         expect(hints.getByText("plain")).toBeInTheDocument()
         expect(hints.queryByText(/\(/u)).not.toBeInTheDocument()
     })
--- a/packages/ui-common/__tests__/components/AgentChat/ChatCommon/SlyDataDialog.test.tsx
+++ b/packages/ui-common/__tests__/components/AgentChat/ChatCommon/SlyDataDialog.test.tsx
@@ -320,7 +320,7 @@ describe("SlyDataDialog", () => {
     })
 
     it("seeds the expected keys as a template, keeping existing values", async () => {
-        act(() => useAgentChatHistoryStore.getState().setSlyData(NETWORK_ID, {x: 3}))
+        useAgentChatHistoryStore.getState().setSlyData(NETWORK_ID, {x: 3})
 
         renderDialog(undefined, MATH_GUY_SCHEMA)
         await user.click(screen.getByRole("button", {name: "Fill sly data template"}))
@@ -358,7 +358,7 @@ describe("SlyDataDialog", () => {
     })
 
     it("shows stored values instead of the template when the network already has sly data", () => {
-        act(() => useAgentChatHistoryStore.getState().setSlyData(NETWORK_ID, {x: 3}))
+        useAgentChatHistoryStore.getState().setSlyData(NETWORK_ID, {x: 3})
 
         renderDialog(undefined, MATH_GUY_SCHEMA)
 
@@ -366,7 +366,7 @@ describe("SlyDataDialog", () => {
     })
 
     it("resets to the template, not to nothing, when clearing on a network with a schema", async () => {
-        act(() => useAgentChatHistoryStore.getState().setSlyData(NETWORK_ID, {extra: 9, x: 3}))
+        useAgentChatHistoryStore.getState().setSlyData(NETWORK_ID, {extra: 9, x: 3})
 
         renderDialog(undefined, MATH_GUY_SCHEMA)
         await user.click(screen.getByRole("button", {name: "Clear sly data"}))
--- a/packages/ui-common/__tests__/components/AgentChat/ChatCommon/SlyDataDialog.test.tsx
+++ b/packages/ui-common/__tests__/components/AgentChat/ChatCommon/SlyDataDialog.test.tsx
@@ -284,12 +284,13 @@ describe("SlyDataDialog", () => {
     })
 
     it("ignores an unusable schema rather than breaking the dialog", () => {
-        vi.spyOn(console, "warn").mockImplementation(() => undefined)
+        const warn = vi.spyOn(console, "warn").mockImplementation(() => undefined)
 
         renderDialog(undefined, {type: "object"})
 
         expect(screen.queryByText(/This network expects/u)).not.toBeInTheDocument()
         expect(getEditor()).toHaveValue("{}")
+        expect(warn).toHaveBeenCalledExactlyOnceWith(expect.stringContaining("sly_data_schema"), expect.anything())
     })
 
     it("lists the keys the network expects, with type, requiredness and description", () => {
--- a/packages/ui-common/__tests__/components/AgentChat/ChatCommon/SlyDataDialog.test.tsx
+++ b/packages/ui-common/__tests__/components/AgentChat/ChatCommon/SlyDataDialog.test.tsx
@@ -296,17 +296,17 @@ describe("SlyDataDialog", () => {
         renderDialog(undefined, MATH_GUY_SCHEMA)
 
         const hints = within(screen.getByTestId(`${DIALOG_ID}-schema-hints`))
-        expect(hints.getByText("x")).toBeInTheDocument()
-        expect(hints.getByText("y")).toBeInTheDocument()
+        hints.getByText("x")
+        hints.getByText("y")
+        hints.getByText(/The first operand/u)
         expect(hints.getAllByText(/\(float, required\)/u)).toHaveLength(2)
-        expect(hints.getByText(/The first operand/u)).toBeInTheDocument()
     })
 
     it("leaves the app-supplied keys out of the schema hints", () => {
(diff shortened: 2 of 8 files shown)


packages/ui-common/utils/SlyDataSchema.ts
```diff
@@ -0,0 +1,193 @@
...
+// spellings cost nothing to accept and schema authors do write them.
+const INT_TYPES: ReadonlySet<string> = new Set(["int", "integer"])
+const FLOAT_TYPES: ReadonlySet<string> = new Set(["double", "float", "number"])
+const BOOL_TYPES: ReadonlySet<string> = new Set(["bool", "boolean"])
+
+/**
+ * One node of a sly_data_schema document. Loose on purpose: extra keys are the schema author's business.
+ * The recursion mirrors JSON Schema's -- an object node describes its keys under `properties`.
+ */
+const SchemaNode = z.looseObject({
```
M-Elsaied, line 52: validates the schema document (server-supplied input we don't control)

packages/ui-common/utils/SlyDataSchema.ts
```diff
@@ -0,0 +1,193 @@
...
+ */
+const parseSchemaRoot = (schema: unknown): SchemaNodeType | null => {
+    if (schema === undefined || schema === null) {
+        return null
+    }
+
+    const parsed = SchemaNode.safeParse(schema)
+    if (!parsed.success || !isObjectNode(parsed.data) || Object.keys(parsed.data.properties ?? {}).length === 0) {
+        // A bad schema is the network author's bug, not the user's: degrade to the schema-less experience.
+        console.warn("Ignoring unusable sly_data_schema:", schema)
```
M-Elsaied, line 103: Being defensive here -  a malformed schema is the network bug (Hopefully never the case, we have faith on the server side :) ), so the dialog silently behaves as if no schema was declared

packages/ui-common/utils/SlyDataSchema.ts
```diff
@@ -0,0 +1,193 @@
...
+ * @param schema The network's sly_data_schema. May be anything; an unusable schema returns `existing` untouched.
+ * @param existing The current sly_data. Every existing key survives verbatim; the template only adds what is missing.
+ * @param excludeKeys Top-level keys never to seed: the ones the UI supplies at send time. Keeping these out matters
+ * for more than tidiness -- `llm_config` holds API keys, which must never be led into the persistent store.
+ * @returns A new object; the input is not mutated.
+ */
+export const buildSlyDataTemplate = (
+    schema: unknown,
+    existing: SlyData,
+    excludeKeys: readonly string[] = []
```
M-Elsaied, line 183: Excluding it here means a template can never lead API keys into the editor/store

packages/ui-common/utils/SlyDataSchema.ts
```diff
@@ -0,0 +1,193 @@
...
+}
+
+/**
+ * Produces the template value for one schema node, folding in whatever the user already has.
+ * An existing value always wins -- even one that contradicts the schema -- because a template must never
+ * destroy user input. Objects are merged recursively so a partially-filled nested value gains its missing keys.
+ */
+const templateValue = (node: SchemaNodeType, existing: unknown, depth: number): unknown => {
+    if (isObjectNode(node)) {
+        if (existing !== undefined && (existing === null || typeof existing !== "object" || Array.isArray(existing))) {
```
M-Elsaied, line 149: An existing value always wins, even one that contradicts the schema

packages/ui-common/components/AgentChat/ChatCommon/SlyDataDialog.tsx
```diff
@@ -53,6 +55,12 @@ export interface SlyDataDialogProps {
     readonly networkDisplayName: string
 
     readonly onClose: () => void
+
+    /**
+     * The sly_data_schema the network's front man advertises, if any. When present and usable, the dialog lists
+     * the expected keys and offers to seed them as a template.
+     */
+    readonly slyDataSchema?: Record<string, unknown>
```
M-Elsaied, line 64: Display metadata only

packages/ui-common/components/AgentChat/ChatCommon/SlyDataDialog.tsx
```diff
@@ -72,8 +80,25 @@ const IMAGE_PREVIEW_SIZE = 96
 const SLY_DATA_EXPLAINER =
     "Data is sent to the agent network in a private channel with each request but kept out of the LLM."
 
+// Styling for a sly_data key name rendered as an inline code chip
+const KEY_CHIP_SX = {
```
M-Elsaied: Pure extraction - No visual change

packages/ui-common/components/AgentChat/ChatCommon/SlyDataDialog.tsx
```diff
@@ -121,6 +147,13 @@ export const SlyDataDialog: FC<SlyDataDialogProps> = ({
 
     const images = useMemo(() => findImageValues(parseResult.value), [parseResult])
 
+    // The keys the network says it expects, minus the ones the app supplies on its own. Empty for the many
+    // networks that do not advertise a schema, in which case the dialog looks exactly as it always has.
+    const schemaEntries = useMemo(
```
M-Elsaied: Everything new is gated on this being non-empty…

packages/ui-common/components/AgentChat/ChatCommon/SlyDataDialog.tsx
```diff
@@ -146,6 +179,10 @@ export const SlyDataDialog: FC<SlyDataDialogProps> = ({
         startOrUpdateDraft("{}")
     }
 
+    const handleFillTemplate = () => {
```
M-Elsaied, line 200: Flows through the existing draft-string model on purpose - no new state…

packages/ui-common/components/AgentChat/ChatCommon/SlyDataDialog.tsx
```diff
@@ -190,6 +227,21 @@ export const SlyDataDialog: FC<SlyDataDialogProps> = ({
 
     const getToolbar = () => (
         <Box sx={{display: "flex", gap: 0.5, justifyContent: "flex-end"}}>
+            {schemaEntries.length > 0 && (
+                <Tooltip title="Insert the keys this network expects (keeps your values)">
+                    <span>
+                        <IconButton
+                            aria-label="Fill sly data template"
+                            disabled={Boolean(parseResult.error)}
```
M-Elsaied, line 253: Same gating as export; the template merges under the current values...

packages/ui-common/components/AgentChat/ChatCommon/SlyDataDialog.tsx
```diff
@@ -320,12 +413,7 @@ export const SlyDataDialog: FC<SlyDataDialogProps> = ({
                         {index > 0 && ", "}
                         <Box
                             component="code"
-                            sx={{
-                                backgroundColor: "action.hover",
-                                borderRadius: 0.5,
-                                fontFamily: "monospace",
-                                px: 0.5,
-                            }}
+                            sx={KEY_CHIP_SX}
```
M-Elsaied: Abstracted on top (ln 83)

packages/ui-common/components/AgentChat/ChatCommon/ChatCommon.tsx
```diff
@@ -197,6 +197,12 @@ export interface ChatCommonProps {
      * If true, indicates that the user is missing API keys for one or more LLM providers.
      */
     readonly hasMissingApiKeys?: boolean
+
+    /**
+     * The sly_data_schema the selected network's front man advertises, if any. Enables schema hints in the
+     * sly_data dialog.
+     */
+    readonly slyDataSchema?: Record<string, unknown>
```
M-Elsaied, line 205: Flagging for the layers discussion in #498: this adds no new `sly_data` layer - nothing is merged into requests and the send path is untouched. It's display metadata only, and `ChatCommon` treats it as an opaque prop passed straight to the dialog. If/when the layers refactor lands, this prop just rides along with wherever `sly_data` handling moves.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -390,13 +393,15 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
...
                 setSupportedByokProviders(new Set())
+                setSlyDataSchema(undefined)
 
                 // Get agent function from server
                 const agentFunction = await getAgentFunction(neuroSanURL, selectedNetwork, username)
                 setNetworkDescription(agentFunction?.function?.description || "")
 
                 const schema = agentFunction?.function?.sly_data_schema
                 setSupportedByokProviders(getSupportedLlmProviders(schema))
+                setSlyDataSchema(schema)
```
M-Elsaied, line 404: No new fetch - `fetchAgentDetails` already retrieved this for BYOK provider detection and discarded the rest.

packages/ui-common/utils/SlyDataSchema.ts
```diff
@@ -0,0 +1,193 @@
...
+const SchemaNode = z.looseObject({
+    description: z.string().optional(),
+    get properties() {
+        return z.record(z.string(), SchemaNode).optional()
+    },
+    required: z.array(z.string()).optional(),
+    type: z.string().optional(),
+})
+
+type SchemaNodeType = z.infer<typeof SchemaNode>
```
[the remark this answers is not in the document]
M-Elsaied: @dsargent I feel that this is factually wrong, isn't `z.LooseObject` first class? Also it's not breaking at runtime :)

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
```diff
@@ -144,6 +144,9 @@ export const MultiAgentAccelerator: FC<MultiAgentAcceleratorProps> = ({
     // LLM providers for which BYOK keys are supported by the current network
     const [supportedByokProviders, setSupportedByokProviders] = useState<ReadonlySet<LLMProvider>>(new Set())
 
+    // The sly_data_schema advertised by the current network, if any. Drives schema hints in the sly_data dialog.
+    const [slyDataSchema, setSlyDataSchema] = useState<Record<string, unknown> | undefined>(undefined)
```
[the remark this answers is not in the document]
M-Elsaied: typing it unknown would be looser than the API contract, and there are no casts anywhere in the path...

packages/ui-common/__tests__/components/AgentChat/ChatCommon/SlyDataDialog.test.tsx
```diff
@@ -251,4 +275,74 @@ describe("SlyDataDialog", () => {
...
+        renderDialog(undefined, {type: "object"})
+
+        expect(screen.queryByText(/This network expects/u)).not.toBeInTheDocument()
+        expect(getEditor()).toHaveValue("{}")
+    })
+
+    it("lists the keys the network expects, with type, requiredness and description", () => {
+        renderDialog(undefined, MATH_GUY_SCHEMA)
+
+        const hints = within(screen.getByText(/This network expects/u).parentElement)
```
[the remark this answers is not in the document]
M-Elsaied: Half-valid - I'd accept the substance, not the reasoning. Addressed

packages/ui-common/utils/SlyDataSchema.ts
```diff
@@ -0,0 +1,193 @@
...
+    excludeKeys: readonly string[] = []
+): readonly SlyDataSchemaEntry[] => {
+    const root = parseSchemaRoot(schema)
+    if (root === null) {
+        return []
+    }
+
+    const required = new Set(root.required)
+
+    return Object.entries(root.properties)
```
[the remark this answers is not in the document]
M-Elsaied: The invariant is enforced by `parseSchemaRoot` in this same file - both call sites are behind it, and it guarantees non-empty properties.

packages/ui-common/__tests__/components/AgentChat/ChatCommon/SlyDataDialog.test.tsx
```diff
@@ -251,4 +275,74 @@ describe("SlyDataDialog", () => {
...
+        expect(hints.getByText("running_cost")).toBeInTheDocument()
+        expect(hints.queryByText("llm_config")).not.toBeInTheDocument()
+    })
+
+    it("renders a bare key when the schema gives no type, requiredness or description", () => {
+        renderDialog(undefined, {properties: {plain: {}}, type: "object"})
+
+        const hints = within(screen.getByText(/This network expects/u).parentElement)
```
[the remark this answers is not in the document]
M-Elsaied: Already addressed

packages/ui-common/utils/SlyDataSchema.ts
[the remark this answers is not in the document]
dsargent: I agree, and we're already using it in `SlyData.ts`. Copilot hallucination.

packages/ui-common/components/MultiAgentAccelerator/MultiAgentAccelerator.tsx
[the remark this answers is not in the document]
dsargent: Copilot is taking "anything" too literally. The server at least guarantees we'll get name-value pairs, so your `<string, unknown>` is correct here.

M-Elsaied:
@dsargent This is ready for your review 👀

dsargent:
First comment - thanks for the clean PR, good description + screenshot, good annotations of your changes, and "making Copilot your first reviewer" as well as addressing the issues it found. All things we encourage!

dsargent:
I was hoping it would pre-populate the input dialog with the appropriate schema, for the user to "fill in the blanks" -- doesn't seem like it would be a lot more effort?

M-Elsaied:
> I was hoping it would pre-populate the input dialog with the appropriate schema, for the user to "fill in the blanks" -- doesn't seem like it would be a lot more effort?

Hmm thank you for the comment, I agree that this is the intent and end state. As I was running from cashed runs so was under the fake impression that is pre-populated. Thank you for flagging this, will address accordingly

packages/ui-common/components/AgentChat/ChatCommon/SlyDataDialog.tsx
```diff
@@ -127,14 +127,33 @@ export const SlyDataDialog: FC<SlyDataDialogProps> = ({
...
 
+    // The keys the network says it expects, minus the ones the app supplies on its own. Empty for the many
+    // networks that do not advertise a schema, in which case the dialog looks exactly as it always has.
+    const schemaEntries = useMemo(
+        () => describeSlyDataSchema(slyDataSchema, extraSlyDataKeys),
+        [slyDataSchema, extraSlyDataKeys]
+    )
+
+    // The network's sly_data skeleton, pretty-printed, or null when it advertises no usable schema
+    const templateText = useMemo(
```
M-Elsaied, line 142: Introduce the template from the `sly-data` skeleton, else null if none exist

packages/ui-common/components/AgentChat/ChatCommon/SlyDataDialog.tsx
```diff
@@ -127,14 +127,33 @@ export const SlyDataDialog: FC<SlyDataDialogProps> = ({
...
+        () =>
+            schemaEntries.length > 0 ? formatSlyData(buildSlyDataTemplate(slyDataSchema, {}, extraSlyDataKeys)) : null,
+        [schemaEntries, slyDataSchema, extraSlyDataKeys]
+    )
+
+    // When the dialog opens onto an empty store and the network declares a schema, start from the template so
+    // the user fills in blanks instead of typing structure. It is an ordinary unsaved draft: nothing touches
+    // the store unless they save. The dialog is mounted fresh on every open, so this is per-open behavior.
+    const prefillText = storedText === "{}" ? templateText : null
```
M-Elsaied, line 151: The pre-population is here

packages/ui-common/utils/SlyDataSchema.ts
```diff
@@ -0,0 +1,193 @@
...
+/**
+ * One node of a sly_data_schema document. Loose on purpose: extra keys are the schema author's business.
+ * The recursion mirrors JSON Schema's -- an object node describes its keys under `properties`.
+ */
+const SchemaNode = z.looseObject({
+    description: z.string().optional(),
+    get properties() {
+        return z.record(z.string(), SchemaNode).optional()
+    },
+    required: z.array(z.string()).optional(),
```
dsargent, line 57: Why isn't `required` just a `boolean`?

packages/ui-common/utils/SlyDataSchema.ts
```diff
@@ -0,0 +1,193 @@
...
+    type: z.string().optional(),
```
dsargent, line 58: Is there a way to make this some kind of "enumerated value", rather than just "any old string"? We only expect a finite set of options here.

packages/ui-common/utils/SlyDataSchema.ts
dsargent: We could consider displaying a banner alert for this case, just so the user knows what's up and they can report the issue.

packages/ui-common/utils/SlyDataSchema.ts
```diff
@@ -0,0 +1,193 @@
...
+ * @param schema The network's sly_data_schema, as returned by the function endpoint. May be anything.
+ * @param excludeKeys Keys to leave out: the ones the UI itself supplies at send time (API keys, login...), which
+ * the user should not be prompted for.
+ * @returns One entry per declared key in declaration order, or an empty array when the schema is unusable.
+ */
+export const describeSlyDataSchema = (
+    schema: unknown,
+    excludeKeys: readonly string[] = []
+): readonly SlyDataSchemaEntry[] => {
+    const root = parseSchemaRoot(schema)
```
dsargent, line 121: I'm glad you used Zod here. Neat pattern, clean.

packages/ui-common/utils/SlyDataSchema.ts
dsargent, line 128: Nice, functional programming style. This is how I would have done it.

packages/ui-common/utils/SlyDataSchema.ts
```diff
@@ -0,0 +1,193 @@
...
+            type: node.type ?? (isObjectNode(node) ? "object" : undefined),
+        }))
+}
+
+/**
+ * Produces the template value for one schema node, folding in whatever the user already has.
+ * An existing value always wins -- even one that contradicts the schema -- because a template must never
+ * destroy user input. Objects are merged recursively so a partially-filled nested value gains its missing keys.
+ */
+const templateValue = (node: SchemaNodeType, existing: unknown, depth: number): unknown => {
```
dsargent, line 147: In general, please add JSDoc for parameters.

packages/ui-common/utils/SlyDataSchema.ts
```diff
@@ -0,0 +1,193 @@
...
+            merged[key] = templateValue(child, merged[key], depth + 1)
+        })
+        return merged
+    }
+
+    if (existing !== undefined) {
+        return existing
+    }
+
+    if (node.type === "string") {
```
dsargent: Could also be a const map with `switch`. Not a biggie.

packages/ui-common/components/AgentChat/ChatCommon/SlyDataDialog.tsx
```diff
@@ -366,6 +475,7 @@ export const SlyDataDialog: FC<SlyDataDialogProps> = ({
             title={getTitle()}
         >
             {getToolbar()}
+            {getSchemaHints()}
```
dsargent, line 472: This is _slightly_ redundant now that you're actually populating the input with the schema shape. I'm torn because I like that it has the description and type, and you technically can't have comments in JSON so you can't really inline the comment and type...but I don't like the duplication in the display.

image

We could do fake-JSON with comments:

```json
{
  // required, float: The running cost of the operation
  "running_cost": 15
}
```

I'm not sure about this. I guess either way is fine, as you have it or with "pseudo JSON".

packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
```diff
@@ -1673,6 +1673,21 @@ describe("ChatCommon", () => {
...
+                slyDataSchema: {
+                    properties: {x: {description: "The first operand", type: "float"}},
+                    required: ["x"],
+                    type: "object",
+                },
+            })
+
+            await openSlyDataEditor()
+
+            expect(screen.getByText(/This network expects/u)).toBeInTheDocument()
```
dsargent: We have a de facto, opinionated (as in, Darren's opinion) convention to leave off the `expect...toBeInTheDocument` since it's not required: `getByText()` _et al_ throw if the item is not found, causing the test to fail.

packages/ui-common/__tests__/components/AgentChat/ChatCommon/SlyDataDialog.test.tsx
```diff
@@ -251,4 +275,101 @@ describe("SlyDataDialog", () => {
...
+
+    it("shows no schema hints and no template button when the network declares no schema", () => {
+        renderDialog()
+
+        expect(screen.queryByText(/This network expects/u)).not.toBeInTheDocument()
+        expect(screen.queryByRole("button", {name: "Fill sly data template"})).not.toBeInTheDocument()
+    })
+
+    it("ignores an unusable schema rather than breaking the dialog", () => {
+        vi.spyOn(console, "warn").mockImplementation(() => undefined)
```
dsargent: Usually, in tests if you're spying on something, it's good to assert that it did what you expect after the test runs.

In this case, `expect(console.warn).toHaveBeenCalledExactlyOnceWith(/something/)`.

I shy away from matching the entire, exact error string as that is fragile and causes annoying test breakages when doing trivial refactors like fixing spelling or punctuation. Just assert that "the string we got has the essential items we expected".

packages/ui-common/__tests__/unit/Utils/SlyDataSchema.test.ts
```diff
@@ -0,0 +1,229 @@
...
+        },
+        running_cost: {description: "The running cost of the operation", type: "float"},
+    },
+    required: ["llm_config"],
+    type: "object",
+}
+
+// The schema of neuro-san's mcp_github example network: objects nested three deep, with a URL as a property key
+const MCP_URL = "https://api.githubcopilot.com/mcp"
+const MCP_SCHEMA = {
```
dsargent, line 55: Does this mean that in theory we could already "do MCP"?! By having the agent network define that it wants MCP info in `sly_data`, and using the new sly_data editor to supply it? That would be neat.

dsargent (dismissed):
Nice work. I had very little to say about this, which is unusual for me as normally I have a _lot_ to say about PRs, so that's a good thing.

I retested the dialog with a sly_data-enabled network and it was fine this time -- correctly prepopulates and I verified that it sends the correct sly_data values in the request to Neuro SAN.

A few minor thoughts but for me it's mergeable as-is.

packages/ui-common/__tests__/components/AgentChat/ChatCommon/SlyDataDialog.test.tsx
```diff
@@ -251,4 +275,101 @@ describe("SlyDataDialog", () => {
...
+    it("renders a bare key when the schema gives no type, requiredness or description", () => {
+        renderDialog(undefined, {properties: {plain: {}}, type: "object"})
+
+        const hints = within(screen.getByTestId(`${DIALOG_ID}-schema-hints`))
+        expect(hints.getByText("plain")).toBeInTheDocument()
+        expect(hints.queryByText(/\(/u)).not.toBeInTheDocument()
+    })
+
+    it("seeds the expected keys as a template, keeping existing values", async () => {
+        act(() => useAgentChatHistoryStore.getState().setSlyData(NETWORK_ID, {x: 3}))
```
swensel: @M-Elsaied I don't think you need `act` here. This runs before `renderDialog`, so this store update isn't triggering any React work. I think this can just be `useAgentChatHistoryStore.getState().setSlyData(NETWORK_ID, {x: 3})`.

Same idea for the other instances below that call `act` before `renderDialog`.

packages/ui-common/components/AgentChat/ChatCommon/SlyDataDialog.tsx
swensel: @M-Elsaied You could rewrite this with `styled`:

```
const KeyChip = styled(Box)(({theme}) => ({
    backgroundColor: theme.palette.action.hover,
    borderRadius: theme.shape.borderRadius * 0.5,
    fontFamily: "monospace",
    paddingLeft: theme.spacing(0.5),
    paddingRight: theme.spacing(0.5),
}))
```

Then:

```
<KeyChip component="code">
    ...
</KeyChip>
```

swensel (dismissed):
LGTM. Good work. 🎉 

Couple comments. No blockers.

packages/ui-common/components/AgentChat/ChatCommon/SlyDataDialog.tsx
M-Elsaied: Changed as per suggestion - used in both the hints and the merged-keys note.

packages/ui-common/utils/SlyDataSchema.ts
M-Elsaied: It's the wire format rather than our design

packages/ui-common/utils/SlyDataSchema.ts
M-Elsaied: The finite set lives one layer down: the `INT_TYPES/FLOAT_TYPES/BOOL_TYPES` sets (+ string/object) are the enumeration. Happy to switch to `z.enum(...).catch(undefined)` if you'd rather hard-drop unknown types, but showing them verbatim seemed more useful...

packages/ui-common/utils/SlyDataSchema.ts
M-Elsaied: Good practice - added..

packages/ui-common/utils/SlyDataSchema.ts
M-Elsaied: Good call - went one further: a single `SCALAR_TYPE_DEFAULTS` map replaced both the if-chain and the three alias Sets it fed

packages/ui-common/components/AgentChat/ChatCommon/SlyDataDialog.tsx
M-Elsaied: Leaning to keeping it as-is.. I hear the duplication concern but feels like good added value and less overhead maintaining comments inside JSON. So decoupling makes sense from this angle

packages/ui-common/__tests__/components/AgentChat/ChatCommon/ChatCommon.test.tsx
M-Elsaied: Done

packages/ui-common/__tests__/unit/Utils/SlyDataSchema.test.ts
M-Elsaied: In theory... yes :)

packages/ui-common/__tests__/components/AgentChat/ChatCommon/SlyDataDialog.test.tsx
M-Elsaied: Good catch... Adressed

packages/ui-common/components/AgentChat/ChatCommon/SlyDataDialog.tsx
```diff
@@ -74,6 +83,26 @@ const SLY_DATA_EXPLAINER =
 
 //#endregion: Constants
 
+// A sly_data key name rendered as an inline code chip
+const KeyChip = styled("code")(({theme}) => ({
```
swensel, line 87: @M-Elsaied Thanks for making this change.

packages/ui-common/__tests__/components/AgentChat/ChatCommon/SlyDataDialog.test.tsx
M-Elsaied: Good catch - Dropped it

M-Elsaied:
@dsargent @swensel I think I addressed most of the comments. Let me know if I missed anything

packages/ui-common/__tests__/components/AgentChat/ChatCommon/SlyDataDialog.test.tsx
swensel: Thanks for updating that.

swensel (approved):

dsargent (approved):

packages/ui-common/utils/SlyDataSchema.ts
dsargent: Some kind of visual cue might be helpful -- "we're, reluctantly, displaying this, but it's not a type we expect."

Issue #404, opened 2026-06-13
Replace Node.js polyfills with browser native APIs


dsargent:
Our app runs client-side so it cannot use Node.js functions. What is likely happening is that Next.js build/Turbopack is silently polyfilling these for us at build time.

For example, these show up when enabling `import-x/no-nodejs-modules`:

```console
neuro-san-ui/apps/main/next.config.ts
  6:1  error  Do not import Node.js builtin module "node:path"  import-x/no-nodejs-modules

neuro-san-ui/jest.setup.ts
  21:1  error  Do not import Node.js builtin module "node:stream/web"  import-x/no-nodejs-modules
  22:1  error  Do not import Node.js builtin module "node:util"        import-x/no-nodejs-modules

neuro-san-ui/packages/ui-common/utils/text.ts
  21:1  error  Do not import Node.js builtin module "node:crypto"  import-x/no-nodejs-modules
```

Copilot says it's outdated and may cause bundle bloat:

> import ... from "node:crypto" should not work in a browser environment, as crypto is a native Node.js module.
The fact that it is working points directly to your suspicion: your build tool is polyfilling it.
Here's the breakdown:
What's Happening: Modern web development build tools (like the Webpack configuration used by Next.js) are designed to smooth over the differences between server and client environments. When the bundler encounters an import for a Node.js built-in module like crypto, fs, or path, it automatically substitutes it with a browser-compatible library that mimics the Node.js API.
The Likely Polyfill: For the crypto module, the most common polyfill is a library called crypto-browserify. Your build tool effectively rewrites the import on the fly:
It sees: import {createHash} from "node:crypto"
It produces in the final bundle: import {createHash} from "crypto-browserify" (or equivalent logic).
Is This a Problem? Not necessarily, but it's something to be aware of:
Bundle Size: Polyfills add weight to your final JavaScript bundle. The crypto-browserify library is not trivial.
Modern Alternatives: For years, this was the only way. Now, browsers have a native and powerful SubtleCrypto API, available via window.crypto.subtle. It's more secure, asynchronous (using Promises), and doesn't require a polyfill.
```

Issue #323, opened 2026-05-13
Explore new user DX for ui-common


dsargent:
We have open sourced neuro-san-ui and we publish `ui-common` to the npm repo.

For a theoretical external dev, how does the experience look when they try to consume `ui-common` and use the shared components in (1) an existing legacy codebase and (2) a freshly-created React.js project?

Where are the rough edges? The awkward dependencies? The unobvious parts?

Reference: Donn's early 2026 FedEx day project.

Issue #209, opened 2026-03-12
[UN-2800] UI: consider getting rid of next-auth


dsargent:
**Migrated from Jira:** [UN-2800](https://leaf-ai.atlassian.net/browse/UN-2800)

## Description
Now that we’ve moved to ALB authentication for all deployed environments, can we get rid of next-auth from the UI?

One issue is that this is how we authenticate when testing locally, via next-auth + Github. We would need to find another way to do that.

## Jira Details
- **Issue Type:** Task
- **Priority:** Major
- **Status:** Nice to Haves
- **Reporter:** Darren Sargent
- **Assignee:** Darren Sargent
- **Created:** 2025-01-17 11:36:33
- **Updated:** 2025-03-13 10:33:29

Issue #163, opened 2026-02-24
Remove Node Polyfill Requirement from ui-common to Improve DX


swensel:
We’re currently seeing build failures when consuming `@cognizant-ai-lab/ui-common` in a fresh project due to missing Node core module polyfills (`crypto`, `fs`, `stream`, `zlib`, etc). Right now, consumers are required to manually configure webpack fallbacks and install polyfill packages to make the library work.

We do not want to require polyfills or extra bundler configuration for consumers. From a DX perspective, installing and importing the library in a standard React/Webpack app should work out of the box with zero additional setup. The goal is to remove any Node-specific assumptions from the browser bundle so the library is frictionless to adopt.

swensel:
Let's revisit this after removing Next.js. MAUI will have a Node.js server then.

PR #522, opened 2026-08-31
#323, #163, #404 - Fix ui-common install and imports for external consumers

swensel:
# Description

Streamlining the DX for `@cognizant-ai-lab/ui-common`.

I scaffolded a Vite app and installing published 1.11.0 failed immediately (TypeScript 6 vs our `typescript@^5.9.3` peer).

That's #323 (fresh React app). Same class of problem as the webpack polyfills in #163 and the Node builtins in #404: `hashString` imported Node crypto, which blows up in the browser.

neuro-ui is still on 1.4.1 and nsflow on 1.5.1, both MUI 7. 1.5.1 (a patch) raised MUI peers to 9.

This PR:

- Drops `typescript` and `next` as required peers. Vite's TS 6 fails against `typescript@^5.9.3`; a non-Next app shouldn't need `next` as a peer. `next-auth` is optional. MUI is `^7.3.1 || ^8 || ^9` so nsflow / neuro-ui can use it (they're on 7; 1.5.1 required 9).
- Stops using Node crypto so a Vite app doesn't need polyfills. That covers the `utils/text.ts` hit in #404. The other hits there run at build time or in tests, so they stay.
- Fixes `./const` (it pointed at a `.ts` file that the published package never shipped).
- Takes `next-auth` off the barrel, so importing anything from the package root no longer pulls it into the module graph.
- Version 1.12.0. Git still said 1.5.0; registry latest is 1.11.0 (version was bumped at publish and not committed).

Verified a Vite app against published 1.11.0 (ERESOLVE) and against a packed tarball from this branch (install works, `MUIDialog` renders). Local GitHub login in `apps/main` still goes through next-auth.

I also traced the import graph of the built `dist/index.js`: across all 87 modules reachable from the barrel, none import `next-auth`.

## Taking next-auth off the barrel

Two modules import `next-auth`: `components/Authentication/Auth` and `utils/Authentication`. Both were re-exported from the package root, so `import {MUIDialog} from "@cognizant-ai-lab/ui-common"` dragged `next-auth/react` into the graph and a consumer without it got a hard resolve failure. Making the peer optional on its own would only have turned an install error into a build error.

There were three edges to cut. Two were the direct re-exports, now removed from `index.ts` with a comment saying why. The third was transitive: `ErrorBoundary` renders `ErrorPage`, which called `useAuthentication`.

So `ErrorPage` now takes the signed-in user, the authentication type, and a sign-out callback as props, and `ErrorBoundary` passes them through. That matches what the file was already doing, since `Navbar` takes `userInfo` as a prop and `ErrorPage` already rendered `Navbar` with props. It also collapses a duplicate sign-out path: `ErrorPage` had its own handler that cleared the user store and called `smartSignOut`, identical to the one in `_app`. There is now one.

`apps/main` keeps using next-auth directly, so local login is unchanged.

## Consumers

Both changes are breaking, and both are compile-time, so nobody silently gets a signed-out app.

- **`ErrorBoundary` gains three required props**: `userInfo`, `authenticationType`, and `signOut`. Anyone rendering it needs to pass what they already know about the user. `apps/main` does this with an `ErrorBoundaryWrapper`, shaped like the existing `NavbarWrapper`.
- **The package root no longer exports** `Auth`, `useAuthentication`, `smartSignOut`, `getAuthenticationType`, or `AD_TENANT_ID`. All still available on their subpaths, for example `@cognizant-ai-lab/ui-common/components/Authentication/Auth`.
- **neuro-ui** keeps rendering next-auth's own `SessionProvider`, so no change there. It will need the `ErrorBoundary` props and subpath imports for the auth helpers when it leaves 1.4.1.
- **nsflow** has no next-auth, which is the case this unblocks. Besides the `ErrorBoundary` props, they can drop the `dist/` alias and use the controller subpath.

## Other notes

- Unclear about the "existing legacy codebase" half of #323.
- I only tried it in a fresh Vite app (I can also try it in Webpack).
- Did not run this repo on MUI 7. Catalog is still 9.0.1. Peers go down to `^7.3.1` because nsflow is on `^7.3.2` and neuro-ui is on `7.3.1`. 1.5.1 raised MUI peers to 9 in a patch, which is why those two never moved.
- `strictNullChecks` is off repo-wide, so the new required props do not actually guarantee a non-undefined user. `ErrorPage` still has to render without one, and there's a test for that.
- #404 is only partly addressed. `utils/text.ts` was the one hit that reached a browser bundle. `jest.setup.ts` no longer exists since we moved to vitest, and `next.config.ts` runs in Node at build time. `import-x/no-nodejs-modules` is still not enabled, so that issue stays open.
- Does not touch #209 (getting rid of next-auth for local login). `apps/main` still signs in through next-auth.


apps/main/pages/_app.tsx
swensel: Only runtime change in apps/main. NextAuthSessionProvider is next-auth SessionProvider plus adapter registration so Auth/useAuthentication still talk to next-auth. Same enableAuthentication wrap. Local GitHub login stays. Without this swap, sign-in throws.

apps/main/__tests__/pages/App.test.tsx
swensel: _app now imports NextAuthAdapter, which imports signIn/signOut from next-auth/react. vi.mock replaces the whole module, so those two vi.fn() exports have to exist or the test cannot load. Not a new assertion.

apps/main/__tests__/pages/MultiAgentAcceleratorPage.test.tsx
swensel, line 1: The page reads username from useAuthentication(), which now uses getSessionAdapter() instead of next-auth useSession. Mocking next-auth/react is a no-op. setSessionAdapter injects the same fake user at the new seam so the username assertion still holds.

packages/ui-common/__tests__/components/Authentication/Auth.test.tsx
swensel, line 1: Auth.tsx no longer imports next-auth. renderWithSession drives the adapter for the same three cases: loading spinner, signIn(auth0) when unauthenticated, children when authenticated.

packages/ui-common/__tests__/components/MultiAgentAccelerator/MultiAgentAccelerator.test.tsx
swensel: Dead mock. This file renders MultiAgentAccelerator, which no longer loads next-auth/react. App.test.tsx still mocks it because _app really imports NextAuthAdapter.

packages/ui-common/__tests__/unit/Utils/Authentication.test.ts
swensel, line 1: smartSignOut / useAuthentication call adapter.signOut now, not next-auth signOut. The spy is registered via setSessionAdapter. beforeEach resets the singleton so tests do not leak.

packages/ui-common/__tests__/unit/Utils/SessionAdapter.test.ts
swensel, line 1: Contract for the no-next-auth default: unregistered signIn/signOut throw (loud, names the provider); useSession returns no user so apps without login still render; auth-disabled useAuthentication must not consult the adapter.

packages/ui-common/__tests__/unit/Utils/text.test.ts
swensel: hashString is no longer MD5. New expected hex is from the JS mix used for React keys. Same checks: stable, different inputs differ, empty and long strings work.

packages/ui-common/components/Authentication/Auth.tsx
swensel, line 1: Same spinner / signIn(auth0) / children. Calls go through getSessionAdapter so the barrel does not import next-auth. If NextAuthSessionProvider was imported, this is still next-auth. If not, sign-in throws with the provider name.

packages/ui-common/components/ErrorPage/ErrorPage.tsx
swensel: Navbar already takes pathname and query as plain props. ErrorPage was the only library file importing next/router for those two fields. getCurrentLocation reads window.location. This is why the next peer can be dropped. apps/main still uses Next.js.

packages/ui-common/const.ts
swensel: process.env throws in a browser bundle where process is undefined. Optional chaining on globalThis.process falls back to Unknown Version. One safe read so importing constants does not need a Node polyfill.

packages/ui-common/utils/Authentication.ts
swensel, line 1: useAuthentication and smartSignOut used next-auth useSession/signOut directly. Now getSessionAdapter(). Same ALB vs NextAuth branch. Unregistered sign-out throws instead of silently no-opping.

packages/ui-common/utils/BrowserNavigation.ts
swensel: getCurrentLocation is window.location plus URLSearchParams. Next-free stand-in for the two fields ErrorPage used to take from useRouter. navigateToUrl was already here.

packages/ui-common/utils/NextAuthAdapter.tsx
swensel, line 1: Only shipped file that imports next-auth/react. Not re-exported from index.ts. Importing it registers the adapter. NextAuthSessionProvider is a drop-in for SessionProvider. Vite apps that never import this file never need next-auth installed.

packages/ui-common/utils/SessionAdapter.ts
swensel, line 1: Socket the barrel talks to instead of next-auth. Default: no session so UIs still render; signIn/signOut throw and name NextAuthSessionProvider so missing wiring is loud. Register once at module init. resetSessionAdapter is tests only.

packages/ui-common/README.md
swensel: Rewrite so docs match the package. Old README used create-react-app, had no working import, and used props that do not exist. New: Vite, real props, MUI 7/8/9, next-auth optional, subpaths vs barrel. Treat as one rewrite.

yarn.lock
```diff
@@ -452,15 +452,16 @@ __metadata:
   peerDependencies:
```
swensel, line 452: Yarn recording the same peer change as package.json (MUI 7||8||9, optional next-auth, no next, no typescript). Only hunk in the lock is the ui-common workspace entry. No resolutions moved, nothing installed or upgraded.

packages/ui-common/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "@cognizant-ai-lab/ui-common",
     "description": "A shared UI component library for Cognizant AI Labs projects",
-    "version": "1.5.0",
+    "version": "1.12.0",
```
swensel: 1.12.0 because git still said 1.5.0 while npm latest is 1.11.0 (bumped at publish, never committed). Publishing 1.5.0 is rejected or steals `latest`. Not 2.0.0: two known consumers; we'll tell them about the SessionProvider swap.

Drop `typescript` peer: we ship compiled JS + `.d.ts`. Consumers don't compile us. `typescript@^5.9.3` made `npm create vite` (TS 6) ERESOLVE on 1.11.0. Reproduced in my-app. TS stays a devDependency of this repo.

Drop `next` peer: nothing on the barrel imports `next` anymore (ErrorPage uses `window.location`). A Vite app should not install Next 16 to render a dialog. `apps/main` is still Next.

`next-auth`: exact `5.0.0-beta.32` → `^5.0.0-beta.29` + optional. neuro-ui is on beta.30; the exact pin excluded them. Optional so a Vite app without login doesn't install it. Still required at runtime if they use `NextAuthSessionProvider`.

MUI `^9` → `^7.3.1 || ^8 || ^9` (x-tree-view `^8.22.0 || ^9.1.0`). 1.5.1 raised peers to 9 in a patch. This repo's catalog is 9; neuro-ui/nsflow are on 7. after/ installs on 7.3.1. Claim is install+tsc on 7, not pixel-identical 7 vs 9.

`./const`: `./const.ts` → `dist/const.js`. `files` is `["dist"]`; the `.ts` was never in the tarball. Map was wrong; JS was always packed.

Subpath exports + `types` before `import`. nsflow aliases into `dist/controller/...` because only `.` and `./const` were exported. This is the supported form. Types-first so TS 6 / bundler resolution picks `.d.ts` before JS.

apps/main/__tests__/pages/App.test.tsx
```diff
@@ -68,6 +68,9 @@ const APP_COMPONENT = (
 
 vi.mock("next-auth/react", () => ({
     SessionProvider: ({children}: {children: ReactNode}) => <>{children}</>,
+    useSession: vi.fn(() => ({data: undefined, status: "unauthenticated"})),
```
swensel, line 71: `_app.tsx` statically imports `Auth` and `useAuthentication`/`smartSignOut` from `utils/Authentication`, both of which import `signIn`/`signOut`/`useSession` from `next-auth/react`. The stub must define everything they import, since ESM fails at import time on a missing export.

apps/main/__tests__/pages/ErrorPage.test.tsx
```diff
@@ -32,49 +32,56 @@ vi.mock("next/router", () => ({
...
             />
         )
 
+    it("Should render correctly", async () => {
+        renderErrorPage(USER_INFO)
+
+        await screen.findByText(new RegExp(LOGO, "u"))
+    })
+
+    it("Should render when authentication has not produced a user yet", async () => {
```
swensel, line 67: `ErrorPage` no longer calls `useAuthentication()`, it takes `userInfo` as a prop. This test covers the case where the caller doesn't have a user yet, by passing `undefined` directly, so it still renders instead of throwing.

apps/main/pages/_app.tsx
```diff
@@ -81,6 +81,24 @@ const NavbarWrapper = (props: Omit<NavbarProps, "userInfo">): ReactElement => {
...
 }
 
+/**
+ * Utility component to pass user data along to the error boundary's fallback page. This component exists for the
+ * same reason as NavbarWrapper: the useAuthentication hook should only be called once the enableAuthentication
+ * setting is known.
+ * @param props All props for ErrorBoundary except userInfo.
+ * @return ErrorBoundary with userInfo passed to it.
+ */
+const ErrorBoundaryWrapper = (props: Omit<ErrorBoundaryProps, "userInfo">): ReactElement => {
```
swensel, line 91: `ErrorBoundaryWrapper` exists for the same reason as `NavbarWrapper`: `useAuthentication` can't be called until `enableAuthentication` is known, so it can't live in `NeuroSanUI` itself.

packages/ui-common/__tests__/components/ErrorPage/ErrorBoundary.test.tsx
```diff
@@ -2,11 +2,16 @@ import {render, screen} from "@testing-library/react"
 import {MockInstance} from "vitest"
 
 import {withStrictMocks} from "../../../../../__tests__/common/strictMocks"
-import {ErrorBoundary} from "../../../components/ErrorPage/ErrorBoundary"
+import {ErrorBoundary, ErrorBoundaryProps} from "../../../components/ErrorPage/ErrorBoundary"
 
 // Mock the ErrorPage component so tests don't need Next router or stores
 vi.mock("../../../components/ErrorPage/ErrorPage", () => ({
-    default: ({errorText}: {errorText: string}) => <div data-testid="mock-error">{errorText}</div>,
+    default: ({errorText, userInfo}: {errorText: string; userInfo: {name: string}}) => (
```
swensel, line 9: Update tests based on `ErrorBoundary` component changes.

packages/ui-common/__tests__/components/MultiAgentAccelerator/MultiAgentAccelerator.test.tsx
```diff
@@ -102,8 +102,6 @@ let setSelectedNetwork: (network: string) => void
 let onEditNetwork: (network: string) => void
 
 // Mock dependencies
-vi.mock("next-auth/react")
```
swensel, line 105: Removed because it's dead: `MultiAgentAccelerator` doesn't import `Auth` or `utils/Authentication`, the only two files here that touch `next-auth/react`, so the mock was never exercised. Deleting it changes nothing about the test.

packages/ui-common/__tests__/unit/Utils/BrowserNavigation.test.ts
```diff
@@ -10,4 +10,37 @@ describe("BrowserNavigation", () => {
         expect(spy).toHaveBeenCalledWith("http://example.com/test-path")
         spy.mockRestore()
     })
+
+    describe("getCurrentLocation", () => {
```
swensel, line 14: `getCurrentLocation` replaces `next/router`'s `useRouter()` in `ErrorPage`, reading `window.location` directly so the component works outside Next.js. These tests cover the pathname/query parsing and the fallback for when `location` has no pathname/search, or is missing entirely, since Next.js can server-render this fallback where there's no `location` global at all.

packages/ui-common/__tests__/unit/Utils/text.test.ts
```diff
@@ -44,10 +44,10 @@ describe("removeLast", () => {
 
 describe("hashString", () => {
     withStrictMocks()
-    it("should return a valid MD5 hash for a given string", () => {
+    it("should return a stable hash for a given string", () => {
```
swensel, line 47: No longer MD5. The `node:crypto` import made `ui-common` unusable in a browser bundle. Now a plain JS mix, fine for the only use (React keys). New hex values are just that function's output.

packages/ui-common/__tests__/unit/Utils/text.test.ts
```diff
@@ -68,12 +68,19 @@ describe("hashString", () => {
...
     })
 
     it("should handle a very long string", () => {
         const input = "a".repeat(1000)
         const result = hashString(input)
-        expect(result).toBe("cabe45dcc9ae5b66ba86600cca6b8ba8")
+        expect(result).toBe("1c0002518a359e")
+    })
+
+    it("should treat a missing code point as zero", () => {
```
swensel, line 80: Covers the `?? 0` fallback in the new hash loop.

packages/ui-common/components/ErrorPage/ErrorBoundary.tsx
```diff
@@ -40,11 +40,23 @@ interface ErrorBoundaryState {
...
 
 /**
- * Interface to define the incoming props for this component
+ * Interface to define the incoming props for this component.
+ *
+ * The session details are only used for the fallback page's navbar, and are passed down rather than read from an
+ * authentication library so that this component works in an app with no login at all.
  */
-interface ErrorBoundaryProps {
+export interface ErrorBoundaryProps {
```
swensel, line 48: Exported for `ErrorBoundaryWrapper` in `_app.tsx`, which needs `Omit<ErrorBoundaryProps, "userInfo">`.

packages/ui-common/components/ErrorPage/ErrorBoundary.tsx
```diff
@@ -40,11 +40,23 @@ interface ErrorBoundaryState {
...
     readonly id: string
     readonly children: ReactNode
+
+    // Info about the currently authenticated user, shown in the fallback page's navbar
+    readonly userInfo: ErrorPageProps["userInfo"]
```
swensel, line 53: Passing these down instead of having `ErrorPage` call `useAuthentication` takes `next-auth` off the package barrel. It's an optional peer dep now, so an app with no login can import from `ui-common`.

packages/ui-common/components/ErrorPage/ErrorPage.tsx
```diff
@@ -15,49 +15,51 @@ limitations under the License.
...
 
 import Box from "@mui/material/Box"
-import {useRouter} from "next/router.js"
 import {FC, ReactElement} from "react"
 
 import {LOGO} from "../../const"
 import {useEnvironmentStore} from "../../state/Environment"
-import {useUserInfoStore} from "../../state/UserInfo"
-import {getAuthenticationType, smartSignOut, useAuthentication} from "../../utils/Authentication"
+import {getCurrentLocation} from "../../utils/BrowserNavigation"
```
swensel, line 22: `Navbar` already takes `pathname` and `query` as plain props, and this was the last `next/router` import in the package. `getCurrentLocation` reads `globalThis.location` instead, so `next` is off `ui-common` entirely now, not even a peer dep. `apps/main` is still a Next app.

packages/ui-common/components/ErrorPage/ErrorPage.tsx
```diff
@@ -15,49 +15,51 @@ limitations under the License.
...
-import {Navbar} from "../Common/Navbar"
+import {Navbar, NavbarProps} from "../Common/Navbar"
 
-interface ErrorPageProps {
+export interface ErrorPageProps {
     id: string
     errorText: string
+
+    // Info about the currently authenticated user, shown in the navbar
+    readonly userInfo: NavbarProps["userInfo"]
```
swensel, line 32: `useAuthentication` moves out to the caller, which is what takes `next-auth` off the package barrel.

packages/ui-common/components/ErrorPage/ErrorPage.tsx
```diff
@@ -15,49 +15,51 @@ limitations under the License.
...
-        setCurrentUser(undefined)
-        setPicture(undefined)
+const ErrorPage: FC<ErrorPageProps> = ({
+    id,
+    errorText,
+    userInfo,
+    authenticationType,
+    signOut,
+}: ErrorPageProps): ReactElement => {
+    const {enableAuthentication, logoServiceToken, supportEmailAddress} = useEnvironmentStore()
```
swensel, line 60: `_app.tsx` already computed `authenticationType` and `handleSignOut` for the navbar, so this was a second copy. It passes the same values down now.

packages/ui-common/utils/BrowserNavigation.ts
```diff
@@ -23,3 +23,12 @@ limitations under the License.
 export const navigateToUrl = (url: string): void => {
     window.location.href = url
 }
+
+/**
+ * @returns The current pathname, and the query string parsed into a plain object
+ */
+export const getCurrentLocation = (): {pathname: string; query: Record<string, string>} => {
```
swensel, line 31: `getCurrentLocation` is `globalThis.location` plus `URLSearchParams`. Next-free stand-in for the two fields `ErrorPage` took from `useRouter`. `navigateToUrl` was already here.

packages/ui-common/package.json
swensel, line 4: `1.12.0` because git still said `1.5.0` while npm latest is `1.11.0` (bumped at publish, never committed). Publishing `1.5.0` is rejected or steals `latest`. Open to `2.0.0` instead: the barrel no longer re-exports the Authentication modules, and `ErrorBoundary`/`ErrorPage` gained required props, so there's a case for major. Nothing is published yet, I've been testing against a tarball from `yarn workspace @cognizant-ai-lab/ui-common pack`.

packages/ui-common/package.json
```diff
@@ -52,15 +52,18 @@
     "peerDependencies": {
```
swensel, line 52: Dropped `typescript`: we ship compiled JS plus `.d.ts`, so consumers don't compile us. `typescript@^5.9.3` made `npm create vite` (TS 6) ERESOLVE on 1.11.0, reproduced in my-app. TS stays a devDependency of this repo. Dropped `next`: nothing on the barrel imports it anymore, `ErrorPage` reads `globalThis.location` now. A Vite app shouldn't install Next 16 to render a dialog. `apps/main` is still Next.

packages/ui-common/package.json
```diff
@@ -52,15 +52,18 @@
...
         "@emotion/react": "^11.14.0",
         "@emotion/styled": "^11.14.1",
-        "@mui/icons-material": "^9.0.0",
-        "@mui/material": "^9.0.0",
-        "@mui/system": "^9.0.0",
-        "@mui/x-tree-view": "^9.1.0",
-        "next": "^16.3.0",
-        "next-auth": "5.0.0-beta.32",
+        "@mui/icons-material": "^7.3.1 || ^8.0.0 || ^9.0.0",
+        "@mui/material": "^7.3.1 || ^8.0.0 || ^9.0.0",
```
swensel, line 56: 1.5.1 raised these to `^9` in a patch. This repo's catalog is 9, but neuro-ui and nsflow are on 7, and `after/` installs on 7.3.1. Claim is install plus tsc on 7, not pixel-identical rendering across 7 and 9.

packages/ui-common/package.json
```diff
@@ -52,15 +52,18 @@
...
+        "@mui/system": "^7.3.1 || ^8.0.0 || ^9.0.0",
+        "@mui/x-tree-view": "^8.22.0 || ^9.1.0",
+        "next-auth": "^5.0.0-beta.29",
```
swensel, line 59: Exact `5.0.0-beta.32` excluded neuro-ui, who are on beta.30. Optional so a Vite app without login doesn't install it. Still required at runtime for anyone importing the `Authentication` subpaths.

packages/ui-common/package.json
```diff
@@ -73,10 +76,33 @@
     "types": "dist/index.d.ts",
     "exports": {
         ".": {
-            "import": "./dist/index.js",
-            "types": "./dist/index.d.ts"
+            "types": "./dist/index.d.ts",
+            "import": "./dist/index.js"
+        },
+        "./const": {
```
swensel, line 82: `./const` pointed at `./const.ts`, which `files: ["dist"]` never packed. The map was wrong, the JS was always there. Subpaths added because nsflow was aliasing into `dist/controller/...` when only `.` and `./const` were exported. `types` before `import` so TS 6 and bundler resolution pick the `.d.ts` first.

packages/ui-common/utils/text.ts
```diff
@@ -54,11 +52,25 @@ export const extractId = (modelId: string, modelType: "prescriptor" | "rio"): st
...
- * Hashes a string using MD5. For example: for generating keys for React nodes from a longer string identifer.
+ * Hashes a string. For example: for generating keys for React nodes from a longer string identifier.
+ *
+ * Not a cryptographic hash. Do not use it for anything security related.
  *
  * @param input String to be hashed
- * @returns Hashed string
+ * @returns Hashed string, as lowercase hex
  */
 export const hashString = (input: string): string => {
```
swensel, line 62: `hashString` used Node MD5 to generate React keys (#163). Bundlers stub that builtin, the app throws, consumers polyfilled. Non-crypto JS mix instead. Keys only, not security.

packages/ui-common/utils/text.ts
```diff
@@ -54,11 +52,25 @@ export const extractId = (modelId: string, modelType: "prescriptor" | "rio"): st
...
-    return createHash("md5").update(input).digest("hex")
+    let acc1 = 0xdeadbeef
+    let acc2 = 0x41c6ce57
+
+    for (let index = 0; index < input.length; index += 1) {
+        const code = input.codePointAt(index) ?? 0
+        acc1 = Math.imul(acc1 ^ code, 2654435761)
+        acc2 = Math.imul(acc2 ^ code, 1597334677)
+    }
```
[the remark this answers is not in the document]
swensel: Real mechanism, but the hash was already stable. For `"\u{1D400}"` the loop mixed in `0x1d400` and then the lone low surrogate `0xdc00`, yet the same input always produced the same output, which is the only contract here (non-crypto, React keys only). Switched to `for...of` anyway, since `codePointAt` next to `index += 1` reads like a bug.

packages/ui-common/const.ts
```diff
@@ -17,7 +17,8 @@ limitations under the License.
 // Name to use for application
 export const LOGO: string = "Neuro® AI"
 
-export const NEURO_SAN_UI_VERSION = process.env["NEXT_PUBLIC_NEURO_SAN_UI_VERSION"] ?? "Unknown Version"
+// Read through globalThis: bundlers targeting the browser do not define `process`
+export const NEURO_SAN_UI_VERSION = globalThis.process?.env?.["NEXT_PUBLIC_NEURO_SAN_UI_VERSION"] ?? "Unknown Version"
```
swensel, line 21: Bare `process.env` is a ReferenceError in a browser bundle. Optional chaining off `globalThis` falls back to `Unknown Version`. Last `process` read in the package, so `/const` imports with no Node polyfill.

packages/ui-common/README.md
```diff
@@ -1,67 +1,78 @@
 # @cognizant-ai-lab/ui-common
 
-A comprehensive React component library and utilities package for building user interfaces for Neuro-san AI
```
swensel, line 3: Rewrite so the docs match the package. The old README used create-react-app, had no working import in its quickstart, and documented props that do not exist (`userInfo`/`backendNeuroSanApiUrl` on `MultiAgentAccelerator`, `targetAgent` on `ChatCommon`, `useEnvironment`/`useUserInfo`). New: Vite, real props, MUI 7/8/9, next-auth optional, subpaths vs barrel, and a troubleshooting section. Features, Technologies Used, and Package Structure are kept. The copyright text is unchanged, moved to the bottom under `## License`.

